PR #1, opened 2017-02-06
Linting changes and new config API

eliperelman:
### Linting

In order to make linting easier to configure, I have moved all the rules to be inline to the plugin config. This means that making a linting rule change just means calling the loader and changes the options rules!

## Config API

I also wrote a new package called [webpack-chain](https://github.com/mozilla-rpweb/webpack-chain), which gives a fluent API for creating Webpack configs. There were other solutions to this out there, but they either didn't support Webpack 2, or didn't give us the granularity we needed to modify configs across projects.

Now any preset that inherits from our presets will get this chaining API for free to make modifications. If they want an escape hatch, they can call `.toConfig()` and just change the config object as before.

## Config from package.json

All too often there will be times where I just want to override something simple in the configuration, but it takes a bit of work to have to create a preset and then switch the package.json to use this new preset. This PR now also merges any Webpack config specific in `config.neutrino` with the preset config.

I know this is big, but I'd appreciate a glance at this before merging, and to discuss any concerns. Thanks!


eliperelman:
I'm going to go ahead and merge this now, but let's still discuss any thoughts here, and we can iterate on master.

helfi92:
This is really neat. A big thumbs up! I didn't have time to go over all the details but from a high level, it makes a lot of sense to have this.

Issue #9, opened 2017-02-08
Expose Neutrino API


eliperelman:
Right now we have `main` set in package.json, but obviously that's not able to be used since it just points to the neutrino bin. Maybe we can instead expose Neutrino as a requireable API as well which the CLI consumes.

Doing this could allow presets to augment Neutrino with additional functionality. This is probably the answer to #8, since we could provide an API for presets to inject the ability to include their own test provider.

On the flip side, this probably also means needing to return functions from presets rather than configuration. Also, overriding can't be done directly against an imported object.

```js
// preset
module.exports = (neutrino) => {
  neutrino.configure({ ...someConfig });

  neutrino.addTestProvider('jest', (config) => {
    executeJestOrSomething(config);
  });
};
```

eliperelman:
Fixed in #10. Neutrino exposes APIs for `.extend` and `.on` events.

Issue #8, opened 2017-02-08
First-class support for Jest in neutrino-preset-react


eliperelman:
I guess Jest is the official testing mechanism for React projects, so it would be nice if you could provide a way to use Jest as the default when using neutrino-preset-react.

On a side note, I'm not sure I like the path that Neutrino is heading down for testing platforms. In order to execute testing, every test function lives in Neutrino itself, and configuration options are passed to it via the presets. This means Karma and mocha both live in Neutrino. Need to add Jest? Well it sounds like that needs a change in Neutrino as well. I'd like to see if there's a way we can overcome this while not putting any burden on the preset consumer.

eliperelman:
Fixed in #10. Jest is its own preset.

Issue #7, opened 2017-02-08
Allow a project to specify multiple presets


eliperelman:
I can imagine a project might want to consume multiple presets:

- neutrino-preset-react
- a flow preset
- a SASS preset

By breaking up presets into smaller chunks, we could theoretically allow larger presets and projects to be compositions of many presets.

We would probably need to allow `--preset` to be used multiple times:

```sh
neutrino build --preset neutrino-preset-react --preset flow --preset sass
```

Also for package.json, probably switch `preset` to `presets`:

```json
{
  "config": {
    "presets": [
      "neutrino-preset-react",
      "flow",
      "sass"
    ]
  },
  "scripts": {
    "build": "neutrino build"
  }
}
```

eliperelman:
Fixed in #10.

PR #10, opened 2017-02-10
Pluggable event architecture mode, new test presets

eliperelman:
@helfi92 forgive me :laughing:

So, this patch attempts to make Neutrino a bit more pluggable. By defining new events that preset authors can attach to, they can hook into the lifecycle of Neutrino and make changes.

What this means for the "test" command is that without a preset that's listening for the "test" event, it will essentially do nothing. It is up to test preset authors to listen for this event, do some kind of build or compilation if necessary, then execute tests. You can see how I accomplished this with 3 different test presets: mocha, Karma, and Jest.

Let me know what you think. I'll merge this probably tomorrow, then start in on docs.


eliperelman:
This pull request solves issues #7, #8, and #9.

helfi92:
@eliperelman It looks great :) How can I run it locally.

`yarn && yarn clean-all && yarn-bootstrap` on neutrino-dev and `yarn link neutrino neutrino-preset-react` on a react project is not working for me.

eliperelman:
If you are already bootstrapped, you can just run `yarn link-all` and it will set up the links.
You are already linking in the dependent project which is fine.

I know it's dumb, but I also have to run `npm i eslint-plugin-react` in `packages/neutrino-preset-base`.

What error do you get?

eliperelman:
This should also fix #11.

helfi92:
Ok so I did what you suggested, but without success. I am getting:

```
TypeError: Cannot read property 'node' of undefined
at emitForAll.then (/Users/hassanali/Documents
/Mozilla/projects/neutrino-dev/packages/neutrino/src/neutrino.js:118:26)
```

Any thoughts?

eliperelman:
Were you trying this in a Node project?

helfi92:
I was trying it with treeherder-manifest

helfi92:
I'll try one more thing, give me a moment

helfi92:
Argh, it's still failing. When I try it on the master branch, it works.

I also tried deleting the neutrino-dev and downloading only the branch with:

```
git clone -b plugin-mode https://github.com/mozilla-neutrino/neutrino-dev.git
yarn && yarn bootstrap
```

I deleted treeherder-manifest, cloned it back again and did:

```
yarn && yarn link neutrino neutrino-preset-react
yarn run start
```

I get the same error in my earlier comment.

eliperelman:
Nuts, ok I'll try and replicate.

eliperelman:
@helfi92 can you try upgrading your installation of Yarn? https://yarnpkg.com/en/docs/install

helfi92:
I upgraded to yarn version 0.19.1. I redid the steps of my last comment but I have still have the same error. Did you try replicating my steps to see if it works on your machine?

eliperelman:
I seem to have problems too when starting from scratch. Maybe it's a problem with yarn-autolink. I'll see if putting a different solution in place will help.

eliperelman:
@helfi92 ok, I think I have most of the kinks worked out. I'm using a different monorepo package that seems to work well. I also added eslint-plugin-react to the base preset's dev dependencies.

So, the process should be:

- In neutrino-dev: `yarn && yarn bootstrap`
- In treeherder-manifest: `yarn && yarn link neutrino neutrino-preset-react`
- In treeherder-manifest, edit package.json and change `config.preset` to `config.presets`.

Let me know.

helfi92:
@eliperelman Worked from the first try :)

By the way, I think HMR is not working. Previously we made sure react-hot-loader/patch appeared first in the `entry.index` using `unshift`. `config.entry.index.unshift(require.resolve('react-hot-loader/patch'));`. Is this still occuring with the new changes?

helfi92:
Ah, I get the following error in the console:

`React Hot Loader: It appears that "react-hot-loader/patch" did not run immediately before the app started. Make sure that it runs before any other code. For example, if you use Webpack, you can add "react-hot-loader/patch" as the very first item to the "entry" array in its config. Alternatively, you can add require("react-hot-loader/patch") as the very first line in the application code, before any other imports.`

eliperelman:
OK, my mistake. Will update. Glad to hear the repo is working now. :)

eliperelman:
@helfi92 ok, updated. I had to add a method to webpack-chain to allow prepending to an entry instead of only appending.

eliperelman:
Going to merge so we can continue to iterate.

helfi92:
@eliperelman, thanks for adding `prepend`. I took a look and I had to make 2 small changes to make HMR work. We need a hot property in `config.devServer`.

First, replace `add` in `.add(require.resolve('react-hot-loader/patch'));` with `prepend` to make use of the new functionality :)

Second, add the following method inside `webpack-chain/DevServer`

```javascript
  hot(hot) {
    return this.set('hot', hot);
  }
```
After that, in `neutrino-preset-web`, append `.hot(true)`to `config.devServer`.

eliperelman:
@helfi92 I thought you couldn't use the `hot` property with the `HotModuleReplacementPlugin`. Am I wrong?

helfi92:
@eliperelman According to [dan's post](https://github.com/gaearon/react-hot-loader/issues/243/#issuecomment-211957140) about enabling HMR, we need both.

eliperelman:
@helfi92 awesome!

eliperelman:
Actually, following the steps in your comment blows up for me. According to that link, we need to wrap the root element we are rendering in an `AppContainer` exported from `react-hot-loader`. This will probably need to be pushed into the dependent project if they want to enable hot reloading. Lemme see if I can get a PR open to demonstrate.

helfi92:
@eliperelman Try this one: https://github.com/helfi92/neutrino-test

eliperelman:
@helfi92 I'm trying it against treeherder-manifest and all I get is this:

error

eliperelman:
Let's move this conversation to #15.

PR #22, opened 2017-02-15
Removing base preset, adding Airbnb preset

### LINKED ISSUE #21: Should we rip out linting from the base preset into linting presets?
The AirBnB linting rules we are using for the base preset and propagating to all other presets can be cumbersome to override. What if instead we extracted the linting preset out to its own? I could see this benefit as:

- neutrino-preset-airbnb
- neutrino-preset-standard

If a project really doesn't want linting, they can exclude it altogether. Thoughts?

eliperelman:
Fixes #21


eliperelman:
Haven't tested this yet, but we can clean up bugs by testing out sample repos.

PR #34, opened 2017-02-20
Fix typo in preset-jest

kedoska:
Documentation shows that to use jest, you must install the Jest presets and use the Mocha script. Which does not work.

Signed-off-by: Marco Casula <mc@clickbit.net>


eliperelman:
Thanks!

Issue #36, opened 2017-02-21
Add dependencies to neutrino-dev


darkwing:
The Jest preset is an awesome step forward.  A few more dependencies we should add:

- react-addons-test-utils
- enzyme

These allow more react-specific testing, like `enzyme` to create `shallow` JSX components.  I also needed `react-addons-test-utils` just for tests to run.

eliperelman:
So, since these are dependencies of the project itself, and need to be required from the test, they will need to also be dependencies of the project. Just like React and ReactDOM aren't included in the `neutrino-preset-react`, these 2 dependencies need to also be required by the project.

Since Jest isn't technically React specific, I think this is going to boil down to better documentation. We should probably add a section to the Jest preset that if you are testing React components, to also include these as `devDependencies`.

eliperelman:
In docs:

https://neutrino.js.org/presets/neutrino-preset-jest/#installation-react-testing

PR #37, opened 2017-02-21
Jest docs: Usage with React

eliperelman:
Addresses #36.

cc: @darkwing @helfi92 If you could let me know if this seems sufficient, that would be great.


helfi92:
Looks good

PR #38, opened 2017-02-21
missing object in the example of preset

kedoska:
`neutrino.config` exposes many objects. In order to create a _webpack's pipe_ it is necessary to use `neutrino.config.module`

Signed-off-by: Marco Casula <mc@clickbit.net>


eliperelman:
Great catch!

PR #41, opened 2017-02-22
Introduce lint base to make lint preset creation much easier

eliperelman:


docs/presets/neutrino-lint-base/README.md
```diff
@@ -0,0 +1,179 @@
...
+- Neutrino v4
+
+## Installation
+
+`neutrino-lint-base` can be installed via the Yarn or npm clients.
+
+#### Yarn
+
+```bash
+❯ yarn add neutrino-lint-base
```
helfi92: Should this be under dev dependencies?

eliperelman: No, the lint base is used in actual preset which is required by the dependent project. Extending from any preset needs it as a dependency.

eliperelman: The preset a **user is creating** will be installed in a project as a development dependency, the preset's dependency on another preset is a normal dep.

helfi92: Ah, OK, that makes sense.

PR #46, opened 2017-02-22
Adding dynamic import capabaility via import()

eliperelman:


eliperelman:
I think we also need `babel-plugin-syntax-dynamic-import` for the Node.js preset, but I haven't tested it yet.

eliperelman:
I haven't gotten this to work on Node.js, and I've filed an issue on the env preset on Babel to try and pin it down.

helfi92:
ouch...ok. keep us updated

eliperelman:
@helfi92 according to the Webpack docs, it seems like it should be possible, but I'm having a terrible time trying to get babel-preset-env to work with `async import()` in Node.js:

https://webpack.js.org/guides/code-splitting-import/

helfi92:
@eliperelman According to the [usage with babel and async await guidelines](https://webpack.js.org/guides/code-splitting-import/#usage-with-babel-and-async-await), to use `async/await` with `import()` we need `syntax-dynamic-import`, `transform-async-to-generator`, `transform-regenerator`,        `transform-runtime` as plugins. Are those taken care of with `babel-preset-env`?

eliperelman:
They theoretically should be excluded (the regenerator stuff), because Node.js 6.9 supports generators. We only need it to transform async to generator, which happens, but then messes up some regenerator stuff.

eliperelman:
Blocked by https://github.com/babel/babylon/issues/378

helfi92:
@eliperelman Seems like they fixed it

eliperelman:
Haha, I was literally just looking into it. For it to be fixed in Webpack, this change needs to make it in:

https://github.com/kesne/acorn-dynamic-import/pull/5

Right now Webpack v2.2.1 is on acorn-dynamic-import v2.0.0, and we need it to be v2.0.2 for the dynamic import fix.

eliperelman:
Derp, webpack uses version ranges, so I just need to regenerate the yarn.lock and it should be OK.

eliperelman:
Note that we still need the babylon fix since babel-preset-env is doing work here and needs to parse the syntax. Looking at the changelog for babylon, this might not happen until Babel 7.

eliperelman:
I think we should probably move forward with dynamic imports for Web only for now, since it will break down dynamic imports to ES5 friendly code based on the current babel-preset-env targets. Thoughts?

helfi92:
@eliperelman Sure, let's worry about making it work on Node later. For now, we can make it work on web. We can then put your suggested changes to react-router-starter and have people use it :)

eliperelman:
Resolving in #86.

PR #48, opened 2017-02-24
add eslint-jest-plugin

kebot:
Please review
https://www.npmjs.com/package/eslint-plugin-jest

### DIFF
--- a/packages/neutrino-preset-jest/package.json
+++ b/packages/neutrino-preset-jest/package.json
@@ -19,6 +19,7 @@
     "babel-plugin-transform-es2015-modules-commonjs": "^6.22.0",
     "babel-preset-jest": "^18.0.0",
     "deepmerge": "^1.3.2",
+    "eslint-plugin-jest": "^19.0.1",
     "jest-cli": "^18.1.0"
   },
   "homepage": "https://neutrino.js.org",
--- a/packages/neutrino-preset-jest/src/index.js
+++ b/packages/neutrino-preset-jest/src/index.js
@@ -65,6 +65,16 @@ module.exports = neutrino => {
       };
     });
 
+  if (neutrino.config.rules.has('lint')) {
+    neutrino.config.module
+      .rule('lint')
+      .loader('eslint', props => merge(props, {
+        options: {
+          plugins: ['jest']
+        }
+      }));
+  }
+
   neutrino.on('test', args => {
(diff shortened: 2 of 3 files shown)


packages/neutrino-preset-jest/src/index.js
```diff
@@ -65,6 +65,14 @@ module.exports = neutrino => {
       };
     });
 
+  neutrino.config.module
```
eliperelman: It isn't known for sure whether a linting rule has been added, so this needs to be wrapped in a conditional.

```js
if (neutrino.config.rules.has('lint')) {
```

packages/neutrino-preset-jest/src/index.js
```diff
@@ -65,6 +65,14 @@ module.exports = neutrino => {
...
+    .rule('lint')
+    .loader('eslint', ({ options }) => ({
+      options: merge(options, {
+        plugins: ['jest']
+      })
+    }))
```
eliperelman: Missing semicolon.

packages/neutrino-preset-jest/src/index.js
```diff
@@ -65,6 +65,14 @@ module.exports = neutrino => {
       };
     });
 
+  neutrino.config.module
+    .rule('lint')
+    .loader('eslint', ({ options }) => ({
```
eliperelman: I know I've established the pattern elsewhere, but I'm currently changing it. Could you move the merge here to the top level?

```js
.loader('eslint', props => merge(props, {
```

packages/neutrino-preset-jest/package.json
```diff
@@ -19,6 +19,7 @@
     "babel-plugin-transform-es2015-modules-commonjs": "^6.22.0",
     "babel-preset-jest": "^18.0.0",
     "deepmerge": "^1.3.2",
+    "eslint-plugin-jest": "^19.0.1",
```
eliperelman: Does this plugin version support the version of Jest we are using here? Jest 19 was recently released but we can't yet switch to it because it's a breaking change.

kebot: This plugin only turns on/off some configs of ESLint, it only depends on ESlint package.

Check it's [package.json](https://github.com/facebook/jest/blob/master/packages/eslint-plugin-jest/package.json) for some details.

PR #63, opened 2017-02-25
Add custom method for generating an eslintrc config

eliperelman:
- [x] Generate an ESLint configuration usable by `.eslintrc.js`
- [x] Document a recipe for creating the eslintrc file through the API:

```js
// .eslintrc.js
const Neutrino = require('neutrino');
const pkg = require('./package.json');
const api = new Neutrino(pkg.config.presets);

module.exports = api.custom.eslintrc();
```

### DIFF
--- a/packages/neutrino-lint-base/src/index.js
+++ b/packages/neutrino-lint-base/src/index.js
@@ -1,9 +1,12 @@
+const clone = require('lodash.clonedeep');
 const path = require('path');
 
 const MODULES = path.join(__dirname, '../node_modules');
 const IF_NOT_DEV =  process.env.NODE_ENV !== 'development';
 
-module.exports = ({ config }) => {
+module.exports = neutrino => {
+  const { config } = neutrino;
+
   config
     .module
       .rule('lint')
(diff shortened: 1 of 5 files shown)


packages/neutrino-lint-base/src/index.js
```diff
@@ -37,4 +40,27 @@ module.exports = ({ config }) => {
...
   config.resolveLoader.modules.add(MODULES);
+
+  neutrino.custom.eslintrc = () => {
+    const options = clone(config.module.rule('lint').loaders.get('eslint').options);
+
+    options.extends = options.baseConfig.extends;
+    options.useEslintrc = true;
+    options.env = options.envs.reduce((env, key) => {
+      env[key] = true;
+      return env;
```
helfi92: @eliperelman Out of curiosity, in what scenario do you not prefer adding an empty line between an assignment and the return value.

eliperelman: I usually always do it, I'll change

packages/neutrino-lint-base/src/index.js
```diff
@@ -37,4 +40,27 @@ module.exports = ({ config }) => {
...
+    options.globals = options.globals.reduce((globals, key) => {
+      globals[key] = true;
+      return globals;
+    }, {});
+
+    delete options.envs;
+    delete options.baseConfig;
+    delete options.failOnError;
+    delete options.emitWarning;
+    delete options.emitError;
```
helfi92: Thoughts about this instead?
```javascript
['envs', 'baseConfig', 'failOnError', 'emitWarning', 'emitError']
  .forEach(prop => delete options[prop]);
```

packages/neutrino-lint-base/src/index.js
```diff
@@ -37,4 +40,27 @@ module.exports = ({ config }) => {
...
+
+    options.extends = options.baseConfig.extends;
+    options.useEslintrc = true;
+    options.env = options.envs.reduce((env, key) => {
+      env[key] = true;
+      return env;
+    }, {});
+    options.globals = options.globals.reduce((globals, key) => {
+      globals[key] = true;
+      return globals;
```
helfi92: Empty line between assignment and return value

Issue #64, opened 2017-02-26
neutrino.custom.eslintrc in lint presets don't output correct eslintrc json


timothyquach:
I've tried getting neutrino linting presets to output hints to vscode and found some bugs.
`neutrino.custom.eslintrc` throws an exception because a global is undefined and assumes `options.envs` is an array of objects.

eliperelman:
According to the docs, the rc file doesn't take `envs` like CLIEngine, but `env`, which is an object of environments to enable.

http://eslint.org/docs/user-guide/configuring

PR #65, opened 2017-02-26
Fixing neutrino.custom.eslintrc to not except

timothyquach:
Addressing https://github.com/mozilla-neutrino/neutrino-dev/issues/64

### DIFF
--- a/packages/neutrino-lint-base/src/index.js
+++ b/packages/neutrino-lint-base/src/index.js
@@ -22,7 +22,7 @@ module.exports = neutrino => {
           root: true,
           plugins: ['babel'],
           baseConfig: {},
-          envs: ['es6'],
+          envs: [{'es6': true}],
           parser: 'babel-eslint',
           parserOptions: {
             ecmaVersion: 2017,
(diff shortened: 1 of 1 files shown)


packages/neutrino-lint-base/src/index.js
```diff
@@ -22,7 +22,7 @@ module.exports = neutrino => {
           root: true,
           plugins: ['babel'],
           baseConfig: {},
-          envs: ['es6'],
+          envs: [{'es6': true}],
```
eliperelman: This property is still used by ESLint's CLIEngine, so this line needs to stay `envs: ['es6'],`.

packages/neutrino-lint-base/src/index.js
```diff
@@ -47,7 +47,7 @@ module.exports = neutrino => {
     options.extends = options.baseConfig.extends;
     options.useEslintrc = true;
     options.env = options.envs.reduce((env, key) => Object.assign(env, { [key]: true }, {}));
-    options.globals = options.globals.reduce((globals, key) => Object.assign(env, { [key]: true }, {}));
+    options.globals = options.globals.reduce((global, key) => Object.assign(global, { [key]: true }, {}));
```
eliperelman: Make each of these words `globals` to match the property name.

eliperelman: Just noticed that the paren for Object.assign is in the wrong spot. I'm going to create a quick patch and incorporate your changes.

eliperelman:
Deployed in v4.3.1.

PR #70, opened 2017-02-26
Adding Module to preset config

hankandre:
Added the `.module` to the `neutrino.config.rule()` statement.


hankandre:
No, thank you! Neutrino is so helpful. I'm loving it.

PR #78, opened 2017-02-28
Add Community section for starter kits

helfi92:


docs/learning-resources.md
```diff
@@ -10,12 +10,18 @@ community, file an issue or submit a pull request.
...
-[Express Quickstart Kit for Neutrino](https://github.com/mozilla-neutrino/express-starter)
+[Express Quickstart Kit](https://github.com/mozilla-neutrino/express-starter)
 
-[React Quickstart Kit for Neutrino](https://github.com/mozilla-neutrino/react-starter)
+[React Quickstart Kit](https://github.com/mozilla-neutrino/react-starter)
 
-[Web Quickstart Kit for Neutrino](https://github.com/mozilla-neutrino/web-starter)
+[Web Quickstart Kit](https://github.com/mozilla-neutrino/web-starter)
+
+## Community Starter Repositories
```
eliperelman: Let's just keep a single "Starter Repositories" section. We can still remove the "for Neutrino" bits, but let's put community contributions alongside core ones.

docs/learning-resources.md
```diff
@@ -10,12 +10,18 @@ community, file an issue or submit a pull request.
...
+
+Have a Neutrino starter kit to share? Send us a pull-request with the link and we'll add it here for others to see.
```
eliperelman: We shouldn't need this statement since it's mentioned at the beginning of the page.

helfi92:
@jerolimov added [feathers-starter](https://github.com/jerolimov?tab=repositories) to the [Learning resources](https://neutrino.js.org/learning-resources.html). Thanks again for reaching out.

jerolimov:
Thanks @helfi92, but why not call it just "Feathers Starter Kit"? We can also move it to your orga if you like it. So anyone can see it there and we can keep it active there together..

jerolimov:
Current version is also great, thanks 👍

helfi92:
@jerolimov Our goal is to let the community grow organically and provide strength for others to be successful. For now, we won't move it to the organization because we won't be maintaining every starter kit.

Btw, I put it as `jerolimov/feathers-starter` since another contributor might create another starter kit but with slightly different config. For example, someone might create a starter kit without Hot Module Replacement. Hope that helps.

eliperelman:
@helfi92 I'm think long for consistency we should also change ours to the same format, repo/name.

eliperelman:
No rush though.

PR #80, opened 2017-02-28
dev-server/client host/port fix

SeeThruHead:
client entry file no longer receives default host and post when the
user sets host and port via config

### DIFF
--- a/packages/neutrino-preset-web/src/index.js
+++ b/packages/neutrino-preset-web/src/index.js
@@ -183,8 +183,12 @@ module.exports = ({ config }) => {
 
   if (process.env.NODE_ENV === 'development') {
     const protocol = !!process.env.HTTPS ? 'https' : 'http';
-    const host = process.env.HOST || 'localhost';
-    const port = parseInt(process.env.PORT) || 5000;
+    const host = process.env.HOST
+      || PKG.config && PKG.config.devServer && PKG.config.devServer.host
+      || 'localhost';
+    const port = parseInt(process.env.PORT)
+      || PKG.config && PKG.config.devServer && parseInt(PKG.config.devServer.port)
+      || 5000;
 
     config.devtool('eval');
     config.devServer


eliperelman:
Have you tried using `config.neutrino.devServer` in package.json?

SeeThruHead:
That is exactly what's broken. That's what this pr is addressing. 

Get Outlook for iOS




On Tue, Feb 28, 2017 at 5:05 PM -0500, "Eli Perelman" <notifications@github.com> wrote:












Have you tried using config.neutrino.devServer in package.json?

eliperelman:
Your PR is currently looking for data on `PKG.config.devServer`, not `PKG.config.neutrino.devServer`, which makes me wonder if you're configuring it from the right namespace.

eliperelman:
I tested it out locally with the following in my package.json:

```json
{
  "config": {
    "presets": [
      "neutrino-preset-web"
    ],
    "neutrino": {
      "devServer": {
        "port": 5769
      }
  }
}
```

And received the following in my console:

```bash
✔ Development server running on: http://localhost:5769
```

Closing, as this seems to work as expected, just probably a mis-configuration. Please re-open if you still see an issue. Thanks!

SeeThruHead:
Sorry I don't think I explained the issue properly.

The problem that i'm seeing is that the dev-server is running on port 4999 but the webpack-dev-server/client is trying to send websocket connections to port 5000 still.

From look at the code it doesn't seem like package.json config will ever reach the variables `port` and `host` in the changed file. 

This is the error:
```
abstract-xhr.js:132 GET http://localhost:5000/sockjs-node/info?t=1488326064431 net::ERR_CONNECTION_REFUSED
```

it can be fixed by doing `PORT=4999 yarn start` but I'm assuming you also want port configured inside package.json to be applied to the client.

my package.json config looks like so 
```
  "config": {
    "neutrino": {
      "devServer": {
        "port": 3999
      }
    },
    "presets": [
      "neutrino-preset-infl.js"
    ]
  },
```

PR #88, opened 2017-03-02
Add support building react-native app

guzart:
Based on [create-react-app webpack config](https://github.com/facebookincubator/create-react-app/blob/7c899fc392c7065b4d041d463b0a947b602ed356/packages/react-scripts/config/webpack.config.dev.js#L93)

Probably needs a merge with existing configuration aliases, e.g.
```javascript
config.resolve.set('alias', Object.assign({ 'react-native': 'react-native-web' }, config.resolve.get('alias'));
```

Do you think it's a good idea to add it to this preset?

### DIFF
--- a/packages/neutrino-preset-react/index.js
+++ b/packages/neutrino-preset-react/index.js
@@ -20,6 +20,7 @@ module.exports = (neutrino) => {
 
   config.resolve.modules.add(MODULES);
   config.resolve.extensions.add('.jsx');
+  config.resolve.alias.set('react-native', 'react-native-web');
   config.resolveLoader.modules.add(MODULES);
   config.externals({
     'react/addons': true,


eliperelman:
I know literally nothing about React Native, but looking at some other projects around the web using react-native-web and Webpack, it appears that this is the only real difference I can find between a normal React project.

My guess is if we merge this, we would need to add a note to the React preset docs about this alias so we avoid confusion about the module name.

@helfi92 any additional thoughts?

gasolin:
Does it make more sense to create `neutrino-preset-react-native-web`, which stack up based on `neutrino-preset-react`?

helfi92:
Agreed. It definitely makes more sense to have it outside the react preset. `neutrino-preset-react-native` should be good.

gasolin:
react-native itself doesn't contain web port, so its actually related to the react-native-web project. name it as preset-react-native will give people wrong expectation that the preset can also generate android/ios build

helfi92:
React native for web [doesn't seem to be stable](https://www.smashingmagazine.com/2016/08/a-glimpse-into-the-future-with-react-native-for-web/#conclusion). There are a lot of unsupported components to be used in a real project. Hmm...

guzart:
The smashing magazine is from August 2016 and I know of a project using it in production, my guess is that it has [better support now](https://github.com/necolas/react-native-web/graphs/contributors). 

Probably the most common use case for `react-native-web` is to export a react-native app to the web, in which case having a separate preset makes perfect sense. You use this specialized preset for exporting your native app to web. 

However, I think react-native and react-native-web made us think about the possibility of building applications for all platforms (web, macos, windows, linux, android, ios) using _one_ code-base with specialized components (similar to how react-native works right now). So in the future it will all fall under the `react` app umbrella. You create a `react` app and export to web, mobile and/or desktop.
Is [HAgnostic-News](https://github.com/grigio/HAgnostic-News) a react native app, or a react app?

packages/neutrino-preset-react/src/index.js
```diff
@@ -71,6 +71,7 @@ module.exports = neutrino => {
 
   config.resolve.modules.add(MODULES);
   config.resolve.extensions.add('.jsx');
+  config.resolve.set('alias', merge({ 'react-native': 'react-native-web' }, config.resolve.get('alias')));
```
eliperelman: This needs a rebase, and then you should be able to do:

```js
config.resolve.alias.set('react-native', 'react-native-web');
```

PR #95, opened 2017-03-05
add react & react-dom as peer dependency, fix #94

### LINKED ISSUE #94: add react and react-dom as peerDependencies in preset-react
it will give hints about what is needed to install separately

gasolin:


eliperelman:
Unfortunately we can't add these as peerDeps for a couple reasons:

- peerDeps can't be disabled. If someone wants to create a neutrino-preset-preact based on the react preset, they won't be able to get rid of the warnings.
- Using versions is a maintenance burden on the preset, and we made this mistake with Neo. By specifying a version, we are forced to track with React when their version changes, possibly falling behind.

I think the right thing to do is document it the best we can.

PR #102, opened 2017-03-08
eslint

aretecode:
add .eslintrc file, 
eslint fix script. 

add deps: prettier, prettier-eslint, eslint, eslint-plugin-react. 

use config from neutrino-preset-mozilla-rpweb


eliperelman:
I think we can remove the React bits. Unquote the property names, rebase, and I'll take another look.

eliperelman:
Thanks!

Issue #112, opened 2017-03-13
Allow for async resolution of presets


jaridmargolin:
I have some async work that needs to be done prior to configuration. Specifically I need to hit an endpoint and utilize the returned data in my `DefinePlugin` config. It would be great if I could do this async work within my preset:

```
module.exports = async ({ config }) => {
  const result = await doSomeAsyncStuff()
  ...
})
```

Architecturally would there be any reason this could not be done?

eliperelman:
I don't see any problem with it. It will probably mean needing to make a change to Neutrino's `use` to support Promised middleware, which should be pretty simple.

eliperelman:
Hmm, while this is easy enough to handle through the CLI, any API use of `.use` really gets a wrench thrown in it:

```js
neutrino.use(alpha);
neutrino.use(beta);
neutrino.use(gamma);

if (x) {
  neutrino.use(delta);
}
```

Each of these would need to wait for the previous use to finish, since we could not know which are async and which are not. If you have async work that needs to be done, you should probably consider doing that prior to starting Neutrino, maybe write it to a file, then have your preset import it.

jaridmargolin:
I was able to hack around and get this working locally by modifying the Neutrino constructor where the presets are required. I'll open up a PR with my approach so we can discuss details of implementation.

Thanks.

jaridmargolin:
Ahh.. Looks like I made my changes on an out of date distribution.

jaridmargolin:
Looking at your comment regarding `use`... Internally, can this method just push tasks onto a queue?

jaridmargolin:
- or even if use doesn't push onto a queue, why not just always have `use` return a promise:

```
Promise.resolve(preset(this, options));
```

And then you could change usage:

```
neutrino.use(alpha)
  .then(() => neutrino.use(beta));
  .then(() => neutrino.use(gamma));
  .then(() => x ? neutrino.use(delta) : undefined);
```

---

I think this is a more sound solution... `selenium-webdriver` at one point tried to implement an internal queue for its promises and was a mess to work with (they are currently in the promise of removing).

eliperelman:
A Queue is one possible solution, but it makes using `use` from middleware tricky. Essentially a call to `use` would then do nothing but put middleware on the stack for unwinding later. This means that everything within middleware must also be pushed onto the queue so it is done in the correct order:

```js
use(middleware, options) {
[...]

PR #113, opened 2017-03-14
Allow for async presets by introducing `load` method in Neutrino class.

jaridmargolin:
PR to start implementation discussion around #112

### DIFF
--- a/packages/neutrino/bin/neutrino
+++ b/packages/neutrino/bin/neutrino
@@ -88,17 +88,10 @@ function run(command, presets) {
       `Ensure this module can be found relative to the root of the project.`);
   });
 
-  // Also grab any Neutrino config from package.json and merge it into the config at a higher precedence
-  if (Object.keys(config).length) {
-    api.config.merge(config);
-  }
+  api.use(() => api.config.merge(config));
 
-  if (args.inspect) {
-    return inspect(api.getWebpackOptions());
-  }
-
-  api[command](args)
-    .then(() => process.exit(0))
+  api.load()
+    .then(() => args.inspect ?  inspect(api.getWebpackOptions()) : api[command](args))
     .catch(err => {
       if (err) {
         console.error(err);
--- a/packages/neutrino/src/neutrino.js
+++ b/packages/neutrino/src/neutrino.js
@@ -21,10 +21,15 @@ class Neutrino extends EventEmitter {
 
     this.config = new Config();
     this.options = merge(options, { root, source, output, tests, node_modules, entry });
+    this.useQueue = [];
   }
 
   use(preset, options = {}) {
-    preset(this, options);
+    this.useQueue.push(() => preset(this, options));
+  }
+
+  load() {
+    return this.useQueue.reduce((result, task) => result.then(task), Promise.resolve());
   }
 
   handleErrors(err, stats) {
--- a/packages/neutrino/test/api_test.js
+++ b/packages/neutrino/test/api_test.js
@@ -101,20 +101,22 @@ test('creates a Webpack config', t => {
         .options({ alpha: 'a', beta: 'b' });
   });
 
-  t.deepEqual(api.getWebpackOptions(), {
-    module: {
-      rules: [
-        {
-          test: /\.js$/,
-          include: ['src'],
-          use: [
-            {
-              loader: 'babel-loader',
-              options: { alpha: 'a', beta: 'b' }
-            }
-          ]
-        }
-      ]
-    }
-  });
+  api.load().then(() => {
+    t.deepEqual(api.getWebpackOptions(), {
+      module: {
+        rules: [
+          {
+            test: /\.js$/,
+            include: ['src'],
+            use: [
+              {
+                loader: 'babel-loader',
+                options: { alpha: 'a', beta: 'b' }
+              }
+            ]
+          }
+        ]
+      }
+    });    
+  })
 });


jaridmargolin:
Assuming we can agree on a direction, I will implement corresponding tests.

eliperelman:
What about using something like [Denque](https://www.npmjs.com/package/denque) which makes `.use` put Promise-returning functions on a queue, which we then completely unwind? Maybe something like:

```js
use(preset, options = {}) {
  this.queue.push(() => Promise.resolve(preset(this, options)));
}

load() {
  let current;
  let previous;

  while (!this.queue.isEmpty()) {
    current = this.queue.shift();

    if (!previous) {
      previous = current();
      continue;
    }

    previous = previous.then(current);
  }

  return previous;
}
```

```js
// usage
const api = new Neutrino();

api.use(alpha);
api.use(beta);
api.use(() => /* */);

api
  .load()
  .then(() => /* */);
```

tj:
Bluebird is a pretty big dep, I'd rather see something like https://github.com/daliwali/promise-compose personally, perf won't matter for a tiny set of promises.

eliperelman:
@tj oooo, nice find! I just took a look at the implementation, and it's very similar to what I was prototyping locally:

```js
constructor() {
  this.queue = [];
}

use(preset, options = {}) {
  this.queue.push(() => Promise.resolve().then(() => preset(this, options)));
}

load() {
  let loading = Promise.resolve();

  if (!this.queue.length) {
    return loading;
  }

  this.queue.forEach(preset => loading = loading.then(preset));

  return loading;
}
```

jaridmargolin:
👍 `use` / `'load` seems like a fine solution. Seems like you are moving forward with this @eliperelman? Or would you like me to continue and modify the work in this PR?

eliperelman:
@jaridmargolin feel free to run with the comments we put here.

eliperelman:
I was merely just seeing if the POC was viable.

packages/neutrino/bin/neutrino
```diff
@@ -88,17 +88,9 @@ function run(command, presets) {
...
-  }
-
-  if (args.inspect) {
-    return inspect(api.getWebpackOptions());
-  }
-
-  api[command](args)
-    .then(() => process.exit(0))
+  api.load()
+    .then(() => api.config.merge(config || {}))
```
eliperelman: This should probably be `use`d prior to load for consistency:

```js
api.use(() => api.config.merge(config));
```

Also, `config` is already defaulted to an empty object on line 62.

packages/neutrino/bin/neutrino
```diff
@@ -88,17 +88,9 @@ function run(command, presets) {
...
+    .then(() => args.inspect ?  inspect(api.getWebpackOptions()) : api[command](args))
```
eliperelman: Let's break this up into two sections:

```js
.then(() => args.inspect && inspect(api.getWebpackOptions()))
.then(() => api[command](args))
```

packages/neutrino/src/neutrino.js
```diff
@@ -21,10 +21,17 @@ class Neutrino extends EventEmitter {
 
     this.config = new Config();
     this.options = merge(options, { root, source, output, tests, node_modules, entry });
+    this.useQueue = []
```
eliperelman: Missing semicolon.

packages/neutrino/src/neutrino.js
```diff
@@ -21,10 +21,17 @@ class Neutrino extends EventEmitter {
...
   }
 
   use(preset, options = {}) {
-    preset(this, options);
+    this.useQueue.push(() => preset(this, options));
+  }
+
+  load() {
+    return this.useQueue.reduce((result, task) => {
```
eliperelman: I like the use of `reduce`, as well as the initial `Promise.resolve` to kick it off. Makes `use` leaner. :+1: 

To pass linting, change to:

```js
load() {
  return this.useQueue.reduce((result, task) => result.then(task), Promise.resolve());
}
```

packages/neutrino/bin/neutrino
jaridmargolin: I was under the impression inspect was supposed to stop execution of `api[command](args)`:

```
-  if (args.inspect) {
-    return inspect(api.getWebpackOptions());
-  }
```

packages/neutrino/bin/neutrino
jaridmargolin: Cool stuffs 👍

packages/neutrino/bin/neutrino
eliperelman: It's mostly a facade, `inspect` does a `process.exit`, see line 14.

jaridmargolin: Hmmmm... I'm not 100% in love with breaking them up. For someone else consuming the code, I don't think it is immediately apparent that inspect results in an exit.

Keeping them together in some sort of if/else or ternary shows that execution is one or the other. Ultimately your call. Just my 2 cents.

packages/neutrino/src/neutrino.js
jaridmargolin: Good catch. No need for assigning the result variable as I am already returning the result in reduce 👍

eliperelman:
I had a thought, and I don't think using a loop is going to work:

```js
api.use(alpha);
api.use(neutrino => new Promise(resolve => {
  neutrino.use(beta);
  resolve();
});
api.use(gamma);
```

Since the `beta` use isn't added to the stack until it is executed, it won't exist in the queue until all items have been pushed to the queue, which only works for top-level `use`s. Probably going to need to take a recursion approach. I'll think about it more and see if there's anything else I'm missing.

jaridmargolin:
Didn't realize this was a use case that needed to be satisfied? What would the expected behavior be in your opinion?

alpha
inline
gamma
beta

or

alpha
inline
gamma
beta

---

This to me is where the queue gets messy. If we went back to the original idea of just having use return a Promise, I think we would avoid a world of hurt.

eliperelman:
Having use return a Promise seems like a non-starter. That means we would have to do this in middleware:

```js
Promise
  .resolve()
  .then(() => neutrino.use(alpha))
  .then(() => neutrino.use(beta))
  .then(() => neutrino.use(gamma));
```

Using Ramda's `pipeP` wouldn't help much here either.

jaridmargolin:
I think the code is significantly more predictable when `use` returns a promise. Is your issue solely with verbosity?

async/await would significantly clean it up (uncertain of your targeted node versions).

```
await neutrino.use(alpha);
await neutrino.use(beta);
await neutrino.use(gamma);
```

eliperelman:
Support is LTS right now (Node.js v6.9+).

jaridmargolin:
I think you either support async and return a promise, or you stay sync. Execution order of nested `use` calls with a queue is going to be ugly and unpredictable.

For the average user consuming neutrino, they probably won't need be dealing directly with the `use` method. The added burden of an async `use` method is shifted to the preset & middleware authors. I personally don't find it bothersome, but I see you do.

Obviously I have a use case where I would prefer it to be async, so I would lean in that direction... but I understand if this isn't a concern for everyone.

jaridmargolin:
Quickly going back to your example:

```
api.use(alpha);
api.use(neutrino => new Promise(resolve => {
  neutrino.use(beta);
  // inline config
  resolve();
});
api.use(gamma);
```

In the queue solution, this could be interpreted several ways, with the authors intent probably being the 3rd (which I'm not even sure would be possible).

alpha
inline
gamma
beta

or

alpha
inline
beta
gamma

or

alpha
beta
inline
gamma

eliperelman:
@jaridmargolin I can't think of a reasonable solution to this either. I think we can move forward with this if you feel it meets your use case without imposing any additional burden on existing middleware or sync presets. Thoughts?

Also there are a few failing tests in Travis. :smiley:

packages/neutrino/bin/neutrino
```diff
@@ -88,17 +88,10 @@ function run(command, presets) {
...
+  api.use(() => api.config.merge(config));
 
-  if (args.inspect) {
-    return inspect(api.getWebpackOptions());
-  }
-
-  api[command](args)
-    .then(() => process.exit(0))
+  api.load()
+    .then(() => args.inspect ?  inspect(api.getWebpackOptions()) : api[command](args))
```
eliperelman: Remove extra space after `?`.

jaridmargolin:
> I think we can move forward with this

Not sure I understand. The queue implementation has way to many edge cases to introduce (IMO). The only reliable solution I see to offer async support is to alter the `use` command to return a resolved Promise (which you at one point called a "non-starter"). This will absolutely have implications for preset/middleware authors as mentioned above:

```
neutrino.use(alpha)
  .then(() => neutrino.use(beta))
  .then(() => neutrino.use(gamma));
  ...
```

Looking at the current presets in neutrino-dev, I could see how this could be undesired. Async/await would make the transition significantly easier, but it seems that would not be possible due to currently supported versions of node.

Now while the syntax would be temporarily clunky, if you plan to support this in the future, it seems like now would be the time to make the change. The farther you kick the can down the road, the more presets/middleware will require breaking changes in the future.

Or the decision can be made that `use` will always be synchronous, and any async work would need to be completed prior to running neutrino.

---

How do you want to proceed?

```
a. Update `use` to return a Promise
b. Close PR and require presets/middleware to execute sync
```

eliperelman:
Sorry, I meant move forward with this PR as it is.

eliperelman:
@helfi92 @tj @SpencerCDixon any opinions for moving Neutrino v5 support to Node 7 w/ async await support?

eliperelman:
We could also add a pipe method which could just be passed presets to load in series:

```js
neutrino.pipe(alpha, beta, gamma);
```

jaridmargolin:
@eliperelman - The current solution, and pipe, would both suffer from the same problem with nested calls to `use`, correct?

Because the ordering of execution effects override precedence, I think it would make more sense for the execution order to be simple/expected than to support async use cases.

eliperelman:
Yes, you are correct. I think for now let's punt on async presets and possibly revisit when Node.js v8 goes LTS. Thanks for putting so much though into this! (And if you think of another idea, let's talk about it.)

SpencerCDixon:
>  any opinions for moving Neutrino v5 support to Node 7 w/ async await support?

+1 for waiting for 8 and LTS

Issue #118, opened 2017-03-15
Remove process.exit(0) on run.


jaridmargolin:
```
  api[command](args)
    .then(() => process.exit(0))
    .catch(err => {
      if (err) {
        console.error(err);
      }

      process.exit(1);
    });
```

I am curious as to why `.then(() => process.exit(0))` needs to be chained off of the api command. As long as the catch does not hit, the process will already exit with a 0, correct?

This also has consequences of stopping execution of various plugins. For example [BundleAnalyzer](https://github.com/th0r/webpack-bundle-analyzer/blob/master/src/BundleAnalyzerPlugin.js#L33) executes after webpack has compiled.

eliperelman:
Resolved in #119.

PR #119, opened 2017-03-16
Neutrino refactor

jaridmargolin:
Tried to break these up into nice little reviewable commit chunks.

**1. Move require preset functionality to api.**
The idea here is to expose the requirePresets functionality for reuse by anyone wishing to consume the api. I don't see any specific reason why it would be specific to the CLI. Any oversight?

**2. Make run method in bin/neutrino self contained.**
This one may be stylistic, but I didn't understand why some variables were declared in the method, and some were declared at the top of the file. This moves all of the variables to where they are actually utilized. Additionally, I believe this has the added benefit of making `run` more portable if in the future there is reason to have it usable outside of the CLI.

**3. Move run process logic to location of call.**
The moves the logging and process management to the location in which `run` is called. I believe this is more appropriate place and makes the code easier to reason about... The idea is, "this is what we do when run fails." Additionally this closes, #118.

\* Hopefully not out of line making structural changes to the code. I have very much enjoyed working with the neutrino community, and thought I could give back a little.

### DIFF
--- a/packages/neutrino/bin/neutrino
+++ b/packages/neutrino/bin/neutrino
@@ -12,6 +12,7 @@ const stringify = require('javascript-stringify');
 const sort = require('deep-sort-object');
 const optional = require('optional');
 
+// eslint-disable-next-line no-console
 const inspect = pipe(sort, partialRight(stringify, [null, 2]), console.log, process.exit);
 const args = yargs
   .option('inspect', {
--- a/packages/neutrino/bin/neutrino
+++ b/packages/neutrino/bin/neutrino
@@ -63,35 +63,10 @@ function run(command, presets) {
   const api = new Neutrino(Object.assign(options, { config }));
 
   // Grab all presets and merge them into a single webpack-chain config instance
-  presets.forEach(preset => {
-    const paths = [
-      join(cwd, preset),
-      join(cwd, 'node_modules', preset),
-      preset
-    ];
-
-    for (let i = 0; i < paths.length; i += 1) {
-      try {
-        return api.use(require(paths[i]));
-      } catch (exception) {
-        if (/Cannot find module/.test(exception.message)) {
-          continue;
-        }
-
-        exception.message = `Neutrino was unable to load the module '${preset}'. ` +
-          `Ensure this module exports a function and is free from errors.\n${exception.message}`;
-        throw exception;
-      }
-    }
-
-    throw new Error(`Neutrino cannot find a module with the name or path '${preset}'. ` +
-      `Ensure this module can be found relative to the root of the project.`);
-  });
+  api.import(presets);
 
   // Also grab any Neutrino config from package.json and merge it into the config at a higher precedence
-  if (Object.keys(config).length) {
-    api.config.merge(config);
-  }
+  api.use(() => api.config.merge(config));
 
   if (args.inspect) {
     return inspect(api.getWebpackOptions());
--- a/packages/neutrino/bin/neutrino
+++ b/packages/neutrino/bin/neutrino
@@ -64,19 +64,12 @@ function run(command, args) {
   // Also grab any Neutrino config from package.json and merge it into the config at a higher precedence
   api.use(() => api.config.merge(config));
 
-  if (args.inspect) {
-    return inspect(api.getWebpackOptions());
-  }
-
-  api[command](args)
-    .then(() => process.exit(0))
-    .catch(err => {
-      if (err) {
-        console.error(err);
-      }
-
-      process.exit(1);
-    });
+  return args.inspect ?
+    inspect(api.getWebpackOptions()) :
+    api[command](args);
 }
 
-run(args._[0], args);
+run(args._[0], args).catch((err) => {
+  console.error(err || `Error during ${args._[0]}`);
+  process.exit(1);
+});
--- a/packages/neutrino/bin/neutrino
+++ b/packages/neutrino/bin/neutrino
@@ -26,6 +26,11 @@ const args = yargs
     default: [],
     global: true
   })
+  .option('env', {
+    description: 'The value for the environment variable, NODE_ENV',
+    string: true,
+    global: true
+  })
   .command('start', 'Build a project in development mode')
   .command('build', 'Compile the source directory to a bundled build')
   .command('test [files..]', 'Run all suites from the test directory or provided files', {
@@ -48,15 +53,12 @@ const args = yargs
   .argv;
 
 function run(command, args) {
-  const environments = { build: 'production', start: 'development', test: 'test' };
   const pkg = optional(join(process.cwd(), 'package.json')) || {};
   const pkgMiddleware = pathOr([], ['neutrino', 'use'], pkg);
   const middleware = [...new Set(pkgMiddleware.concat(args.use))];
   const options = pathOr({}, ['neutrino', 'options'], pkg);
   const config = pathOr({}, ['neutrino', 'config'], pkg);
   const api = new Neutrino(Object.assign(options, { config }));
-
-  process.env.NODE_ENV = environments[command];
 
   // Grab all middleware and merge them into a single webpack-chain config instance
   api.import(middleware);
--- a/packages/neutrino/bin/neutrino
+++ b/packages/neutrino/bin/neutrino
@@ -10,17 +10,9 @@ const { pathOr, pipe, partialRight } = require('ramda');
 const { join } = require('path');
 const stringify = require('javascript-stringify');
 const sort = require('deep-sort-object');
+const optional = require('optional');
 
 const inspect = pipe(sort, partialRight(stringify, [null, 2]), console.log, process.exit);
-const cwd = process.cwd();
-let pkg = {};
-
-try {
-  pkg = require(join(cwd, 'package.json'));
-} catch (ex) {}
-
-const pkgPresets = pathOr([], ['neutrino', 'presets'], pkg);
-const environments = { build: 'production', start: 'development', test: 'test' };
 const args = yargs
   .option('inspect', {
     description: 'Output a string representation of the configuration used by Neutrino and exit',
@@ -55,12 +47,16 @@ const args = yargs
   .help()
   .argv;
 
-function run(command, presets) {
-  process.env.NODE_ENV = environments[command];
-
+function run(command, args) {
+  const environments = { build: 'production', start: 'development', test: 'test' };
+  const pkg = optional(join(process.cwd(), 'package.json')) || {};
+  const pkgPresets = pathOr([], ['neutrino', 'presets'], pkg);
+  const presets = [...new Set(pkgPresets.concat(args.presets))];
   const options = pathOr({}, ['neutrino', 'options'], pkg);
   const config = pathOr({}, ['neutrino', 'config'], pkg);
   const api = new Neutrino(Object.assign(options, { config }));
+
+  process.env.NODE_ENV = environments[command];
 
   // Grab all presets and merge them into a single webpack-chain config instance
   api.import(presets);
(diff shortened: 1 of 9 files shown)


jaridmargolin:
Additionally, I didn't see tests built around the CLI functionality. Any insight on how you are currently working with this code and maintaining confidence in the changes?

eliperelman:
@jaridmargolin I typically don't test CLI-specific functionality as it usually involves spawning a process and reading its output. I'll typically stick with testing APIs. If someone wanted to add a PR for a way to test the CLI that was simple enough, I would accept it.

packages/neutrino/bin/neutrino
```diff
@@ -10,17 +10,8 @@ const { pathOr, pipe, partialRight } = require('ramda');
 const { join } = require('path');
 const stringify = require('javascript-stringify');
 const sort = require('deep-sort-object');
+const optional = require('optional');
```
eliperelman: Oooo, never seen `optional` before, very cool!

packages/neutrino/bin/neutrino
```diff
@@ -55,57 +46,32 @@ const args = yargs
...
-
+function run(command, args) {
+  const environments = { build: 'production', start: 'development', test: 'test' };
+  const pkg = optional(join(process.cwd(), 'package.json')) || {};
+  const pkgPresets = pathOr([], ['neutrino', 'presets'], pkg);
+  const presets = [...new Set(pkgPresets.concat(args.presets))];
   const options = pathOr({}, ['neutrino', 'options'], pkg);
   const config = pathOr({}, ['neutrino', 'config'], pkg);
   const api = new Neutrino(Object.assign(options, { config }));
+  const inspect = pipe(sort, partialRight(stringify, [null, 2]), console.log, process.exit);
```
eliperelman: Let's move `inspect` above `api` so the flow to Neutrino is clear.

packages/neutrino/bin/neutrino
```diff
@@ -55,57 +46,32 @@ const args = yargs
...
-        throw exception;
-      }
-    }
+  process.env.NODE_ENV = environments[command];
 
-    throw new Error(`Neutrino cannot find a module with the name or path '${preset}'. ` +
-      `Ensure this module can be found relative to the root of the project.`);
-  });
+  // Grab all presets and merge them into a single webpack-chain config instance
+  api.requirePresets(presets)
```
eliperelman: Let's either one-line this or put each chain on its own line.

```js
api
  .requirePresets(presets)
  .forEach(preset => api.use(preset));
```

packages/neutrino/bin/neutrino
```diff
@@ -55,57 +46,32 @@ const args = yargs
...
-  api[command](args)
-    .then(() => process.exit(0))
-    .catch(err => {
-      if (err) {
-        console.error(err);
-      }
-
-      process.exit(1);
-    });
+  return args.inspect
```
eliperelman: Put ternary operators on the preceding line:

```js
return args.inspect ?
  inspect(api.getWebpackOptions()) :
  api[command](args);
```

packages/neutrino/bin/neutrino
```diff
@@ -55,57 +46,32 @@ const args = yargs
...
+    ? inspect(api.getWebpackOptions())
+    : api[command](args);
 }
 
-run(args._[0], [...new Set(pkgPresets.concat(args.presets))]);
+run(args._[0], args)
```
eliperelman: There's only a single method call here, so we can probably put this one on the same line:

```js
run(args._[0], args).catch(err => {
  console.error(err || `Error during ${args._[0]}`);
  process.exit(1);
});
```

I like the updated error messaging. :)

packages/neutrino/src/neutrino.js
```diff
@@ -6,6 +6,7 @@ const Config = require('webpack-chain');
 const ora = require('ora');
 const merge = require('deepmerge');
 
+const cwd = process.cwd();
```
eliperelman: I'd like to not use `process.cwd()` here, if possible, outside of the constructor. See line 16, which introduces `options.root`.

packages/neutrino/src/neutrino.js
```diff
@@ -23,6 +24,34 @@ class Neutrino extends EventEmitter {
     this.options = merge(options, { root, source, output, tests, node_modules, entry });
   }
 
+  requirePresets(presets) {
+    return presets.map(preset => this.requirePreset(preset));
+  }
+
+  requirePreset(preset) {
+    const paths = [ join(cwd, preset), join(cwd, 'node_modules', preset), preset ];
```
eliperelman: If we use `options.root`, which is usually `process.cwd()`, we shouldn't need `cwd`. Remove inner spaces from the array brackets. Paths is intermediary, so we can remove the variable:

```js
const module = [
  join(this.options.root, preset),
  join(this.options.root, 'node_modules', preset),
  preset
].find(path => this.requirePath(path));
```

packages/neutrino/bin/neutrino
```diff
@@ -55,57 +46,32 @@ const args = yargs
...
-          `Ensure this module exports a function and is free from errors.\n${exception.message}`;
-        throw exception;
-      }
-    }
+  process.env.NODE_ENV = environments[command];
 
-    throw new Error(`Neutrino cannot find a module with the name or path '${preset}'. ` +
-      `Ensure this module can be found relative to the root of the project.`);
-  });
+  // Grab all presets and merge them into a single webpack-chain config instance
```
eliperelman: I just had a thought. We are using two different nomenclatures for _items_ we are requiring: "presets" and "middleware". Presets are a carryover from pre-v5, and presets are technically middleware. Instead of naming things by _what_ we are loading, what if we named them by the _action_ we are going to take on them?

**use**

We already do this in the API. We _use_ presets. We _use_ middleware. The CLI interface should be synonymous with the API action.

So I'm proposing to change the CLI flag from `--presets` to `--use`. Would you make that change along with this? Then the package.json object would also change from `neutrino.presets : []` to `neutrino.use : []`.

That would inform the name of this methods better too:

```js
api
  .requireMiddleware(middleware)
  .forEach(middleware => api.use(middleware))
```

Or maybe `requireMiddleware` is too verbose, and the only thing we ever require **is** middleware, so it can be simply `api.require` or `api.import`. One thing I like about the connotation of `import` is that it presupposes `use`:

```js
middleware.forEach(middleware => api.import(middleware)) // Does a require AND a use
```

TLDR; Let's change the CLI to use `--use`, change `neutrino.presets` to `neutrino.use` for package.json, use the variable of `middleware` instead of `presets`, and create `Neutrino#import` which does a `require` and a `use`.

eliperelman:
Really happy with the changes this PR brings, and I've given some comments about how we can improve the API and its clarity with the CLI even more.

jaridmargolin:
I will add tests for `import` functionality now that it is moved into the api and the functionality can more easily be run in isolation.

jaridmargolin:
Other than that, I believe I addressed all concerns mentioned in the review.

packages/neutrino/src/neutrino.js
```diff
@@ -23,8 +23,44 @@ class Neutrino extends EventEmitter {
...
 
-  use(preset, options = {}) {
-    preset(this, options);
+  use(middleware, options = {}) {
+    middleware(this, options);
+  }
+
+  import (middlewares) {
+    middlewares = Array.isArray(middlewares) ? middlewares : [middlewares];
+    this.requireMiddlewares(middlewares).forEach(middleware => api.use(preset));
```
eliperelman: Can probably inline the middleware (also still a reference to `preset` here):

```js
this
  .requireMiddleware(Array.isArray(middleware) ? middleware : [middleware])
  .forEach(middleware => api.use(middleware));
```

packages/neutrino/src/neutrino.js
```diff
@@ -23,8 +23,44 @@ class Neutrino extends EventEmitter {
...
+  }
+
+  requireMiddlewares(middlewares) {
```
eliperelman: We can remove this method, if we do the mapping in `requireMiddleware`, and normalize the array. See my comment there.

packages/neutrino/src/neutrino.js
```diff
@@ -23,8 +23,44 @@ class Neutrino extends EventEmitter {
...
+    return middlewares.map(middleware => this.requireMiddleware(middleware));
+  }
+
+  requireMiddleware(middleware) {
```
eliperelman: We can make this handle singular or plural middleware, just like `import`:

```js
requireMiddleware(middleware) {
  if (!Array.isArray(middleware)) {
    middleware = [middleware];
  }

  return middleware.map(middleware => {
    // ...
  });
}
```

packages/neutrino/src/neutrino.js
```diff
@@ -23,8 +23,44 @@ class Neutrino extends EventEmitter {
...
+    const paths = [
```
eliperelman: This should be named `module` instead of `paths`.

packages/neutrino/src/neutrino.js
jaridmargolin: This is why I need to get some tests in place :)

packages/neutrino/src/neutrino.js
jaridmargolin: This is why I need to get some tests in place :)

eliperelman: I'm actually curious how this wasn't picked up by linting in travis.

packages/neutrino/test/requireMiddleware_test.js
```diff
@@ -0,0 +1,45 @@
...
+const cwd = process.cwd()
+const poutputFile = pify(outputFile);
+const premove = pify(remove);
+
+const rootPath = join(__dirname, 'test-module');
+const rootMiddlewarePath = join(rootPath, 'middleware.js');
+const errorMiddlewarePath = join(rootPath, 'errorMiddleware.js');
+const modulePath = join(rootPath, 'node_modules', 'mymodule');
+const moduleMiddlewarePath = join(modulePath, 'index.js');
+
```
eliperelman: Remove extra line.

packages/neutrino/test/requireMiddleware_test.js
```diff
@@ -0,0 +1,45 @@
...
+
+test.before(t => {
```
eliperelman: Async functions:

```js
test.before(async (t) => {
  await Promise.all([
    outputFile(rootMiddlewarePath, `module.exports = 'root';`),
    outputFile(errorMiddlewarePath, '[;'),
    outputFile(moduleMiddlewarePath, `module.exports = 'mymodule';`)
  ]);
  process.chdir(rootPath);
});
```

packages/neutrino/test/requireMiddleware_test.js
```diff
@@ -0,0 +1,45 @@
...
+  return Promise.all([
+    poutputFile(rootMiddlewarePath, 'module.exports = "root"'),
+    poutputFile(errorMiddlewarePath, '[;'),
+    poutputFile(moduleMiddlewarePath, 'module.exports = "mymodule"')
+  ]).then(() => process.chdir(rootPath));
+});
+
+test.after.always(t => {
```
eliperelman:
```js
test.after.always(async (t) => {
  await remove(rootPath);
  process.chdir(cwd);
};
```

packages/neutrino/bin/neutrino
```diff
@@ -55,57 +51,28 @@ const args = yargs
...
-      process.exit(1);
-    });
+  return args.inspect ?
+    inspect(api.getWebpackOptions()) :
+    api[command](args);
 }
 
-run(args._[0], [...new Set(pkgPresets.concat(args.presets))]);
+run(args._[0], args).catch(err => {
+  console.error(err || `Error during ${args._[0]}`)
```
jaridmargolin: Ahhh good. I'll rebase your changes (I've been using standard lately, and sometimes its difficult to switch between projects which don't use it).

eliperelman:
Very well done.

PR #123, opened 2017-03-19
Rename `getWebpackOptions` to `getWebpackConfig`.

jaridmargolin:


jaridmargolin:
One other option here could be to actually just use a getter and access via `api.webpackConfig`. Thoughts?

jaridmargolin:
Although, `getWebpackConfig()` may better describe that work is done to get the value.

eliperelman:
Agreed, I like having the method call over the prop.

docs/api/README.md
```diff
@@ -289,14 +289,14 @@ api
   .then(() => console.log('all passed'));
 ```
 
-### `getWebpackOptions()`
+### `getWebpackConfig()`
 
 While tools like webpack-chain provide a convenient API for creating Webpack configurations, this is not a format that
-is understandable by Webpack. With `getWebpackOptions()`, the webpack-chain instance at `.config` will be converted to
+is understandable by Webpack. With `getWebpackConfig()`, the webpack-chain instance at `.config` will be converted to
 an configuration object readable directly by Webpack.
```
eliperelman: Can you change "an" to "a" here?

PR #127, opened 2017-03-19
Neutrino webpack

jaridmargolin:
This is the least tested functionality in the `neutrino` package. Hard to make changes with a high degree of confidence. I am willing to write tests around this, but wanted to discuss what it would look like first. Any thoughts? I don't see a path to mocking the tests. May actually have to compile?

### DIFF
--- a/packages/neutrino/src/neutrino.js
+++ b/packages/neutrino/src/neutrino.js
@@ -1,16 +1,16 @@
 const { join, isAbsolute } = require('path');
 const { EventEmitter } = require('events');
-const DevServer = require('webpack-dev-server');
-const webpack = require('webpack');
 const Config = require('webpack-chain');
-const ora = require('ora');
 const merge = require('deepmerge');
-const { defaultTo } = require('ramda');
+const { defaultTo, identity } = require('ramda');
 const requireMiddleware = require('./requireMiddleware');
+const build = require('./webpack/build');
+const develop = require('./webpack/develop');
+const watch = require('./webpack/watch');
 
 const normalizePath = (path, root) => (isAbsolute(path) ? path : join(root, path));
 
-class Neutrino extends EventEmitter {
+module.exports = class Neutrino extends EventEmitter {
   constructor(options = {}) {
     super();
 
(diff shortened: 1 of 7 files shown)


packages/neutrino/src/neutrino.js
```diff
@@ -67,18 +47,18 @@ class Neutrino extends EventEmitter {
...
-    return this.runCommand('start', args, () => (this.getWebpackOptions().devServer ? this.devServer() : this.watcher()));
+    return this.runCommand('start', args, this.getWebpackConfig().devServer ? develop : watch);
   }
 
   test(args) {
     return this.runCommand('test', args);
   }
 
-  runCommand(command, args = {}, fn) {
+  runCommand(command, args = {}, fn = noop) {
```
eliperelman: We probably don't need a special `noop` here, we can either use `identity` from Ramda, or `Function.prototype`:

```js
runCommand(command, args = {}, fn = identity) {
```

```js
runCommand(command, args = {}, fn = Function.prototype) {
```

packages/neutrino/src/webpack/develop.js
```diff
@@ -0,0 +1,31 @@
+const ora = require('ora');
+const merge = require('deepmerge');
+const webpack = require('webpack');
+const DevServer = require('webpack-dev-server');
+
+module.exports = _config => new Promise((resolve) => {
+  const config = merge({}, _config);
```
eliperelman: Maybe this could be consolidated to some defaults:

```js
const config = merge({
  devServer: {
    https: false,
    host: 'localhost',
    port: 5000,
    noInfo: true
  }
}, _config);
const protocol = config.devServer.https ? 'https' : 'http';
const { host, port } = config.devServer;
const url = `${protocol}://${host}:${port}`;
```

packages/neutrino/src/webpack/utils.js
```diff
@@ -0,0 +1,24 @@
+module.exports = {
```
eliperelman: Seeing this makes me wish we had ES modules here. _sigh_

Maybe we can save some nesting by just naming them:

```js
module.exports.handle = /* ... */

module.exports.logStats = /* ... */

module.exports.logErrors = /* ... */
```

We could also use object shorthand to reduce nesting:

```js
function handle() {}

function logStats() {}

function logErrors() {}

module.exports = {
  handle,
  logStats,
  logErrors
};
```

eliperelman:
Testing the stuff that touches Webpack is hard, mostly because we want to avoid actually building something, and if we do, how would we ensure the output is correct, or that is starts as expected? Half that battle is just making sure Webpack operates as expected, and I'm not sure that is a road we even want to go down.

jaridmargolin:
I don't think the output is necessary to test. It would be great however, if we could test if the command runs without errors. I think this WOULD require actually building assets. Could be similar to the setup in requireMIddleware tests?

Issue #129, opened 2017-03-20
Re-expose compile targets on options


eliperelman:
In master we pass the compile targets directly to the Babel preset with some default values, but these are then difficult to re-access in other places. We should put the defaults also on `neutrino.options.compile.targets` for ease of re-use.

eliperelman:
Fixed in #137.

PR #137, opened 2017-03-23
Functional refactor

eliperelman:
I've been wanting to do a functional architecture for a while in Neutrino, and hopefully this gives us a good base. I'm hoping this leads to make testable units inside the API without so much reliance on state.


packages/neutrino-middleware-clean/index.js
```diff
@@ -6,5 +6,5 @@ module.exports = (neutrino, options) => {
 
   neutrino.config
     .plugin('clean')
-    .use(CleanPlugin, [paths, { root }]);
+    .use(CleanPlugin, [paths, { root, verbose: false }]);
```
jaridmargolin: 👍 seemed unnecessary to see this on every build

packages/neutrino/bin/neutrino
```diff
@@ -53,26 +50,58 @@ const args = yargs
...
+  const errors = Array.isArray(err) ? err : [err];
 
-  return args.inspect ?
-    inspect(api.getWebpackConfig()) :
-    api[command](args);
-}
+  errors.forEach((err) => {
+    console.error(err.stack || err);
+    err.details && console.error(err.details);
+  });
```
jaridmargolin: Hmmm... I am assuming this was moved from the commands, which I think makes sense, but maybe this should be exposed somewhere, so people consuming the api are capable of easily logging result as well?

packages/neutrino/bin/neutrino
```diff
@@ -53,26 +50,58 @@ const args = yargs
...
-  // Grab all middleware and merge them into a single webpack-chain config instance
-  api.import(middleware);
+  return Object.assign(options, { config, args });
+};
 
-  // Also grab any Neutrino config from package.json and merge it into the config at a higher precedence
-  api.use(() => api.config.merge(config));
+// fail :: Error|Array Error -> IO ()
+const fail = (err) => {
+  const errors = Array.isArray(err) ? err : [err];
```
jaridmargolin: Maybe we could add a `castToArray` method to utils. I believe this is used in multiple places.

eliperelman: Agreed.

packages/neutrino/bin/neutrino
eliperelman: Good idea, I'll normalize this to always pass an array of errors.

eliperelman:
Also fixes #129.

PR #141, opened 2017-03-25
Adding documention about web publicPath

### LINKED ISSUE #133: default publicPath of './' causes 404 with deep paths
Currently the `publicPath` defaults to `./`, so the generated `index.html` file has paths like `./index.bundle.js` referenced in it.

This works fine when loading the base page in the dev server at `http://localhost:5000`, but if you happen to start browsing in your dev server at a deep link like `http://localhost:5000/some/deep/link` the bundled js files return a 404 because of the leading `.`.

A (possibly naive) fix is changing the `publicPath` to `/` instead, which changes the bundle js references to `/index.bundle.js` which work from any url depth in the dev server. I haven't looked very deeply to see if there are any negative consequences to this change, but from working with this change on a project myself it seems to not cause any problems in either the dev server or built output.

Maybe I'm just doing something wrong, but I thought I'd at least bring it up. Thoughts?

eliperelman:
Fixes #133.

### DIFF
--- a/docs/presets/neutrino-preset-react/README.md
+++ b/docs/presets/neutrino-preset-react/README.md
@@ -121,6 +121,13 @@ manifest.2211d9c1970bbd3c952b.bundle.js    1.41 kB       1  [emitted]  manifest
 
 You can either serve or deploy the contents of this `build` directory as a static site.
 
+## Paths
+
+The `neutrino-preset-web` preset loads assets relative to the path of your application by setting Webpack's
+[`output.publicPath`](https://webpack.js.org/configuration/output/#output-publicpath) to `./`. If you wish to load
+assets instead from a CDN, or if you wish to change to an absolute path for your application, customize your build to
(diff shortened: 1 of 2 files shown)


tj:
hmm is there an example of what this looks like? I just get:

```
TypeError: Path must be a string. Received { publicPath: '/' }
    at assertPath (path.js:7:11)
```

For:

```js
module.exports = {
  options: {
    output: {
      publicPath: '/'
    }
  }
}
```

eliperelman:
Hey @tj, depending on what your preset or RC file looks like:

```js
module.exports = neutrino => {
  neutrino.config.output.publicPath('/');
};
```

Or


```js
module.exports = {
  use: [
    'neutrino-preset-react',
    neutrino => neutrino.config.output.publicPath('/')
  ]
};
```

Something like that should work. Lemme know!

tj:
Thanks man! I'll give that a try

PR #146, opened 2017-03-27
use correct value for package.json config

stoikerty:
This might have been an old way to define the configuration?
Only works when I use `neutrino`, otherwise I get an error:
```
WebpackOptionsValidationError: Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
- configuration misses the property 'entry'.
...
```


eliperelman:
Thanks!

PR #148, opened 2017-03-27
Simplify error logging.

jaridmargolin:
PENDING https://github.com/webpack/webpack/pull/4568. Will also need to bump webpack version across packages.


packages/neutrino/bin/build.js
```diff
@@ -7,10 +7,7 @@ module.exports = (middleware, options) => {
   return build(middleware, options)
     .fork((errors) => {
       spinner.fail('Building project failed');
-      errors.forEach((err) => {
-        console.error(err.stack || err);
-        err.details && console.error(err.details);
-      });
+      errors.forEach(err => console.error(err));
```
eliperelman:
```js
errors.forEach(console.error);
```

packages/neutrino/bin/start.js
```diff
@@ -7,11 +7,7 @@ module.exports = (middleware, options) => {
   return start(middleware, options)
     .fork((errors) => {
       spinner.fail('Building project failed');
-      errors.forEach((err) => {
-        console.error(err.stack || err);
-        err.details && console.error(err.details);
-      });
-
+      errors.forEach(err => console.error(err));
```
eliperelman:
```js
errors.forEach(console.error);
```

packages/neutrino/bin/build.js
jaridmargolin: Problem is, forEach passes `currentValue, index, array`... We only want to log `currentValue`.

packages/neutrino/bin/start.js
jaridmargolin: Problem is, forEach passes currentValue, index, array... We only want to log currentValue.

packages/neutrino/bin/build.js
eliperelman: Ah yes, always forget that. You can solve that by brining in `unary` from ramda and putting console.error in it:

```js
const { unary } = require('ramda');

errors.forEach(unary(console.error));
```

eliperelman: Or if you prefer, put a `logError` or something in utils that has the same thing.

```js
const logError = unary(console.error);

module.exports = {
  // ...
  logError,
  // ...
};
```

eliperelman:
What will this do for exceptions that aren't bubbled from webpack, e.g. can't find a package?

packages/neutrino/bin/build.js
tj: the rambda stuff seems not worth it IMO, adds an extra layer of indirection for a lot of these things, making it harder to reason about if you're not familiar with them all. Some might be more obvious but if I saw `unary()` I'd just be like huh wtf is that

jaridmargolin:
By default, `console.error(new Error('test'))` logs the stacktrace. The problem was that Webpack was adding custom properties to their Error types, and as a result the error was being logged as an object. The PR I submitted to webpack, was to correct the errors so that they log as they were intended.

In short, errors not bubbled from webpack should work just fine.

packages/neutrino/bin/build.js
eliperelman: I agree that just slapping `unary()` right in the `forEach` probably isn't very descriptive, but putting it in utils with a comment should be OK. In fact, adding comments to all our utility functions would be a good thing. ™

eliperelman:
@jaridmargolin any idea if this had shipped in Webpack yet?

jaridmargolin:
Merged but not shipped.

eliperelman:
Looks like Webpack 2.4.1 is out and I think it has your changes.

jaridmargolin:
Rebased and bumped webpack version to 2.4.1.

I did not make any changes to the forEach block discussed above. I think the current implementation `errors.forEach(err => console.error(err));` is extremely readable, and introducing a util would be unnecessary indirection. If you are extremely opinionated here, I CAN make the change.. lmk

eliperelman:
Looks like all the webpack versions were changed by modifying the version number in package.json, but this won't update what we keep locked in yarn. If you want to upgrade a package from the root, use:

```bash
yarn upgrade webpack
yarn deps:upgrade neutrino webpack
yarn deps:upgrade neutrino-middleware-chunk
# etc...
```

This will update package.json and yarn.lock files too.

jaridmargolin:
yarnified 👍

eliperelman:
Looks like there's a merge conflict, could you rebase?

eliperelman:
Released in v5.7.0.

PR #152, opened 2017-03-28
Allow inspecting the API default exports without erroring

eliperelman:


eliperelman:
I tested this in the terminal:

```bash
node -e "console.log({ inspect(depth, opts) { return opts; } });"
```

And received the following output:

```js
{ showHidden: false,
  depth: 2,
  colors: false,
  customInspect: true,
  showProxy: false,
  maxArrayLength: 100,
  breakLength: 60,
  seen: [ [Circular], [1]: [Circular] ],
  stylize: [Function: stylizeNoColor] }
```

It appears Node.js will pass some custom arguments to the inspect function:

https://nodejs.org/dist/latest-v7.x/docs/api/util.html#util_util_inspect_object_options

Basically sniffing that to see if this is run via `util.inspect` or an actual call.

jefffriesen:
This will be great!

PR #160, opened 2017-03-31
Allow setting karma middleware options as overridable

eliperelman:
Right now setting `neutrino.options.karma` will merge the configuration, which can cause duplication problems.


eliperelman:
@mozilla-neutrino/core-contributors looking for a review from someone for sanity.

PR #163, opened 2017-04-03
Update build executable to respect config stats.

jaridmargolin:


packages/neutrino/bin/build.js
```diff
@@ -14,10 +15,12 @@ module.exports = (middleware, options) => {
       process.exit(1);
     }, (stats) => {
       spinner.succeed('Building project completed');
-      console.log(stats.toString({
+      const options = defaultTo({
```
eliperelman: Place a blank line above the `const` declaration.

PR #171, opened 2017-04-07
neutrino-middleware-image-loader: Use svg-url-loader for svg rule

timkelty:
Because base64 encoding SVG content doesn't make sense.

### DIFF
--- a/packages/neutrino-middleware-image-loader/index.js
+++ b/packages/neutrino-middleware-image-loader/index.js
@@ -3,13 +3,14 @@ const merge = require('deepmerge');
 module.exports = ({ config }, options) => {
   const { limit } = merge({ limit: 8192 }, options);
   const urlLoader = require.resolve('url-loader');
+  const svgUrlLoader = require.resolve('svg-url-loader');
 
   config.module
     .rule('svg')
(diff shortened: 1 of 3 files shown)


packages/neutrino-middleware-image-loader/index.js
```diff
@@ -3,13 +3,14 @@ const merge = require('deepmerge');
...
   const urlLoader = require.resolve('url-loader');
+  const svgUrlLoader = require.resolve('svg-url-loader');
 
   config.module
     .rule('svg')
     .test(/\.svg(\?v=\d+\.\d+\.\d+)?$/)
-    .use('url')
-      .loader(urlLoader)
-      .options({ limit, mimetype: 'image/svg+xml' });
+    .use('svgUrl')
```
eliperelman: We can actually probably just leave this `.use` as `url`. It's still a URL "loader", even though the actual loader is different. The `.use` call just defines the ID of the loader.

timkelty: Yep, wondered about that.
Change pushed.

eliperelman:
Released in v5.4.0.

PR #173, opened 2017-04-11
Polyfills

constgen:
The problem is that 'babel-preset-env' doesn't consider defined platforms to filter only necessary polyfills. This change fixes this. Also polyfills are extracted to the 'vendor' chank.
*Before*
before
*After*
after
**From 247KB down to 89KB**


packages/neutrino-preset-web/index.js
```diff
@@ -85,8 +84,10 @@ module.exports = (neutrino) => {
     .when(process.env.NODE_ENV !== 'test', () => neutrino.use(chunk))
     .target('web')
     .context(neutrino.options.root)
+	 .entry('vendor')
```
eliperelman: Dedent this line so it lines up with `.context()`.

packages/neutrino-preset-web/polyfills.js
```diff
@@ -0,0 +1 @@
+import 'babel-polyfill';
```
eliperelman: Add a blank line to the end of this file.

packages/neutrino-preset-web/index.js
```diff
@@ -67,14 +67,13 @@ module.exports = (neutrino) => {
   neutrino.use(htmlTemplate, neutrino.options.html);
   neutrino.use(namedModules);
   neutrino.use(compileLoader, {
-    include: [neutrino.options.source, neutrino.options.tests],
+    include: [neutrino.options.source, neutrino.options.tests, require.resolve('./polyfills.js')],
```
eliperelman: Do we have to add this file to the compilation? Does it work if we switch the polyfills file to use `require`?

```js
require('babel-polyfill');
```

packages/neutrino-preset-web/index.js
```diff
@@ -85,8 +84,10 @@ module.exports = (neutrino) => {
     .when(process.env.NODE_ENV !== 'test', () => neutrino.use(chunk))
     .target('web')
     .context(neutrino.options.root)
+	 .entry('vendor')
+      .add(require.resolve('./polyfills.js'))
```
eliperelman: Removing the polyfills file different, since we can't just remove based on the location of babel-polyfill. I'm not saying this is a bad thing, just different. We should probably add this to the docs, about how to remove the polyfills.

```js
neutrino.config
  .entry('vendor')
    .delete(require.resolve('neutrino-preset-web/polyfills'))
```

eliperelman:
@jaridmargolin any thoughts?

packages/neutrino-preset-web/index.js
constgen: Ok, will do

packages/neutrino-preset-web/index.js
constgen: If you mean to avoid './polyfills.js' it will not work. `babel-preset-env` just want to see some where the importing of 'babel-polyfill'. And it doesn't see it if this file is an entry point itself.

packages/neutrino-preset-web/polyfills.js
constgen: My bad. Usually VSCode helps me with this automatically.

packages/neutrino-preset-web/index.js
constgen: Originally this refactoring had a separate entry point `.entry('polyfills')` which produced one more file chunk. In this case polyfills could be disabled easy by name. But I decided to make changes more soft and joined vendor with polyfills togehter. Now splitting vendor and polyfills seems not a bad idea from the configuration perspective.

jaridmargolin:
@eliperelman @constgen - The idea of creating a separate file that imports is definitely a creative solution that solves the problem.

I am curious if automatically including it in the `vendor` is correct. If I begin utilizing new es features in my index entry, I wouldn't want my vendor bundle cache to be busted.

> Originally this refactoring had a separate entry point .entry('polyfills') which produced one more file chunk. In this case polyfills could be disabled easy by name. But I decided to make changes more soft and joined vendor with polyfills togehter. Now splitting vendor and polyfills seems not a bad idea from the configuration perspective.

I think the the separate `polyfills` entry might be a better solution.

packages/neutrino-preset-web/index.js
eliperelman: I mean, if they want to not have polyfills in their project at all.

eliperelman: They should have the ability to remove it if they wish.

eliperelman:
I'm leaning towards the separate polyfills entrypoint as well.

constgen:
Made necessary changes. Now in my example I also added the 'lodash' library to entries and imported it in the  **index.js** just to see how everything works together. The output is:
after-after

packages/neutrino-preset-web/index.js
constgen: To be honest didn't think about this "include". I see that it will not affect the build until somebody will not import exactly this file `./polyfills.js`.

jaridmargolin:
^ above I was referring to some findings here: https://youtu.be/6m_E-mC0y3Y?list=PLNYkxOF6rcICc687SxHQRuo9TVNOJelSZ&t=248

Its not an exact science but the overall takeaway is, in certain cases, making multiple requests in parallel can be comparable or even faster than requesting a single bundle. I think we could benefit from spending a tiny bit of time profiling performance before making a decision here. 

\* There are definitely additional resources on this topic but I had a difficult time finding them.

constgen:
Additionally when we are in a dev mode (`npm start`) the rebuild will be faster because only index chunk is processed.

eliperelman:
Tested this out locally, everything worked great!

eliperelman:
Sooo, do we need to add polyfills to the chunks? Right now the only defaulted chunks from the chunk middleware are `vendor` and `manifest`. Thoughts?

constgen:
Only if somebody is going to import/require polyfills to his code base. I believe it's not going to happen. But my opinion may be wrong.

eliperelman:
I must've missed something, because when running this while using `async`, I received a message about regeneratorRuntime being missing. Modifying the web middleware to also include `polyfill` in the chunk seems to have fixed it.

constgen:
Interesting. This is definitely related to one of commits in this branch where I removed 'transform-regenerator' from babel preset because it is included in the polyfills on demand.

PR #176, opened 2017-04-11
Adding getters and setters for path options, options for middleware

eliperelman:


timkelty:
Tested setting `source`, `output`, and `entry` with relative paths from middleware - works for me!

I'm not sure how important it is - but it seems there still isn't really a good way to (re)set the defaults, but still allow overrides from package.json.

timkelty:
Thanks! Eagerly awaiting a release :)

eliperelman:
@timkelty it's already released, in v5.5.0! :tada:

timkelty:
BOSS.

PR #181, opened 2017-04-13
Consider case when 'core-js' is located in 'babel-polyfill' root

constgen:
This change fixes situation when there are more than one 'core-js' library versions in the project. In this case 'babel-polyfill' downloads its verison to '/node_modules/babel-polyfill/node_modules/core-js'. That means `babel-preset-env` can't correctly resolve paths to 'core-js' when process `import 'babel-polyfill'`. Babel will try to find it in '/node_modules/core-js' where the wrong version will be located.

There is a particular project example when build fails because of core-js@1.x.x in the root node_modules and core-js@2.x.x in the babel-polyfill
error
paths

This fix also expects that *babel-polyfill* structure is '/libs/index.js'. Hope it will not be changed.


packages/neutrino-preset-web/index.js
```diff
@@ -99,6 +99,10 @@ module.exports = (neutrino) => {
       .chunkFilename('[id].[chunkhash].js')
       .end()
     .resolve
+    .alias
```
eliperelman: Indent this line 2 more spaces.

packages/neutrino-preset-web/index.js
```diff
@@ -99,6 +99,10 @@ module.exports = (neutrino) => {
...
+        // TIP: make sure 2 version of "core-js" always match in package.json and babel-polyfill/package.json
```
eliperelman: - We can remove the "TIP" word here
- Capitalize "make" to "Make"
- Replace "version" with "versions"

packages/neutrino-preset-web/index.js
```diff
@@ -99,6 +99,10 @@ module.exports = (neutrino) => {
...
+        .set('core-js', join(require.resolve('core-js'), '..'))
```
eliperelman: Instead of a join with `..`, let's use `dirname` from path:

```js
const { join, dirname } = require('path');

// ...

.set('core-js', dirname(require.resolve('core-js')))
```

constgen: Agree. Should think of this from the very beginning.

packages/neutrino-preset-web/index.js
constgen: Yes. I thought what is the best here. "Make sure 2 versions of "core-js" always match in package.json and babel-polyfill/package.json" sounds like the next line of code do this. But it actually not. This comment should play a role of a reminder.

Issue #196, opened 2017-05-02
Remove Buffer from being included in web by default


eliperelman:
Adding `Buffer` to the Node.js configuration for the Web preset includes 50KB of polyfills, for a feature that mostly likely won't even be used. We should set this value to `false`, and make this feature opt-in.

barraponto:
I wonder why we set every [default value for the node configuration](https://webpack.js.org/configuration/node/) in our [web-preset](https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/neutrino-preset-web/index.js#L123-L133). Can't we just set the overrides?

barraponto:
Btw, the [documentation](https://webpack.js.org/configuration/node/#other-node-core-libraries) says:
> By default, Webpack will polyfill each library if there is a known polyfill or do nothing if there is not one.

Does it *really* mean that we're polyfilling every node library by default?

eliperelman:
Closed via #201.

eliperelman:
So the defaults are:

```js
node: {
  console: false,
  global: true,
  process: true,
  __filename: "mock",
  __dirname: "mock",
  Buffer: true,
  setImmediate: true
}
```

And yes, we just re-set the defaults. I can't remember why this is, but it's probably no longer needed. Only things marked `true` here would be polyfilled by default.

PR #201, opened 2017-05-08
Buffer: true; -> Buffer: false

aretecode:
remove not-needed polyfill
https://webpack.js.org/configuration/node/
https://github.com/mozilla-neutrino/neutrino-dev/issues/196


constgen:
The task is not finished yet. `Buffer` is still included in ESlint rules 
https://github.com/aretecode/neutrino-dev/blob/8901d3ef30d2b796cfe92509ab011077a69ff899/packages/neutrino-preset-web/index.js#L136

Issue #197, opened 2017-05-04
Switch to fast-async by default instead of regenerator


eliperelman:
I've been using fast-async with [neutrino-preset-react-components](https://github.com/eliperelman/neutrino-preset-react-components/blob/master/index.js#L128-L137) and have been pretty happy. Using the `spec: true` option, async is compiled to promises, no generators. We may have to add additional options to consolidate the runtime for a bundle. Any thoughts?

eliperelman:
Resolved in #202.

PR #202, opened 2017-05-09
V6 dev fast async

aretecode:
https://github.com/babel/babel-preset-env
https://github.com/mozilla-neutrino/neutrino-dev/issues/197

- node version to current LTS 
- add babel-fast-async to neutrino-preset-web
- add babel-fast-async to neutrino-preset-node
- yarn lock files


packages/neutrino-preset-node/index.js
```diff
@@ -25,7 +25,7 @@ module.exports = (neutrino) => {
     Object.assign(neutrino.options, {
       compile: {
         targets: {
-          node: 6.9
+          node: 6.10
```
eliperelman: Since this is a number, it will evaluate to `6.1`. This probably needs to be a string now `'6.10'`.

packages/neutrino-preset-node/index.js
```diff
@@ -35,11 +35,13 @@ module.exports = (neutrino) => {
   neutrino.use(compile, {
     include: [neutrino.options.source, neutrino.options.tests],
     babel: {
-      plugins: [require.resolve('babel-plugin-dynamic-import-node')],
+      plugins: [[require.resolve('fast-async'), { spec: true }], require.resolve('babel-plugin-dynamic-import-node')],
```
eliperelman: Does using this with `spec: true` put the nodent runtime in every file that uses async?

packages/neutrino-preset-node/index.js
```diff
@@ -35,11 +35,13 @@ module.exports = (neutrino) => {
...
       presets: [
         [require.resolve('babel-preset-env'), {
           modules: false,
-          targets: neutrino.options.compile.targets
+          useBuiltIns: false,
```
eliperelman: Why switch off `useBuiltIns`?

packages/neutrino-preset-node/package.json
```diff
@@ -15,6 +15,7 @@
   "homepage": "https://neutrino.js.org",
   "bugs": "https://github.com/mozilla-neutrino/neutrino-dev/issues",
   "dependencies": {
+    "fast-async": "6.2.2",
```
eliperelman: Not sure how this dependency was added without semver ranges. Did you add this via `yarn deps:add neutrino-preset-node fast-async`? Do you have `--exact` set somewhere?

aretecode: @eliperelman added in pkg json, will add minor ^

packages/neutrino-preset-node/index.js
aretecode: https://github.com/mozilla-neutrino/neutrino-dev/issues/197 @eliperelman

packages/neutrino-preset-node/index.js
aretecode: https://github.com/eliperelman/neutrino-preset-react-components/blob/master/index.js#L128-L137

packages/neutrino-preset-web/index.js
```diff
@@ -67,15 +67,18 @@ module.exports = (neutrino) => {
...
     include: [neutrino.options.source, neutrino.options.tests, require.resolve('./polyfills.js')],
     babel: {
-      plugins: [require.resolve('babel-plugin-syntax-dynamic-import')],
+      plugins: [[require.resolve('fast-async'), { spec: true }], require.resolve('babel-plugin-syntax-dynamic-import')],
       presets: [
         [require.resolve('babel-preset-env'), {
           modules: false,
-          useBuiltIns: true,
-          targets: neutrino.options.compile.targets
+          useBuiltIns: false,
```
constgen: Still can't understand why `useBuiltIns: false,`. Especially in the Web Preset.

packages/neutrino-preset-node/index.js
eliperelman: `useBuiltIns` is on there because it's generating a library, while this preset is for application code. Having `useBuiltIns` will let babel-preset-env reduce babel-polyfill.

packages/neutrino-preset-web/index.js
aretecode: @eliperelman @constgen updated

packages/neutrino-preset-node/index.js
eliperelman: This is what I'm concerned about:

https://github.com/MatAtBread/fast-async#runtimepattern

packages/neutrino-preset-web/index.js
```diff
@@ -67,15 +67,18 @@ module.exports = (neutrino) => {
   neutrino.use(imageLoader);
   neutrino.use(htmlTemplate, neutrino.options.html);
   neutrino.use(namedModules);
+
```
eliperelman: We can eliminate these comments since the code below is already explanatory.

packages/neutrino-preset-web/index.js
```diff
@@ -67,15 +67,18 @@ module.exports = (neutrino) => {
...
+  // add babel-preset-env with fast async, & dynamic import support
   neutrino.use(compileLoader, {
     include: [neutrino.options.source, neutrino.options.tests, require.resolve('./polyfills.js')],
     babel: {
-      plugins: [require.resolve('babel-plugin-syntax-dynamic-import')],
+      plugins: [[require.resolve('fast-async'), { spec: true }], require.resolve('babel-plugin-syntax-dynamic-import')],
```
eliperelman: Let's put these on separate lines for readability's sake:

```js
plugins: [
  [require.resolve('fast-async'), { spec: true }],
  require.resolve('babel-plugin-syntax-dynamic-import')
],
```

packages/neutrino-preset-web/index.js
aretecode: you just hate comments ;-P

packages/neutrino-preset-node/index.js
aretecode: @eliperelman so you want it `false` ?

eliperelman: I don't know what value I want. What I want:

- Using async functions without generators (no requirement on babel-polyfill, transform-regenerator, etc.)
- Not having a nodent runtime in every file that uses async functions

aretecode: @eliperelman so to confirm, we should add only to entry, right? what if there are multiple entries? what if entry is a function?

eliperelman: By default we control entry. If they override entry afterwards its fine, and overriding it before will act funny anyway. I think it should be ok to add only to `index` entry.

aretecode: https://github.com/MatAtBread/fast-async#runtimepattern

aretecode: @eliperelman https://github.com/mozilla-neutrino/neutrino-dev/pull/202/commits/354764382cd9681a31f0ad4496b4d1d0b654c208 how's this?

packages/neutrino-preset-web/index.js
```diff
@@ -67,15 +67,25 @@ module.exports = (neutrino) => {
   neutrino.use(imageLoader);
   neutrino.use(htmlTemplate, neutrino.options.html);
   neutrino.use(namedModules);
+
+  // turn string into regex for fast-async
+  // e.g., /user/src/index.js (match: true) /user/src/eh.js (match: false)
+  const entryPattern = neutrino.options.entry.replace(/[|\\{}()[\]^$+*?.]/g, '\\$&');
+
+  // add babel-preset-env with fast async, & dynamic import support
```
eliperelman: Remove this comment.

packages/neutrino-preset-web/index.js
```diff
@@ -67,15 +67,25 @@ module.exports = (neutrino) => {
   neutrino.use(imageLoader);
   neutrino.use(htmlTemplate, neutrino.options.html);
   neutrino.use(namedModules);
+
+  // turn string into regex for fast-async
+  // e.g., /user/src/index.js (match: true) /user/src/eh.js (match: false)
+  const entryPattern = neutrino.options.entry.replace(/[|\\{}()[\]^$+*?.]/g, '\\$&');
```
eliperelman: Don't we need this also in the Node preset?

packages/neutrino-preset-node/index.js
eliperelman: Let's put each plugin on its own line.

aretecode:
@eliperelman updated

packages/neutrino-preset-web/index.js
```diff
@@ -67,15 +67,22 @@ module.exports = (neutrino) => {
   neutrino.use(imageLoader);
   neutrino.use(htmlTemplate, neutrino.options.html);
   neutrino.use(namedModules);
+
+  neutrino.config.entry('index').prepend(require.resolve('nodent-runtime'));
```
eliperelman: We are already creating the `index` entry in this file on line 101, so you can change it to:

```js
.entry('index')
  .add(require.resolve('nodent-runtime'))
  .add(neutrino.options.entry)
```

packages/neutrino-preset-node/index.js
```diff
@@ -25,21 +25,28 @@ module.exports = (neutrino) => {
...
       compile: {
         targets: {
-          node: 6.9
+          node: '6.10'
         }
       }
     });
   }
 
+  neutrino.config.entry('index').prepend(require.resolve('nodent-runtime'));
```
eliperelman: Remove this and change it on line 72:

```js
.entry('index')
  .add(require.resolve('nodent-runtime'))
  .add(neutrino.options.entry)
```

PR #209, opened 2017-05-19
Fix error in test usage example

lukechilds:


eliperelman:
Good catch, thank you!

PR #213, opened 2017-05-21
Bugfix/react set state

constgen:
As community suggests I added `transform-es2015-classes` babel preset in development mode. This means that some legacy browsers (like IE10) will support ES class syntax in dev but not in prod which is not expected. But it should not be a problem. If it launched in old browser during development than usually such browser is listed in `browserList`. Also the price of always include `transform-es2015-classes` into build may be too high.


lielran:
merge?

packages/neutrino-preset-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "neutrino-preset-react",
-  "version": "5.8.0",
+  "version": "6.0.0",
```
eliperelman: Let's roll this back to 5.8.1 and ship it before v6.

packages/neutrino-preset-react/package.json
```diff
@@ -21,9 +21,10 @@
     "eslint-plugin-react": "^6.10.3",
     "neutrino-middleware-loader-merge": "^5.8.0",
     "neutrino-preset-web": "^5.8.0",
-    "react-hot-loader": "^3.0.0-beta.6"
+    "react-hot-loader": "^3.0.0-beta.6",
+    "deepmerge": "^1.3.2"
```
eliperelman: We don't need deepmerge as we already have `neutrino-middleware-loader-merge` here.

packages/neutrino-preset-react/index.js
```diff
@@ -1,5 +1,6 @@
 const loaderMerge = require('neutrino-middleware-loader-merge');
 const web = require('neutrino-preset-web');
+const merge = require('deepmerge');
```
eliperelman: Remove this.

packages/neutrino-preset-react/index.js
```diff
@@ -39,8 +40,19 @@ module.exports = (neutrino) => {
       'react/lib/ReactContext': 'window'
     })
     .when(process.env.NODE_ENV === 'development', config => config
+      .module
```
eliperelman: We could combine this with the previous babel loader merge at the beginning of this preset:

```diff
  neutrino.use(loaderMerge('compile', 'babel'), {
    presets: [require.resolve('babel-preset-react')],
    plugins: [require.resolve('babel-plugin-transform-object-rest-spread')],
    env: {
      development: {
        plugins: [
          require.resolve('react-hot-loader/babel'),
+         require.resolve('babel-plugin-transform-es2015-classes')
        ]
      }
    }
  });
```

packages/neutrino-preset-react/README.md
```diff
@@ -22,7 +22,7 @@
 
 - Node.js v6.9+
 - Yarn or npm client
-- Neutrino v5
+- Neutrino v6
```
eliperelman: Let's switch this to v5.

packages/neutrino-preset-react/index.js
constgen: I suppose it should be done only in the "development" and should not increase the bundle size without a necessary

PR #215, opened 2017-05-23
add option for node debugging

### LINKED ISSUE #208: Feature request: Better debugging capabilities for node presets
I was wondering how would you go about debugging `neutrino-node` apps? 

What I'm hoping for is a possibility to be able to start the node server with the fairly new `--inspect` flag. That would allow for a great debugging experiences, as outlined by [Paul Irish](https://medium.com/@paul_irish/debugging-node-js-nightlies-with-chrome-devtools-7c4a1b95ae27). 
However I was not able to set that up using your node preset.

It should be possible to implement that in a fairly straightforward way, since the [start-server-webpack-plugin](https://github.com/ericclemmons/start-server-webpack-plugin), which you use in `start-server` middleware, allows you to do that with an additional property `nodeArgs: ['--inspect']`.

I'm not sure about the best way to toggle debugging behaviour, since you use the `-- inspect` flag to print the webpack conf.

If there is another way to start the underlying node server in debugging mode, that could be an opportunity to improve the docs.

kriswep:
As discussed in #208 this pr adds a possibility to start the node server with `--inspect` option.
This is mostly relevant for neutrino node apps.

A `--debug` flag was added to neutrino, to which `neutrino-middleware-start-server` reacts and starts the `start-server-webpack-plugin` with the inspect flag. For that to work I had to update `neutrino-middleware-start-server` to v2.2.0.

Users could add a debug script to package.json like this:
```JavaScript
  "scripts": {
    "dev": "neutrino start --use neutrino-preset-node --use neutrino-middleware-start-server --debug",
    "build": "neutrino build --use neutrino-preset-node"
  }
```
Hope that helps to add this feature and improve debugging possibilities for neutrino node apps.
 
fixes #208

### DIFF
--- a/packages/neutrino-middleware-start-server/index.js
+++ b/packages/neutrino-middleware-start-server/index.js
@@ -1,13 +1,8 @@
 const StartServerPlugin = require('start-server-webpack-plugin');
 
-module.exports = (neutrino, options) => {
-  if (neutrino.options.args && neutrino.options.args.debug) {
-    // start server with --inspect
-    return neutrino.config.plugin('start-server').use(StartServerPlugin, [{
-      name: options.name,
-      nodeArgs: ['--inspect']
-    }]);
-  }
-  // normal start-server
-  return neutrino.config.plugin('start-server').use(StartServerPlugin, [options.name]);
-};
+module.exports = (neutrino, options) => neutrino.config
+  .plugin('start-server')
+  .use(StartServerPlugin, [{
+    name: options.name,
+    nodeArgs: neutrino.options.debug ? ['--inspect'] : []
+  }]);
--- a/packages/neutrino-middleware-start-server/index.js
+++ b/packages/neutrino-middleware-start-server/index.js
@@ -1,3 +1,13 @@
 const StartServerPlugin = require('start-server-webpack-plugin');
 
-module.exports = ({ config }, options) => config.plugin('start-server').use(StartServerPlugin, [options.name]);
+module.exports = (neutrino, options) => {
+  if (neutrino.options.args && neutrino.options.args.debug) {
+    // start server with --inspect
+    return neutrino.config.plugin('start-server').use(StartServerPlugin, [{
+      name: options.name,
+      nodeArgs: ['--inspect']
+    }]);
+  }
+  // normal start-server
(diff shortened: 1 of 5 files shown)


packages/neutrino/bin/neutrino
```diff
@@ -64,10 +70,12 @@ const normalizeOptions = (args) => {
   const pkg = getPackageJson();
   const options = pathOr({}, ['neutrino', 'options'], pkg);
   const env = args.inspect ? getNodeEnv(args._[0], args.env) : args.env;
+  const debug = args.debug ? args.debug : options.debug;
```
eliperelman: This variable is only used to define an object property, so we can inline it into the object itself.

```js
debug: args.debug || options.debug
```

packages/neutrino/src/api.js
```diff
@@ -66,6 +67,14 @@ const getOptions = (options = {}) => {
       set(value) {
         entry = defaultTo('index', value);
       }
+    },
+    debug: {
```
eliperelman: This section isn't needed. We have the getter/setters for the path options since we need to normalize the paths when they are overridden. `debug` isn't a path value, only a Boolean, so this can be removed.

packages/neutrino-middleware-start-server/index.js
```diff
@@ -1,3 +1,13 @@
 const StartServerPlugin = require('start-server-webpack-plugin');
 
-module.exports = ({ config }, options) => config.plugin('start-server').use(StartServerPlugin, [options.name]);
+module.exports = (neutrino, options) => {
```
eliperelman: This could probably be shortened to:

```js
module.exports = (neutrino, options) => neutrino.config
  .plugin('start-server')
    .use(StartServerPlugin, [{
      name: options.name,
      nodeArgs: neutrino.options.debug ? ['--inspect'] : []
    }]);
```

Thoughts?

kriswep:
Integrated the requested changes. 
Looking forward to hear from you again...

packages/neutrino-middleware-start-server/index.js
kriswep: Yes that worked, after I removed the getter/setters for the path options in neutrino api.js

eliperelman:
Released in v5.9.0!

eliperelman:
Thank you!

PR #217, opened 2017-06-01
Use string version of specifying target

marcusstenbeck:
I get the following, very annoying, warning when building. This change fixes it.

```
Warning, the following targets are using a decimal version:

  node: 6.9

We recommend using a string for minor/patch versions to avoid numbers like 6.10
getting parsed as 6.1, which can lead to unexpected behavior.
```


eliperelman:
Published to npm at v5.9.1.

PR #218, opened 2017-06-01
Is it possible to support ES6+ ?

heibor:
Is it possible to support ES6+ ?
I'm using neutrino-preset-react, when I use some ES6+ features, such as:

```
class Video extends React.Component {
  static defaultProps = {
    autoPlay: false,
    maxLoops: 10,
  }
  static propTypes = {
    autoPlay: React.PropTypes.bool.isRequired,
    maxLoops: React.PropTypes.number.isRequired,
    posterFrameSrc: React.PropTypes.string.isRequired,
    videoSrc: React.PropTypes.string.isRequired,
  }
  state = {
    loopsRemaining: this.props.maxLoops,
  }
}
```

I will get a syntax error. I want to ask whether it's possible to use ES6+ in the future.


eliperelman:
@heibor right now the react preset has the following Babel plugins and presets available:

- syntax-dynamic-import
- react
- transform-object-rest-spread

We have been shooting for the lowest common denominator in terms of standards support. If you want to propose using some additional presets or plugins in the core, feel free to file an issue or join our Slack channel and we can discuss it.

heibor:
I use:


```
npm install --save-dev babel-plugin-transform-class-properties
```

edit the file: ```.babelrc```

```
{
  "plugins": ["transform-class-properties"]
}
```

Now, the ES6+ syntax supporting has been down. I think maybe the babel-plugin-transform-class-properties can be integrated into the neutrino-preset-react.

eliperelman:
I agree, and think it's common enough with react users to warrant discussion. Could you file an issue about adding this plugin to the react preset?

heibor:
Ok, I've filed the issue #228, I hope the ES6+ syntax can bring more excited developing experience for React.

PR #223, opened 2017-06-01
Caching based on hashed and inline manifest

eliperelman:


eliperelman:
@emorley would you be interested in taking a look at this?

edmorley:
There's finally a PR open based on the best practices from that blog post etc - perhaps there might be some further wins from that when it lands:
webpack/webpack.js.org/pull/1436

Issue #27, opened 2017-02-17
Offline/PWA support


eliperelman:
Using https://github.com/NekR/offline-plugin this could be pretty straightforward, and a win for everyone.

sotayamashita:
It also useful stuff. https://github.com/webpack/webpack-pwa

eliperelman:
It would be *really nice* to have out-of-the-box offline and PWA build support, maybe a combination of sane defaults with some conventions. Let's plan this for v6, but we can push it back more if it's just too much right now.

cc: @addyosmani in case he has any nice links for some generic webpack configuration we can borrow for PWA stuffs.

eliperelman:
Some good resources here as well: https://github.com/vuejs/pwa/tree/prototype/template

barraponto:
I'd like this to be a preset, so web-preset users don't see any change until they opt-in to this.

eliperelman:
Have a first draft landed in the v6-dev branch. We can iterate on it more in the future.

constgen:
Currently we use Hot Module Replacement only in development mode. This is useful for development purposes. But I see some profit for enabling it in production in case we build a PWA. HMR is designed to work in production also to help with seamless deploy when changes are pushed to the client side without a page refresh

PR #227, opened 2017-06-01
PWA middleware

eliperelman:
Draft of proposed PWA middleware. The goal of this is to be able to drop into a web project and get SW caching for free with no upfront configuration.


eliperelman:
This would resolve #27.

constgen:
Can we try to hot reload the application instead of page refresh when new version of resources arrived?

eliperelman:
Hot reload is only enabled in development, and service worker is only enabled in production. Not to mention that HMR is optional, and the concerns are separate from those of SW. In the core case, they aren't related.

eliperelman:
@developit thanks for the 👍!

PR #230, opened 2017-06-02
Allow environment-specific configuration overrides

### LINKED ISSUE #185: Allow environment-specific configuration overrides
Using the package.json overrides, all configuration is added to every environment. If you want to override something that only exists in a certain environment, this will blow up. We should support making overrides on an environment or command-specific basis.

Environment-specific:

```js
{
  "neutrino": {
    "env": {
      "production": {
        "config": {
          // e.g. edit minify plugin options
        }
      }
    }
  }
}
```

Command-specific:

```js
{
  "neutrino": {
    "command": {
      "build": {
        "config": {
          // e.g. edit minify plugin options
        }
      }
    }
  }
}
```

helfi92:
Resolves #185

### DIFF
--- a/packages/neutrino/bin/neutrino
+++ b/packages/neutrino/bin/neutrino
@@ -73,6 +73,7 @@ const normalizeOptions = (args) => {
 
   return Object.assign(options, {
     config: pathOr({}, ['neutrino', 'config'], pkg),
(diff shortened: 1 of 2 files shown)


eliperelman:
Note that this will probably change completely once we have `.neutrinorc.js` and object-or-function as middleware.

Issue #199, opened 2017-05-04
Remove package.json configuration in favor of an rc file


eliperelman:
Using package.json for overrides can be quite limiting. I'd like to consider having a `neutrinorc.js` (name negotiable of course) file which is a merging of "simple" and "advanced" configuration overrides. We could keep pretty much everything the same, with the benefit that simpler configuration could utilize JS instead of being limited to JSON.

It may seem odd, but we could vary how this file works based on the export.

```js
module.exports = {
  use: [alpha, beta, gamma],
  options: {...},
  config: {...}
};
```

```js
module.exports = neutrino => {
  // ...
};
```

Basically if the exports are an object, we could treat them very similarly to how we do with the package.json content. If the exports are a function, it can just be a piece of middleware to load. The tricky part is determining precedence. Does the rc file take precedence over `--use` middleware, or vice versa?

Going this route we eliminate the need for users to manually specify in their package.json that they are using an override file, since we can pick it up automatically by name.

Thoughts?

timkelty:
Love this idea. If no one has started work I may pitch in.

@eliperelman Especially given that Neutrino/Webpack chain seems to have be built to _avoid_ big configuration objects, I'd actually be in favor dropping the object/package.json style config all together.

My preference would be to _only_ support this:
```
module.exports = neutrino => {
  neutrino.options = {
      source: './app/assets',
      output: './public/assets',
      entry: './js/index.js',
  };

  neutrino.use(require('neutrino-preset-web');
};
```
in favor of:
```
module.exports = {
  options: {
      source: './app/assets',
      output: './public/assets',
      entry: './js/index.js',
  },
  use: [
    'neutrino-preset-web'
  ],
};
```

Seems just as easy, and more explicit.

It would also clear up where MWs/presets should get thier options from (passed via `.use` vs `neutrino.options.myArbitraryMiddlewareProp.myOption`. If they were always called with `.use`, then they can easily accept options.

Thought on `neutrinorc.js`vs `.neutrinorc.js`?

constgen:
My vision is:
```js
module.exports = {
  use: [alpha, beta, 'gamma', function(neutrino){ ... }],
  options: {...},
  config: {...}
};
```
This is similar to what we have in `package.json` now but with JS runtime possibilities. I think `use` may consume module, module name or middleware function (which is the same module). In my opinion this approach keeps the start steps simple as they are now and gives the more consumer friendly advanced customization.

[...]

Issue #165, opened 2017-04-05
Support passing custom parameters to Neutrino via the CLI


edmorley:
A typical way of passing options to commands would be to define the package.json script entry like so:
`NODE_ENV=production FOO=bar webpack -p --progress`

Unfortunately on Windows environment variables must instead be set like `set NODE_ENV=production`, so to write a cross-platform compatible package.json, previously the recommended solution was to use cross-env:
`cross-env NODE_ENV=production FOO=bar webpack -p --progress`

However this adds yet another dependency and boilerplate to package.json.

webpack 2 now supports an `--env` flag for passing of arbitrary options, so the above could be replaced with:
`webpack -p --env.production --env.foo=bar --progress`

...which is now the recommended way to pass options in a cross-platform compatible way, according to the cross-env author. See the second half of this guide (search for "cross-env"):
https://blog.flennik.com/the-fine-art-of-the-webpack-2-config-dc4d19d7f172#.297u8iuz1

However currently Neutrino doesn't support passing arbitrary arguments (at least according to the usage text and/or CLI docs).

I don't have a strong preference as to what format the flags should take, but I imagine it might be more intuitive if Neutrino just copied webpack 2's syntax to save having to remember another.

Webpack docs:
https://webpack.js.org/guides/environment-variables/
https://webpack.js.org/configuration/configuration-types/

edmorley:
So unfortunately Neutrino already supports an `--env` parameter, however it's used to set the value of `NODE_ENV`:
https://github.com/mozilla-neutrino/neutrino-dev/blob/3f56635381131369279ea54c931c8277ef39aca5/packages/neutrino/bin/neutrino#L28-L32
https://github.com/mozilla-neutrino/neutrino-dev/blob/3f56635381131369279ea54c931c8277ef39aca5/packages/neutrino/bin/neutrino#L66

I still think it would be preferable to match webpack's parameter scheme, though this would mean having to support the old `--env=production` form as well as the new `--env.foo=bar` form.

For reference, webpack's env parameter handling can be found here:
https://github.com/webpack/webpack/blob/38f9686a5d11a15d5d14c8f9008093da0f5bb5b3/bin/config-yargs.js#L23-L26
https://github.com/webpack/webpack/blob/38f9686a5d11a15d5d14c8f9008093da0f5bb5b3/bin/convert-argv.js#L100-L103

edmorley:
[...]

PR #233, opened 2017-06-04
Move from package.json configuration to .neutrinorc.js, allow CLI options

eliperelman:
Everything seems to be working so far. I have a local node, react, and web project with the following `.neutrinorc.js` files in each respectively:

```js
// node project
module.exports = neutrino => {
  neutrino.use('neutrino-preset-node', {
    babel: {
      plugins: [require.resolve('babel-plugin-transform-react-jsx')]
    }
  });
  neutrino.use('neutrino-preset-mocha');
};
```

```js
// react project
module.exports = {
  use: [
    'neutrino-preset-react',
    'neutrino-preset-jest'
  ],
  env: {
    production: {
      options: {
        output: 'dist'
      },
      use: [
        neutrino => neutrino.config.merge({
          entry: {
            vendor: [
              'react',
              'react-dom'
            ]
          },
          "module": {
            "rule": {
              "style": {
                "use": {
                  "css": {
                    "options": {
                      "modules": true
                    }
                  }
                }
              }
            }
          }
        })
      ]
    }
  }
};
```

```js
// web project
module.exports = {
  use: [
    'neutrino-preset-airbnb-base',
    ['neutrino-preset-web', { devServer: { port: 5769, host: '192.168.0.107' } }],
    'neutrino-preset-karma'
  ]
};
```


eliperelman:
I've also removed all environment variables at the moment. This means that all IP lookups were removed and must be manually passed into the middleware in `.neutrinorc.js`.

eliperelman:
This will resolve #199.

eliperelman:
I'm going to change the `.env` layout to accommodate accepting command line options and a complex env structure.

eliperelman:
This PR now also addresses #165. Here is an updated react example based on the one above:

```js
// react project
module.exports = {
  use: [
    'neutrino-preset-react',
    'neutrino-preset-jest'
  ],
  env: {
    NODE_ENV: {
      production: {
        options: {
          output: 'dist'
        },
        use: [
          neutrino => neutrino.config.merge({ entry: { vendor: ['react','react-dom'] } })
        ]
      }
    }
  }
};
```

`env` is now an object instead of a single value. This means you can configure based on arbitrary envs passed to Neutrino:

```js
module.exports = {
  env: {
    ALPHA: {
      // The following middleware is used when `neutrino.options.env.ALPHA === 'beta'`
      beta: { ... }
    }
  }
};
```

You can now also set Neutrino options and environment variables from the command line:

```sh
neutrino start --options.output dist --options.env.NODE_ENV production --options.env.ALPHA beta
```

eliperelman:
I've added back the ability to configure the dev server from the command line, although with different options now, no environment variables:

```sh
neutrino start --options.host something.local --options.port 3000 --options.https
```

Nested booleans don't work out of the box with strings, so setting something to "false" will take manual parsing in middleware, e.g. you can't do this without parsing:

```sh
neutrino start --options.https false
```

Other than that, it works well!

constgen:
Why not this layout for the "use"?
```js
module.exports = ({
  use: [
    'alpha',
    {
      use: 'neutrino-preset-web',
      options: {
        devServer: {
          port: 3000
        }
    },
    neutrino => {
      // whatever
    }
  ]
});
```

rodgobbi:
Should the CLI options be working on version `6.2.1`?
I've tried to inject an env variable via CLI but it always returns `undefined`.

eliperelman:
@rodgobbi it's possible there is a bug in that version. Could you provide a simplified test case repo to reproduce?

rodgobbi:
@eliperelman thank you for your attention. I just pass in to the `start` script the command: `neutrino start --options.env.ALPHA test` and it prints `undefined` the line `console.log(process.env.ALPHA)`.

eliperelman:
@rodgobbi is the `console.log` in the application, or a file being `use`d by Neutrino?

eliperelman:
I just downloaded Neutrino v6.2.1 and ran used the following `test.js` file:

```js
module.exports = (neutrino) => {
  console.log('\n\n', process.env.ALPHA, '\n\n');
};
```

```bash
❯ yarn neutrino start --options.env.ALPHA beta --use test.js
yarn run v1.5.1
warning package.json: No license field
$ /private/var/folders/dl/9lrzd5110zlft676qdpr7xn40000gp/T/repro/node_modules/.bin/neutrino start --options.env.ALPHA beta --use test.js
⠋ Building project

 beta
```

So this part seems to work for me

eliperelman:
You would need to be using some env middleware. If you have some code I could view, I could help give you a more definitive answer.

rodgobbi:
I understand, I thought I could pass it directly, it won't be possible for me to show any code today, but I'll try the Middleware tomorrow, or soon as I can. Thanks a lot for the help.

eliperelman:
@rodgobbi sure thing. The reason we rely on middleware for you to be explicit about what env vars to inject comes down to security. If we placed all env vars into the bundle, it unintentionally inject private information. Let me know if I can help you out more tomorrow.

rodgobbi:
@eliperelman I've installed the `@neutrinojs/env` and I don't know why, it only works if I put it in the end of the `use` array, after the `neutrino-preset-react`, but still, I got what I needed. Before putting it to work (when I put it on the beginning of the `use` array), I was getting really frustrated hehehhe, if you have any idea why did this happen, I'm curious to hear. And thanks a lot for the fast and good support!

eliperelman:
@rodgobbi that's because the react preset internally also uses this preset. If you put it before, the react preset values will override it.

PR #236, opened 2017-06-05
Custom commands 😍

eliperelman:
Custom CLI commands!

I'm actually in awe of how easy this was. The eslint middleware is the first thing to utilize this:

```sh
neutrino lint
```

Any middleware can register a command with a name, and a function that **returns a future, a promise, or a value**. If that command is used, all the middleware is added as normal and the command is run during the normal event lifecycle. If there is a resolved value from the function, it will be output to the console, and any errors thrown/rejected/etc will follow the same flow as the rest of our error handling.

What does this mean?

- No urgent need for `.eslintrc.js`
- ESLint can run separately from the build, and doesn't need weird ignores, and uses the same config as Neutrino
- No extra boilerplate
- Middleware can register commands

😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍


jaridmargolin:
Only have time to review your description but.... 😍

PR #240, opened 2017-06-08
Upgrading outdated dependencies

eliperelman:


edmorley:
Is the switch from a tilde-specified webpack to an explicit version intentional?

eliperelman:
@edmorley no, and I'm not sure how that happened. Good catch.

PR #243, opened 2017-06-09
Hotfix/neutrino preset jest/relative alias handling

l2silver:
Check for relative vs absolute path
Check for wildcard path resolution

### DIFF
--- a/packages/neutrino-preset-jest/src/index.js
+++ b/packages/neutrino-preset-jest/src/index.js
@@ -8,12 +8,12 @@ const loaderMerge = require('neutrino-middleware-loader-merge');
 
 function getFinalPath(path) {
   if (path[0] === '/') {
-    return path
+    return path;
   }
   if (path[0] === '.') {
-    return join('<rootDir>', path)
+    return join('<rootDir>', path);
   }
-  return join('<rootDir>', 'node_modules', path)
+  return join('<rootDir>', 'node_modules', path);
 }
 
 function normalizeJestOptions(jestOptions, config, args) {
@@ -21,9 +21,9 @@ function normalizeJestOptions(jestOptions, config, args) {
   const aliases = config.resolve.alias.entries() || {};
   Object
     .keys(aliases)
-    .map(key => {
-      const finalPath = getFinalPath(aliases[key])
-      return Object.assign(options.moduleNameMapper, { [`${key}(.*)`]: `${finalPath}$1`})
+    .map((key) => {
+      const finalPath = getFinalPath(aliases[key]);
+      return Object.assign(options.moduleNameMapper, { [`${key}(.*)`]: `${finalPath}$1` });
     });
 
   options.moduleFileExtensions = [...new Set([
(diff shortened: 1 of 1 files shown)


packages/neutrino-preset-jest/src/index.js
```diff
@@ -6,13 +6,25 @@ const { tmpdir } = require('os');
 const clone = require('lodash.clonedeep');
 const loaderMerge = require('neutrino-middleware-loader-merge');
 
+function getFinalPath(path) {
+  if (path[0] === '/') {
```
eliperelman: It looks like this is trying to determine if a path is relative or absolute. Is that the case? Could you explain what this is supposed to achieve?

packages/neutrino-preset-jest/src/index.js
```diff
@@ -6,13 +6,25 @@ const { tmpdir } = require('os');
...
+}
+
 function normalizeJestOptions(jestOptions, config, args) {
   const options = clone(jestOptions);
   const aliases = config.resolve.alias.entries() || {};
-
   Object
     .keys(aliases)
-    .map(key => Object.assign(options.moduleNameMapper, { [key]: join('<rootDir>', aliases[key]) }));
+    .map((key) => {
```
eliperelman: Since final path is only used one place here, we could re-write this as:

```js
.map(key => Object.assign(options.moduleNameMapper, {
  [`${key}(.*)`]: `${getFinalPath(aliases[key])}$1`
}));
```

eliperelman:
Could you elaborate on what the problem is, and how your PR fixes this? Thanks!

packages/neutrino-preset-jest/src/index.js
l2silver: It all depends on how you setup your aliases. If you use relative paths, then going off the <rootDir> will work fine, but if you use absolute paths, <rootDir>/absolutePath doesn't really work. Likewise, if you've aliased a module in node_modules, you'll need to explicitly same rootDir/node_modules/moduleName. Unless there's something off with my setup, loading aliases modules was probably never working...

packages/neutrino-preset-jest/src/index.js
l2silver: Good point, I'll make the change

packages/neutrino-preset-jest/src/index.js
eliperelman: OK, I think I get it now. So to make this a little more portable, let's bring in `isAbsolute` from path and try this:

```js
const getFinalPath = path => {
  if (isAbsolute(path)) {
    return path;
  }

  return path.startsWith('.') ?
    join('<rootDir>', path) :
    join('<rootDir>', 'node_modules', path);
};
```

packages/neutrino-preset-jest/src/index.js
```diff
@@ -1,18 +1,29 @@
...
 const merge = require('deepmerge');
 const { tmpdir } = require('os');
 const clone = require('lodash.clonedeep');
 const loaderMerge = require('neutrino-middleware-loader-merge');
 
+function getFinalPath(path) {
+  if (isAbsolute(path)) {
+    return path;
+  }
+  return path.startsWith('.') ?
```
eliperelman: I know it's a nit, but could you put a blank line before the return?

PR #250, opened 2017-06-13
Document conditional configuration

### LINKED ISSUE #180: Document that some plugins and rules are only available in certain environments
For example, the Web preset only exposes a minify plugin during `production`, leading to issues when trying to modify its settings, but blows up during `development`. Document that these configs should be modified conditionally based on the environment.

helfi92:
Resolves #180

### DIFF
--- a/docs/customization/advanced.md
+++ b/docs/customization/advanced.md
@@ -225,7 +225,7 @@ here, please refer to its documentation for all ways you can modify a config ins
 ### Conditional configuration
 
 Some plugins and rules are only available in certain environments. For example, the Web preset only exposes a minify
-plugin during production, leading to issues when trying to modify its settings, but blows up during development.
+plugin during production, leading to issues when trying to modify its settings, but throws an exception during development.
 
 _Example: Access the entries object of the `minify` plugin when using the Web preset._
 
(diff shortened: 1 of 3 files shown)


docs/customization/advanced.md
```diff
@@ -221,3 +221,17 @@ module.exports = neutrino => {
 Presets can also have their own custom data in addition to the Neutrino config. See your respective preset for details.
 Again, rather than reiterate the documentation for [webpack-chain](https://github.com/mozilla-neutrino/webpack-chain)
 here, please refer to its documentation for all ways you can modify a config instance to solve your use cases.
+
+### Conditional configuration
+
+Some plugins and rules are only available in certain environments. For example, the Web preset only exposes a minify
+plugin during production, leading to issues when trying to modify its settings, but blows up during development.
```
eliperelman: Switch "blows up" to "throws an exception". I typically avoid using colloquial prose in formal documentation.

eliperelman:
Note that this won't completely solve #180. We need to walk through the documentation of all middleware and in the list of plugins, etc., mark if they are in all environments or in certain environments.

helfi92:
@eliperelman makes sense. We only need to document the presets since they are the ones using the middlewares conditionally right? I made some [changes](https://github.com/mozilla-neutrino/neutrino-dev/pull/250/commits/6a589714dea2527f570cb9b4afc6ba2feedb39e2) to the documentation of the Web preset. I am planning to do the same thing for the rest of the presets. Let me know if I missed something or have a better idea.

eliperelman:
@helfi92 OOOOO I love the table idea! Yeah, go ahead and update the presets, I'll take care of the middleware.

helfi92:
Changed the Web and Node presets. I believe the rest don't add additional plugins.

eliperelman:
@helfi92 looks like Jest and Mocha document their changes to the stuff they use.

PR #253, opened 2017-06-14
Remove travis cache

eliperelman:


.travis.yml
```diff
@@ -3,37 +3,6 @@ node_js:
 - '6.10'
 - '7'
 - '8'
-cache:
-  yarn: true
```
edmorley: From experimentation the other day, I found the fastest runs where where the `node_modules` directories were not cached themselves, but the `~/.cache/yarn` directory (which is set via the `yarn: true` alias) still was.

As such, I'd leave caching enabled, but use this config (uses the shorthand for when there is only one entry):
```yaml
cache: yarn
```
(From https://docs.travis-ci.com/user/caching/#yarn-cache)

I believe this works the best since the cache keeps the files in archives rather than extracted, and is also reused multiple times over (compared to the individual `node_modules` directories where each is only used once per package, and is fully extracted) - so gives the best bang for buck per MB/file count.

edmorley:
Once this is merged it will be worth clearing the Travis caches (https://travis-ci.org/mozilla-neutrino/neutrino-dev/caches) to get rid of any used older versions of packages from the yarn cache.

eliperelman:
Pulling the trigger.

eliperelman:
Caches blown away.

edmorley:
I've retriggered the Travis run for the v6-dev branch commit where this merged, so the re-run uses the cache created by the landing. The run times are now <4 minutes rather than 5-6 minutes which is much nicer :-)

https://travis-ci.org/mozilla-neutrino/neutrino-dev/builds/242904097

The jobs should also be marginally faster when using the Travis Ubuntu trusty environment, since it does less busywork installing security updates (the v6-dev branch doesn't have that change from master).

eliperelman:
That's great! Thank you!

PR #256, opened 2017-06-14
Adding rudimentary tests to all packages

eliperelman:
Not extensive but enough I think for sanity checks. It *did* find a problem with the PWA middleware, so I guess that's something.


eliperelman:
Make that two problems: html-loader was missing neutrino as a peerDependency.

PR #259, opened 2017-06-15
Upgrading oao

eliperelman:


eliperelman:
Will need resolution on some issues from https://github.com/guigrpa/oao/issues/42 to continue on this.

PR #269, opened 2017-06-19
Reject test runners when non-zero exit code

eliperelman:


eliperelman:
@edmorley I really want to get to the place where we can have better integration-style tests. Right now all the tests for middleware and presets are sanity-based: as long as neutrino is able to require and `use` them, we let everything move forward.

It would be nice if preset tests could validate their output: either by running neutrino with the preset and making it doesnt bomb, or validating the actual output.

PR #270, opened 2017-06-20
Add Travis badge to README

edmorley:
The Travis badge is a handy way to jump to the Travis overview, when wanting to check settings/caches/see if master is green etc.

Also improves mark-up readability (has no effect on display layout).


eliperelman:
@edmorley can you also copy this to `docs/README.md`?

edmorley:
Ah yes - done :-)

PR #280, opened 2017-06-23
Ensure test cases are not run 2x in watch mode.

jaridmargolin:
Prior to this change, modifications of test files resulted in test cases being run 2x.

Reference: https://github.com/nikku/karma-browserify/issues/67#issuecomment-84448491

### DIFF
--- a/packages/neutrino-preset-karma/index.js
+++ b/packages/neutrino-preset-karma/index.js
@@ -24,7 +24,12 @@ module.exports = (neutrino, opts = {}) => {
       }
     },
     frameworks: ['mocha'],
-    files: [tests],
+    files: [{
+      pattern: tests,
+      watched: false,
+      included: true,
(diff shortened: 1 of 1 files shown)


packages/neutrino-preset-karma/index.js
```diff
@@ -23,7 +23,12 @@ module.exports = (neutrino, opts = {}) => {
       }
     },
     frameworks: ['mocha'],
-    files: [tests],
+    files: [{
+      pattern: tests,
+      watched: false,
```
eliperelman: Should `watched` be driven by the `args.watch`?

jaridmargolin: It should never be watched... The compiled test files are watched and that is all that is needed. No need to additionally watch the test src files.

PR #283, opened 2017-06-23
Update to utilize ChromeHeadless as browser in preset-karma.

jaridmargolin:
What a wonderful world we now live in.

Reference: https://developers.google.com/web/updates/2017/05/nic59

### DIFF
--- a/packages/neutrino-preset-karma/index.js
+++ b/packages/neutrino-preset-karma/index.js
@@ -17,10 +17,10 @@ module.exports = (neutrino, opts = {}) => {
     ],
     basePath: neutrino.options.root,
     logLevel: neutrino.options.debug ? LOG_DEBUG : LOG_INFO,
-    browsers: [process.env.CI ? 'ChromeCI' : 'Chrome'],
+    browsers: [process.env.CI ? 'ChromeCI' : 'ChromeHeadless'],
     customLaunchers: {
       ChromeCI: {
-        base: 'Chrome',
+        base: 'ChromeHeadless',
         flags: ['--no-sandbox']
       }
     },


eliperelman:
I think this would be a breaking change, so I've held off on this until we can move forward on v7.

barraponto:
I like this but I am not sure of adoption of latest Chrome. I expect everyone to be on the latest release, but then again there's always someone stuck with older versions. Angular CLI, for instance, is not using chrome headless (but they updated karma-chrome-launcher to make it possible, see https://github.com/angular/angular-cli/pull/6216)

edmorley:
I did some digging to see if we could work out the risk/timeframe for this.

The Chrome 59 blog post above links to another post with a few more details:
https://developers.google.com/web/updates/2017/04/headless-chrome

In there, it mentions headless isn't supported on Windows until Chrome 60 (ie later than other platforms) - though not sure if many people use Neutrino on Windows. However reading [chromium issue 686608](https://bugs.chromium.org/p/chromium/issues/detail?id=686608) it also sounds like there were a few other non-platform specific bugs that weren't fixed until Chrome 60 - so it sounds like that's the minimum version people should be targeting regardless.

The latest stable version of Chrome is now 61, with Chrome 60 having been released on the stable channel [on 2017-08-24](https://chromereleases.googleblog.com/2017/08/stable-channel-update-for-desktop_24.html). Surprisingly, the stock chromium-browser package for Ubuntu is already on Chrome 61 for even LTS as old as trusty ([see here](https://packages.ubuntu.com/trusty/chromium-browser)).

The last remaining pieces are presumably what versions of Chrome/Chromium TravisCI/CircleCI have pre-installed.

For TravisCI, they use the official google chrome binaries rather than chromium. These are installed on all images apart from "connie" (aka `lang: {bash,shell,sh,minimal}` and friends):
https://github.com/travis-ci/packer-templates/search?utf8=%E2%9C%93&q=chrome&type=

...so there may be a number of people relying on the pre-installed version rather than [manually specifying](https://docs.travis-ci.com/user/chrome) `chrome: stable` in their `.travis.yml`.

As for what that pre-installed version is, Travis's cookbook uses the APT repo stable channel rather than a pinned version, so it depends on when the `group: stable` images were last refreshed:
https://github.com/travis-ci/travis-cookbooks/blob/155f838f825eceb7f46b9e1731809948fbbdf917/cookbooks/travis_build_environment/recipes/google_chrome.rb
https://docs.travis-ci.com/user/build-environment-updates/2017-09-06/

To confirm, some test builds:
https://travis-ci.org/edmorley/travis-testing/builds/281684602

ie: For users on `dist: trusty` ([now the default](https://blog.travis-ci.com/2017-08-31-trusty-as-default-status)) they'll have 60.0.3112.113 installed. Though users who are still using the legacy `dist: precise` or have rolled back to an older trusty image using `group: deprecated-*` will not have a new enough version unless they are specifying `chrome: stable` in their `.travis.yml`.

As for CircleCI, I don't have much experience with it, but it appears their "2.0" platform images use the APT stable channel too, and were updated 14 hours ago - so should be fine:
https://hub.docker.com/r/circleci/node/
https://github.com/circleci/circleci-images/blob/2c10aedf786b24b9ada62bc8c4c05b77701fb9c5/shared/images/Dockerfile-browsers.template#L38-L43

They also have a legacy "1.0" platform, which is still running Chrome 54.0.2840.100, presuming their docs are up to date with what's actually installed in the Docker image (it's 8GB compressed so not inclined to try):
https://circleci.com/docs/1.0/build-image-trusty/#google-chrome
https://hub.docker.com/r/circleci/build-image/tags/

In conclusion: this could definitely be a breaking change for a very small number of people, so will need a major version bump - however unlike back in July when the previous discussion occurred, it's much more viable to go ahead and land this soon :-)

eliperelman:
Let's see if we can rebase this and target it to land when we are ready to ship v8.

eliperelman:
@jaridmargolin are you up for rebasing this?

jaridmargolin:
Sure thing.

eliperelman:
I merged this from the command line, not sure why GitHub didn't close this. Anyway, this is now landed!

PR #291, opened 2017-06-29
Add an .npmignore to all packages

### LINKED ISSUE #289: Packages on NPM include unnecessary files
For example the `tests/` directory and `yarn.lock`.

The latter is because the publish script uses `npm publish` rather than `yarn publish`, so doesn't know to exclude `yarn.lock`. However it's probably best to just whitelist the files we want explicitly using the `package.json` `files` key:
https://docs.npmjs.com/files/package.json#files

I'll open a PR.

edmorley:
Excluding the test directory and `yarn.lock` reduces the package size significantly (for example 85-90% reduction in both compressed and uncompressed size of neutrino-preset-web), plus reduces the noise when consumers need to grep their local `node_modules`.

An `.npmignore` has been used instead of the `files` directive in `package.json` since the latter can cause breakage that isn't shown in CI if the directory layout changes (which is likely given the Neutrino packages have a mixture of using the `src` directory and not, depending on how many files they include). 

`.npmignore` was used instead of `.yarnignore` since npm doesn't support the latter whereas yarn supports both.

Fixes #289.

### DIFF
--- a/packages/neutrino-middleware-banner/.npmignore
+++ b/packages/neutrino-middleware-banner/.npmignore
@@ -1,0 +1,2 @@
+/test/
+yarn.lock
--- a/packages/neutrino-middleware-chunk/.npmignore
+++ b/packages/neutrino-middleware-chunk/.npmignore
@@ -1,0 +1,2 @@
+/test/
+yarn.lock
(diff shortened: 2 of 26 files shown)


eliperelman:
Wow, that much of a reduction?! 😮

edmorley:
Yeah - yarn.lock is quite big in comparison to the actual middleware/presets!

Issue #287, opened 2017-06-26
Add `findRelatedTests` CLI option to jest preset


okonet:
It looks like Jest preset CLI options are missing `findRelatedTests` options that is useful with tools like [lint-staged](https://github.com/okonet/lint-staged)

https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/neutrino-preset-jest/src/index.js#L97

eliperelman:
@okonet Thanks for the issue! I couldn't find much in the way of documentation of this, but it appears to want string paths. How do you suppose this option should work? If specified, should we set it to `neutrino.options.entry`?

```diff
  const cliOptions = {
    config: configFile,
    coverage: args.coverage,
    watch: args.watch,
+   findRelatedTests: neutrino.options.source
  };
```

okonet:
It's suppose to get arguments from CLI. So in case with lint-staged, it will call `neutrino test --findRelatedTests ./path/to/file1.js ./path/to/file2.js ...`. Can this be accomplished?

okonet:
Here is the related documentation: http://facebook.github.io/jest/docs/cli.html#findrelatedtests-spaceseparatedlistofsourcefiles

eliperelman:
Yeah, that should be doable. I'll see if I can figure out a nice generic way to grab any extra CLI arguments from the Neutrino `args` and pass them on.

okonet:
Any updates on this one? Do you need any help?

okonet:
@eliperelman I think this issue can be expanded to any command. I tried to run `npm run lint -- --fix` but it won't get passed to `eslint` which is vey unfortunate IMO. Do you think we could expand the architecture to allow this?

okonet:
@eliperelman should I open a separate issue in relation to lint and other tasks? https://github.com/mozilla-neutrino/neutrino-dev/issues/287#issuecomment-328820679

eliperelman:
@okonet yes, go ahead! Right now flags aren't passed to eslint because we use ESLint CLIEngine, not the CLI itself.

eliperelman:
@okonet although not sure why `--fix` isn't working:

https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/neutrino-middleware-eslint/index.js#L79

PR #294, opened 2017-07-10
Allow arbitrary command-line args to be used in Jest

eliperelman:
Addresses #287.

@okonet I tried using this in a project with Jest, but I don't think it's working properly; it just seems to run all the tests. Do you think you would be able to tell if my changes are doing what they should, and potentially offer any advice here?

### DIFF
--- a/packages/neutrino-preset-jest/src/index.js
+++ b/packages/neutrino-preset-jest/src/index.js
@@ -94,8 +94,12 @@ module.exports = (neutrino, opts = {}) => {
     return new Promise((resolve, reject) => {
       const configFile = join(tmpdir(), 'config.json');
       const options = normalizeJestOptions(opts, neutrino, args);
-      const cliOptions = { config: configFile, coverage: args.coverage, watch: args.watch };
+      const cliOptions = merge(
+        omit(['inspect', 'quiet', '_', 'version', 'help', 'use', 'options', '$0', 'files'], args),
+        { config: configFile, coverage: args.coverage, watch: args.watch }
+      );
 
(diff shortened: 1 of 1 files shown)


packages/neutrino-preset-jest/src/index.js
```diff
@@ -94,8 +94,12 @@ module.exports = (neutrino, opts = {}) => {
     return new Promise((resolve, reject) => {
       const configFile = join(tmpdir(), 'config.json');
       const options = normalizeJestOptions(opts, neutrino, args);
-      const cliOptions = { config: configFile, coverage: args.coverage, watch: args.watch };
+      const cliOptions = merge(
+        omit(['inspect', 'quiet', '_', 'version', 'help', 'use', 'options', '$0', 'files'], args),
+        { config: configFile, coverage: args.coverage, watch: args.watch }
+      );
 
+      cliOptions.findRelatedFiles = cliOptions.findRelatedFiles.split(' ');
```
okonet: I'm not sure I understand this line. The command lint-staged is calling `jest --findRelatedTests file1.js file2.js file3.js`. Is this aligned with this `split` behaviour?

okonet: Oh I think I start getting it. Can you point me to the Node API of Jest?

okonet: So I think it should be something like `jest.runCLI({ findRelatedTests: true, '_': [args._]})`

eliperelman:
@mozilla-neutrino/core-contributors anyone have any cycles to try and continue on this implementation while I'm on PTO?

eliperelman:
Superseded by #313.

Issue #277, opened 2017-06-21
Live reload is triggered only every second time.


constgen:
When start project in a development mode with `neutrino start` not every file change triggers browser resources reload. CLI reports about changes and recompilation every time but reload happens only every second time. Expected that a browser will react to every file change and recompilation.
The issue can be reproduced on `neutrino-preset-web` and any on top of it.

constgen:
Didn't find a solution, but have some investigation results. I tried to localize the problem and seems that it is somewhere in the Neutrino core, not middlewares, not `webpack-chain`. I have created a [test suite](https://www.dropbox.com/s/0iqjcrrc8xeqiod/web-test-hot-reload.zip?dl=0).
There are two ways to start:
#### Using Neutrino
```
npm start
```
#### Using a regular Webpack config
```
npm run start-webpack
```
They both use the same configuration chain.

You can observe these results:
- starting from Webpack works as expected. Every change in `index.js` reloads the browser window. 
- starting from Neutrino doesn't reload at all. 

The next thing is the most interesting. Open `webpack-middleware.js` and switch from `only-dev-server` to `dev-server`. Now the config is like in `neutrino-middlware-dev-server`. The results are
- starting from Webpack throws an error. It fails even with simple configuration without `webpack-chain`. I could reproduce this error even in clean Webpack setup, so this is not our fault. 
- starting from Neutrino is successful and window is reloaded only every second time the file is changed. These is how all presets work now.

Now I am asking for support. We need somebody who is an expert in Webpack or Neutrino core.

constgen:
The issue is fixed. Need somebody to patch the version and publish a new release.

eliperelman:
Released in 6.1.5 by #304.

PR #304, opened 2017-07-20
Fix hot reload

constgen:
This fixes the issue when reloading is triggered not every time https://github.com/mozilla-neutrino/neutrino-dev/issues/277

### DIFF
--- a/packages/neutrino/src/webpack.js
+++ b/packages/neutrino/src/webpack.js
@@ -25,7 +25,7 @@ const compile = pipe(
 // devServer :: Object config -> Future Error Object
 const devServer = pipe(
   compiler,
-  Future.map(compiler => {
+  Future.map((compiler) => {
     const server = new DevServer(compiler, compiler.options.devServer);
     server.compiler = compiler;
     return server;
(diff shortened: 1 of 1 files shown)


packages/neutrino/src/webpack.js
```diff
@@ -25,9 +25,14 @@ const compile = pipe(
 // devServer :: Object config -> Future Error Object
 const devServer = pipe(
   compiler,
-  Future.map(compiler => assoc('compiler', compiler, new DevServer(compiler, compiler.options.devServer)))
+  Future.map((compiler) => {
```
eliperelman: This is essentially the same thing that was happening before, except that `assoc` was:

> Makes a shallow clone of an object, setting or overriding the specified property with the given value. Note that this copies and flattens prototype properties onto the new object as well. All non-primitive properties are copied by reference.

This was probably messing with the DevServer instance. Switch to an `Object.assign` instead:

```js
Future.map(compiler => Object.assign(new DevServer(compiler, compiler.options.devServer), { compiler }))
```

packages/neutrino/src/webpack.js
```diff
@@ -25,9 +25,14 @@ const compile = pipe(
...
+    const server = new DevServer(compiler, compiler.options.devServer);
+    server.compiler = compiler;
+    return server;
+  })
 );
 
+
```
eliperelman: Remove this extra line.

packages/neutrino/src/webpack.js
constgen: It was definitely the reason of the issue. Redesigned as requested.

PR #314, opened 2017-09-12
chore: Add lint-staged and husky

okonet:
Add lint-staged with an auto-fix configuration for all JS files.

See https://github.com/okonet/lint-staged

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
     "docs:deploy": "yarn docs:build && gh-pages --dist _book --remote upstream",
     "docs:serve": "gitbook serve",
     "release": ".scripts/publish",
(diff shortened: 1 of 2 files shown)


constgen:
Are you sure you want to automatically add all files to stage. Somebody may commit some trash and even will not notice this. Please disable this

okonet:
What do you mean? This will prevent committing trash. Please read my post.

okonet:
This won't add all files to stage. All commands run only on staged files so it only add same files back after fixing eslint rules that are auto-fixable I.e. prettier.

constgen:
I was wrong. Didn't realize how `lint-staged` works. `git add` confused me. I take my words back.

PR #317, opened 2017-09-15
Small API docs change to avoid confusion.

trusktr:
The API is first assigned to `neutrino`, but the rest of the doc uses `api`, which was seeming to come from nowhere.


eliperelman:
Deployed!

PR #325, opened 2017-09-21
Misc documentation fixes

edmorley:
See individual commits for more info.


edmorley:
I've updated this with some other changes.

PR #333, opened 2017-09-27
Make the lint command check .jsx files too

### LINKED ISSUE #332: Neutrino lint results differ from neutrino build
As part of upgrading to Neutrino v6, I was going through Treeherder's existing eslint `globals` configuration, to see if there were any entries that were no longer required. I found that if I commented out the `'React'` global, then whilst `neutrino lint` would report no errors, `neutrino build` returned:

```
× Building project failed
./ui/js/reactrevisions.jsx
Module build failed: Module failed because of a eslint error.

C:\Users\Ed\src\treeherder\ui\js\reactrevisions.jsx
  14:11  error  'React' is not defined  no-undef
  57:15  error  'React' is not defined  no-undef
  ...
```

At first I thought this might be a `NODE_ENV` difference, however it turns out to actually be due to the file being a `.jsx` file.

Looking at the source for [eslint-loader](https://github.com/MoOx/eslint-loader/blob/756330c19d2779c4102d40fb7bfbb1ad183b714b/index.js#L217) (which is what runs when using `neutrino build`) I see it uses eslint's `executeOnText()`, which just operates on the passed text regardless of file extension (since the webpack loader config already filtered by file extension).

However the `neutrino lint` command instead uses the eslint API via `executeOnFiles()` - and whilst that's likely the better choice for this use-case, means that if `include` is a directory (which it will be for Neutrino, since it will be set to `neutrino.options.source`), then eslint filters the files by the value for `extension`, which defaults to `['.js']`:
https://eslint.org/docs/developer-guide/nodejs-api#cliengine

As such, the lint command needs to override `extensions`, setting it to `['.js', '.jsx']`.

I'll open a PR.

edmorley:
Previously the `neutrino lint` command would only check `.js` files, which was inconsistent with the build/start commands, since they use `eslint-loader` and so instead use the `test` regex of `/\.(js|jsx)$/`.

This now means there is redundancy in the eslint middleware options, however there isn't a reliable way to turn the `test` regex into a flat array of file extensions. 

In a future major release, perhaps the eslint middleware could drop support for `options.test` and instead only accept the `options.extensions` array, which could then be mapped back to the `test` regex internally. However this would mean less flexible selection of files to be matched, plus mean that there would be inconsistent support for `options.test` between the various loader middlewares, so not sure whether it's the best way forwards anyway.

Fixes #332.

### DIFF
--- a/docs/middleware/neutrino-middleware-eslint/README.md
+++ b/docs/middleware/neutrino-middleware-eslint/README.md
@@ -46,6 +46,7 @@ neutrino.use(eslint, {
     cwd: neutrino.options.root,
     useEslintrc: false,
     root: true,
+    extensions: ['js', 'jsx'],
     plugins: ['babel'],
     baseConfig: {},
     envs: ['es6'],
--- a/packages/neutrino-middleware-eslint/README.md
+++ b/packages/neutrino-middleware-eslint/README.md
@@ -46,6 +46,7 @@ neutrino.use(eslint, {
     cwd: neutrino.options.root,
     useEslintrc: false,
     root: true,
+    extensions: ['js', 'jsx'],
     plugins: ['babel'],
     baseConfig: {},
     envs: ['es6'],
--- a/packages/neutrino-middleware-eslint/index.js
+++ b/packages/neutrino-middleware-eslint/index.js
@@ -56,6 +56,9 @@ module.exports = (neutrino, opts = {}) => {
             cwd: neutrino.options.root,
             useEslintrc: false,
             root: true,
+            // eslint-loader uses executeOnText(), which ignores the `extensions` setting.
+            // However it's still needed for the lint command, as it uses executeOnFiles().
+            extensions: ['js', 'jsx'],
             plugins: ['babel'],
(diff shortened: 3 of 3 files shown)


docs/middleware/neutrino-middleware-eslint/README.md
```diff
@@ -37,6 +37,7 @@ const eslint = require('neutrino-middleware-eslint');
 // Usage shows default values
 neutrino.use(eslint, {
   test: /\.(js|jsx)$/,
+  extensions: ['js', 'jsx'],
```
eliperelman: Should these be part of the eslint options?

edmorley: The option will be ignored when used by eslint-loader, but yeah it probably belongs in `eslint` anyway. I'll update the PR and add a comment.

edmorley:
The Travis failure appears unrelated.

eliperelman:
Thanks @edmorley!

eliperelman:
Released in v6.2.1.

Issue #272, opened 2017-06-21
`svg-url-loader` works incorrectly when load images not from CSS files.


constgen:
### The issue:
https://mega.nz/#!KZFlCD6B!ziIF6WMlEArcD8RPJChIsteJLIQfDgzKmlTxXyUlSxI
This is a simple project that demonstrates the problem with `svg-url-loader`.

It works as expected when SVG images are referenced from CSS code. But when load image with `import` and inject it in HTML with interpolation or by reference in JSX it works incorrectly. It is caused by wrapping quotes around resulting string `"..."`. Need to strip this quotes every time when you load SVG in JS
```
import logoSvg from './images/logo.svg'
const imageSvg = JSON.parse(logoSvg)
```

It is expected that `svg-url-loader` will work the same way as `image-loader`

### The solution:
Webpack loader config has an `"issuer"` property. There is a good example of usage in the Less loader https://webpack.js.org/loaders/less-loader/#non-less-imports . We can pass different config to `svg-url-loader` dependeing on a dependency parent. `svg-url-loader` supports `"noquotes"` option that disables that quotes.

But `webpack-chain` doesn't support `"issuer"` property, that's why I propose to add it. @eliperelman 

Also there is a chance that we will have to contribute to `svg-url-loader` and add an additional option.

eliperelman:
Note: webpack-chain supports issuer, it just doesn't have a shorthand method for it:

```js
neutrino.config.rule(name).set('issuer', ...)
```

Also, feel free to PR webpack-chain and add issuer to Rule extensions.

constgen:
Ok, I will investigate it.

eirikurn:
This is affecting us badly. @constgen did you get this to work with `issuer`?

Does it make sense to revert to `file-loader`, which "just works", until this is solved?

eliperelman:
@timkelty you did the original PR for switching to svg-url-loader. What is your opinion here?

timkelty:
Hmm - yeah I was under the impression `svg-url-loader` was a drop-in replacement, but apparently it isn't.

Sounds like maybe it is a good idea to revert back to `file-loader` until we figure out the proper fix.

constgen:
Started working on it. Currently have an issue with separating `{ noquotes: true }` and `{ noquotes: false}` configs. Seams like a bug in Webpack but I am not sure. This is what I tried:
```js
  config.module
    .rule('svg')
[...]

PR #334, opened 2017-09-28
Bugfix/svg loader

constgen:
Fixes https://github.com/mozilla-neutrino/neutrino-dev/issues/272
Uses old fashion "oneOf" declaration because older version of `webpack-chain` is used. Now settings for SVG Loader are split between styles and all other modules.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ install:
 - yarn install --frozen-lockfile
 before_install:
 # Required due to: https://github.com/travis-ci/travis-ci/issues/7951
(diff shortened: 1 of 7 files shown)


eliperelman:
I'm going to close this as it seems like we want to use the url-loader based on analysis by @edmorley in https://github.com/mozilla-neutrino/neutrino-dev/issues/272#issuecomment-332885244. Right now v7 uses url-loader, so we can revisit when packages get updated.

Thanks anyway for this!

PR #315, opened 2017-09-12
Upgrading Webpack to v3, merge config as middleware

eliperelman:

PR #339, opened 2017-10-02
Remove unused dependencies

edmorley:
**1) Moves preset-airbnb-base dependencies out of middleware-eslint**

The `eslint-config-airbnb-base` and `eslint-plugin-import` packages are not used by `neutrino-middleware-eslint` itself, but instead by `neutrino-preset-airbnb-base`.

**2) Removes unused dependencies**

* `webpack-dev-middleware` can be removed from the karma preset, since it's only used by `karma-webpack` which already has it as a dependency.
* `webpack-dev-server` can be removed from the web preset since it's a leftover from when `neutrino-middleware-dev-server` was created.
* the web preset's `core-js` and `babel-polyfill` dependencies are leftover from #315.
* everything else was manually determined to be unused via correlating the package list against greps of the source, plus checking that the package wasn't a peer dependency of another listed package.

The yarn.lock comment changes are because I'm using an up to date version of yarn that no longer adds the version numbers to the lockfile (which aims to reduce lock file churn due to version differences). Hopefully shortly we'll be on a newer yarn everywhere, thereby avoiding this.


packages/neutrino-middleware-eslint/package.json
```diff
@@ -20,14 +20,10 @@
     "eslint": "^4.7.2",
     "eslint-loader": "^1.9.0",
     "eslint-plugin-babel": "^4.1.2",
-    "eslint-plugin-import": "^2.7.0",
     "fluture": "^7.1.3",
     "lodash.clonedeep": "^4.5.0",
     "ramda": "^0.24.1"
   },
-  "devDependencies": {
-    "eslint-config-airbnb-base": "^12.0.1"
```
eliperelman: In the past the reason this devDependency existed had to do with local development and linking. ESLint didn't like to resolve where this package existed for the Airbnb preset when you linked it, probably due to the nested node_modules structure. My guess is that problem still exists, but can probably be fixed if we move to yarn workspaces and hoisted deps.

edmorley: I agree there is a bug in yarn where it's over-eager to complain about "missing" peer deps (I'm going to file that separately). However in this case, the `devDependency` was actually in the wrong package - it's airbnb base that's using it :-)

eliperelman:
Released in v7.0.2.

packages/neutrino-middleware-eslint/package.json
eliperelman: Yeah, that's what I mean. When you link `neutrino-preset-airbnb-base`, ESLint couldn't resolve `eslint-config-airbnb-base` because the eslint binary lives in `neutrino-middleware-eslint` but it was trying to use a working directory relative to `neutrino-preset-airbnb-base`.

edmorley: Ah yeah that makes sense. I thought you meant a related bug that I ran into, where yarn outputs "missing peer" for a case that npm doesn't.

I'm looking forward to yarn workspaces working well and all of this going away! :-)

eliperelman: Me too Ed, me too.

edmorley:
> Released in v7.0.2.

There are a few 7.0.1 version references that escaped being updated to 7.0.2:
https://github.com/mozilla-neutrino/neutrino-dev/blob/7e7ecfc9147df945396ac472b98890e68fce9534/packages/neutrino-preset-airbnb-base/package.json#L21
https://github.com/mozilla-neutrino/neutrino-dev/blob/7e7ecfc9147df945396ac472b98890e68fce9534/packages/neutrino-preset-mocha/package.json#L20
https://github.com/mozilla-neutrino/neutrino-dev/blob/7e7ecfc9147df945396ac472b98890e68fce9534/packages/neutrino-preset-react/package.json#L27

I don't know if this matters at all?

eliperelman:
For major versions it would definitely matter, for patches and minors probably not, but this is definitely wrong. Maybe it's a race condition between the publish of dependencies and the updating of them in package.json.

edmorley:
> I thought you meant a related bug that I ran into, where yarn outputs "missing peer" for a case that npm doesn't.

I've filed yarnpkg/yarn#4675 for this other case.

PR #344, opened 2017-10-05
Preserve ESLint default include when just exclude set

edmorley:
Previously if `exclude` was set manually, it would prevent the default value for `include` being used. Now consumers of the middleware can pass just `exclude` and not have to pass both. See:
https://github.com/mozilla-neutrino/neutrino-dev/pull/315#discussion_r141779549

The default include in neutrino-preset-airbnb-base has also been removed, since it just duplicated that in neutrino-middleware-eslint.


eliperelman:
Released in v7.1.0.

PR #346, opened 2017-10-06
Feature: upgrade webpack-chain to use .before() and .after()

eliperelman:


docs/webpack-chain.md
```diff
@@ -523,37 +528,37 @@ _NOTE: Do not use `new` to create the plugin, as this will be done for you._
 ```js
 neutrino.config
   .plugin(name)
-  .use(WebpackPlugin, args)
+    .use(WebpackPlugin, args)
```
edmorley: Some of these indentation fixes are also applicable to the upstream webpack-chain docs and could be upstreamed at some point :-) (I'm using DiffMerge to do partial line comparisons between the two)

eliperelman: Grr, I thought I caught all those there. 😭

eliperelman:
Released in v7.1.0.

PR #347, opened 2017-10-07
Update neutrino-middleware-hot to use named plugin

leesiongchan:
This is to fix the warning of
```
[HMR] Consider using the NamedModulesPlugin for module names.
```


packages/neutrino-middleware-hot/index.js
```diff
@@ -1,3 +1,5 @@
-const { HotModuleReplacementPlugin } = require('webpack');
+const { HotModuleReplacementPlugin, NamedModulesPlugin } = require('webpack');
 
-module.exports = ({ config }) => config.plugin('hot').use(HotModuleReplacementPlugin);
+module.exports = ({ config }) => config.plugin('hot')
+  .use(HotModuleReplacementPlugin)
+  .use(NamedModulesPlugin);
```
eliperelman: Using `use` twice on the same plugin will replace the first with the second. You want to create another plugin instance and use that:

```js
.plugin('hot')
  .use(HotModuleReplacementPlugin)
  .end()
.plugin('named-modules')
  .use(NamedModulesPlugin);

eliperelman:
Could you also update the READMEs showing the plugins that are loaded in each environment with the new plugin? Let me know if you need links.

PR #349, opened 2017-10-08
Update neutrino-middleware-hot to use named plugin

leesiongchan:
Sorry, accidentally made a silly mistake, (https://github.com/mozilla-neutrino/neutrino-dev/pull/347)

### DIFF
--- a/packages/neutrino-middleware-hot/README.md
+++ b/packages/neutrino-middleware-hot/README.md
@@ -57,6 +57,7 @@ The following is a list of plugins and their identifiers which can be overridden
 | Name | Description | Environments and Commands |
 | --- | --- | --- |
 | `hot` | Enables Hot Module Replacement. | all |
+| `named-modules` | Show the relative path of the module to be displayed when HMR is enabled. | all |
 
 ## Contributing
 
--- a/packages/neutrino-middleware-hot/index.js
+++ b/packages/neutrino-middleware-hot/index.js
@@ -1,3 +1,10 @@
-const { HotModuleReplacementPlugin } = require('webpack');
+const { HotModuleReplacementPlugin, NamedModulesPlugin } = require('webpack');
 
-module.exports = ({ config }) => config.plugin('hot').use(HotModuleReplacementPlugin);
+module.exports = ({ config }) => {
+  config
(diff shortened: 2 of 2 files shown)


packages/neutrino-middleware-hot/index.js
```diff
@@ -1,3 +1,10 @@
-const { HotModuleReplacementPlugin } = require('webpack');
+const { HotModuleReplacementPlugin, NamedModulesPlugin } = require('webpack');
 
-module.exports = ({ config }) => config.plugin('hot').use(HotModuleReplacementPlugin);
+module.exports = ({ config }) => {
+  config
+    .plugin('hot')
+      .use(HotModuleReplacementPlugin)
+      .end()
+    .plugin('named-modules')
```
edmorley: It's not clear whether the order of the plugins matters here - for that and the fact that the suggestion of using these plugins together isn't actually documented anywhere (other then the console output), I've filed webpack/webpack.js.org/issues/1623.

packages/neutrino-middleware-hot/README.md
```diff
@@ -57,6 +57,7 @@ The following is a list of plugins and their identifiers which can be overridden
 | Name | Description | Environments and Commands |
 | --- | --- | --- |
 | `hot` | Enables Hot Module Replacement. | all |
+| `named-modules` | Show the relative path of the module to be displayed when HMR is enabled. | all |
```
edmorley: The description for the other usages of `NamedModulesPlugin` (from neutrino-middleware-chunk) use this string:

> Enables named modules for improved debugging and console output.

...I think it might be good to make them consistent?

Also, sadly when updating the package docs, we have to make identical edits to both the copy under `packages/<PACKAGE>/README.md` and `docs/{middleware,presets}/<PACKAGE>/README.md`. 

Finally in this case, because `neutrino-middleware-hot` is used by both `neutrino-preset-web` and `neutrino-preset-node`, those package docs will also need updating (with the slight catch that for NODE_ENV=production, the web preset already used `named-modules` via `neutrino-middleware-chunk`).

eliperelman:
This was resolved in #365.

PR #350, opened 2017-10-08
Fix neutrino-middleware-loader-merge when using with babel

### LINKED ISSUE #348: compile-loader is unable to extend once its created
```
neutrino.use(compileLoader, {
  babel: compileLoader.merge(babelOpts, {
    plugins: [
      'AAA',
    ],
  }),
});

neutrino.use(loaderMerge('compile', 'babel'), {
  plugins: [
    'BBB',
  ],
});
```

This cannot work as the plugin `BBB` will simply be ignored.

leesiongchan:
This is to fix #348 

But yet I know it isn't safe to hardcode to check for `babel` only, but how can we know when to use `babelMerge`?

### DIFF
--- a/packages/neutrino-middleware-loader-merge/index.js
+++ b/packages/neutrino-middleware-loader-merge/index.js
@@ -1,6 +1,7 @@
+const babelMerge = require('babel-merge');
 const merge = require('deepmerge');
 
 module.exports = (ruleId, loaderId) => ({ config }, options) => config.module
   .rule(ruleId)
   .use(loaderId)
-  .tap(opts => merge(opts, options));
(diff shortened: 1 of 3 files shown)


eliperelman:
I don't think we can do this since, as you said, this requires hardcoding loader IDs, which isn't very reusable if you want to merge babel configuration but your rules are named differently.

I like the idea in concept, I'm just not sure if there is much to be gained over this instead of doing:

```js
neutrino.module
  .rule('compile')
    .use('babel')
      .tap(options => babelMerge(options, newOptions));
```

leesiongchan:
Ok, understandable, then I close this PR for now.

PR #355, opened 2017-10-10
Use latest version of yarn on Travis

edmorley:
The only remaining error with oao + recent yarn turned out to be due to OOM (https://github.com/guigrpa/oao/issues/51), which can be solved by switching to Travis' fully virtualised infra, which [has more RAM](https://docs.travis-ci.com/user/reference/overview/#Virtualization-environments).

The `--frozen-lockfile` parameter has also been moved to `.travis.yml` so that the lockfiles are still updated when working making changes locally.

Being on latest yarn makes it a lot easier to experiment with switching to yarn workspaces and/or lerna.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,8 @@
+# Use the fully virtualised infra not for sudo, but since their container
+# infra's 4GB of RAM is not sufficient to prevent yarn bootstrap OOMs:
+# https://github.com/guigrpa/oao/issues/51
+# https://docs.travis-ci.com/user/reference/overview/#Virtualization-environments
+sudo: required
 dist: trusty
 language: node_js
 cache: yarn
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,3 +1,8 @@
+# Use the fully virtualised infra not for sudo, but since their container
```
eliperelman: s/, but//

.travis.yml
```diff
@@ -1,3 +1,8 @@
...
+# infra's 4GB of RAM is not sufficient to prevent yarn bootstrap OOMs:
+# https://github.com/guigrpa/oao/issues/51
+# https://docs.travis-ci.com/user/reference/overview/#Virtualization-environments
+sudo: required
```
eliperelman: Do we need sudo?

.travis.yml
edmorley: That changes the meaning to the opposite of the what it's intending to say.

To explain in another way - Travis has two trusty based stacks:
* container (`sudo: false`): no-sudo, less RAM
* fully virtualised (`sudo: required`): has sudo, more RAM, few other differences

Much of the time people use the latter because of the need to use `sudo` commands rather than because of the difference in machine specifications, which is why Travis use the `sudo: {false,required}` setting to control which environment is used. 

However in this case, the need for sudo commands is not why we're setting `sudo: required`, it's because we want the higher spec fully virtualised environment.

I'm open to suggestions to another way of working that makes that clearer?

eliperelman: No no, I was just confused as to why travis configures it this way. Proceed. 😄

PR #361, opened 2017-10-17
Update to the newest version of html-webpack-template

edmorley:
Counter-intuitively the most recent version of the package is actually 5.6.0, rather than 6.0.0 or 6.0.1 due to a publishing error. See:
https://github.com/jaketrent/html-webpack-template/issues/59

Version 5.6.0 is also the version one gets when running `yarn add html-webpack-template`.


edmorley:
Btw I have a few more small papercuts type PRs to open - I think it may be worth waiting a bit before publishing new releases to batch them up a bit :-)

PR #365, opened 2017-10-20
feat(hot): Include NamedModulesPlugin for better developer experience

okonet:
https://webpack.js.org/plugins/named-modules-plugin/ is a recommended plugin to use with HMR and it enables better debug information in the console.


eliperelman:
So in #349 there is a similar pull request, but it has gone inactive. Would you be willing to make the changes that we request from there and incorporate them here?

eliperelman:
I'll cherry-pick the docs and get this out. Thank you so much!

PR #366, opened 2017-10-20
fix: Use cheap-module-eval-source-map for faster recompilation

okonet:
https://webpack.js.org/guides/build-performance/ suggests to use `cheap-module-eval-source-map` to improve incremental build speed. This reduces recompilation times by a magnitude on bigger projects.


eliperelman:
I'll wait to release this one to go along with the HMR changes.

PR #368, opened 2017-10-21
Allow cleaning directories outside of project root

### LINKED ISSUE #362: Setting options.output to relative directory above project directory causes clean middleware to not work
If you override the `output` directory like so:

```js
module.exports = {
  options: {
    output: '../build'
  },
};
```

The `clean` middleware/plugin no longer works.

eliperelman:
Fixes #362.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "docs:build": "gitbook build && cp CNAME _book",
     "docs:deploy": "yarn docs:build && gh-pages --dist _book --remote upstream",
     "docs:serve": "gitbook serve",
-    "precommt": "lint-staged",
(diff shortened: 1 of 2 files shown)


edmorley (commented):
Is this a bit risky? Eg if someone accidentally uses an output directory of '/' rather than './'  ?

Perhaps the neutrino API could block some clearly typoed cases?

eliperelman:
Maybe instead of allowing this, we should just document that paths outside of the project root need an override? I don't want to get into blocking paths.

eliperelman:
In order to avoid risk, I'm going to close this in favor of just documenting the option in the clean middleware.

PR #372, opened 2017-10-24
Follow-ups to the switch to Lerna

edmorley:
* Removes the now unused publish script and its dependencies.
* Removes the `lerna bootstrap` from the Travis run, since it's equivalent to the `yarn install` already run prior.
* Removes `yarn.lock` from the package's `.npmignore`, since there is now only the repo root instance of the file.
* Updates the development docs to reflect the new workflow.
* Sets node and yarn version ranges in `package.json` to help avoid issues caused by using a version not recommended by the docs.
* Adds `yarn release` and `release:preview` aliases using lerna, for easier publishing.
* Removes version and unnecessary `devDependencies` from root `package.json`.


edmorley:
I've also cleared the Travis caches for this repo, to reduce the bloat from old package versions, so once re-populated after the next master push, between that and the switch to lerna, the travis run setup phase should be pretty fast :-)

package.json
```diff
@@ -7,6 +7,10 @@
   "homepage": "https://neutrino.js.org",
   "bugs": "https://github.com/mozilla-neutrino/neutrino-dev/issues",
   "repository": "mozilla-neutrino/neutrino-dev",
+  "engines": {
+    "node": ">=6.1.0",
```
eliperelman: Should this be `>=6.10.0`?

edmorley: Ah yes it should, good spot :-)

edmorley:
I've updated the PR to address that review comment, but have also added a few other things (see the OP for what's new), which made it a bit longer, so I've now split up into separate commits for easer review.

package.json
```diff
@@ -21,18 +24,14 @@
...
     "docs:serve": "gitbook serve",
     "precommit": "lint-staged",
-    "release": "node .scripts/publish",
+    "release": "lerna publish --force-publish=*",
+    "release:preview": "lerna publish --force-publish=* --skip-git --skip-npm",
     "test": "nyc --reporter lcov ava packages/*/test"
   },
   "devDependencies": {
     "ava": "^0.22.0",
-    "babel-eslint": "^8.0.1",
```
eliperelman: Removing all these deps is nicer now that hoisting is in place.

PR #373, opened 2017-10-24
Automatically generate preset and middleware docs from package READMEs

### LINKED ISSUE #271: Sync docs with packages READMEs
Right now it's a mess of copy-pasting between docs/{presets,middleware}/README.md and packages/*/README.md. We should sync the items in packages to docs when running yarn docs scripts.

eliperelman:
Fixes #271.

Now we don't need to make changes in 2 places for presets and middleware docs. We only need to change the readmes in the package directory, and the script will copy the file over to `docs/` with the URL resolved for gitbook and for github traversal.


edmorley:
Before I review this, I wonder whether it's worth seeing if we can move away from the duplicated docs entirely? It looks like perhaps we could have multiple versions using:
https://plugins.gitbook.com/plugin/versions

With that, presumably we could make the `packages/*/README.md` file the canonical file (since it needs to exist for the package published to NPM), and just include it from the `docs/*/` copy (since we'd no longer need to keep the relative URLs working when browsing GitHub), using:
https://toolchain.gitbook.com/templating/conrefs.html

eliperelman:
Going to close this as it's quite bitrotted. Gitbook says they will release their beta at the end of the month, so *reaaaaaaaaaly* hoping we can get into that soon.

https://github.com/GitbookIO/gitbook/issues/1961#issuecomment-356860092

PR #376, opened 2017-10-25
Scaffold new projects with a CLI helper

helfi92:


helfi92:
This pull-request allows the user to scaffold a project application, library, or components.

#### Applications
* React (one with react-router and one without)
* Web
* Node

#### Libraries
* Web Library (uses taskcluster-web-library. can we rename it to neutrino-preset-web-library)

#### Components
* React Components

#### Linters
* Jest
* Karma
* Mocha

packages/create-neutrino-project/LICENSE
```diff
@@ -0,0 +1,373 @@
+Mozilla Public License Version 2.0
```
eliperelman: We can delete this file now that this is part of the monorepo.

packages/create-neutrino-project/bin/create-neutrino-project.js
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env node
+
+'use strict';
```
eliperelman: Remove this `use strict`.

packages/create-neutrino-project/bin/create-neutrino-project.js
```diff
@@ -0,0 +1,34 @@
...
+
+const yeoman = require('yeoman-environment');
+const Vorpal = require('vorpal');
```
eliperelman: Since we did away with vorpal in favor of yargs in the Neutrino CLI, let's use that here as well.

packages/create-neutrino-project/commands/init/generator/install.js
```diff
@@ -0,0 +1,11 @@
+const commandExists = require('command-exists');
+
+module.exports = function() {
+  const isYarnAvailable = commandExists.sync('yarnpkg');
```
eliperelman: Even though the user may have `yarn` installed, that isn't enough to know whether they want it used on this particular project. Instead, you can find out through the environment variables.

```js
const isYarn = process.env.npm_config_user_agent.includes('yarn');
```

packages/create-neutrino-project/commands/init/generator/questions.js
```diff
@@ -0,0 +1,140 @@
...
+  {
+    name: 'projectType',
+    type: 'list',
+    message: 'What kind of components?',
+    when: (data) => data.project === 'components',
+    choices: [
+      {
+        name: 'React Component',
+        value: 'react-components',
+        checked: false
```
eliperelman: We can probably default this to being checked since it's the only option.

packages/create-neutrino-project/commands/init/generator/questions.js
```diff
@@ -0,0 +1,140 @@
...
+      },
+    ]
+  },
+  {
+    name: 'router',
+    type: 'confirm',
+    message: 'Do you want to build a single-page application? I can set up code-splitting for you.',
```
eliperelman: Let's rephrase this to be route-based:

"Would you like to use React Router? If so, I will automatically code-split your routes for you."

packages/create-neutrino-project/commands/init/generator/questions.js
```diff
@@ -0,0 +1,140 @@
...
+    when: (data) => data.projectType === 'react'
+  },
+  {
+    name: 'testRunner',
+    type: 'list',
+    message: 'What kind of test runner do you want?',
```
eliperelman: "Would you like to add a test runner to your project?"

packages/create-neutrino-project/commands/init/generator/questions.js
```diff
@@ -0,0 +1,140 @@
...
+    choices: [{
+      name: 'Web',
+      value: 'taskcluster-web-library',
+      checked: true
+    }]
+  },
+  {
+    name: 'projectType',
+    type: 'list',
+    message: 'What kind of components?',
```
eliperelman: "What kind of components would you like to create?"

packages/create-neutrino-project/commands/init/generator/questions.js
```diff
@@ -0,0 +1,140 @@
...
+      { name: 'React',
+        value: 'react',
+        checked: true
+      },
+      {
+        name: 'Web',
+        value: 'web',
+        checked: false
+      },
+      { name: 'Node',
```
eliperelman: Change "Node" to "Node.js".

packages/create-neutrino-project/commands/init/generator/questions.js
```diff
@@ -0,0 +1,140 @@
+const path = require('path');
+
+module.exports = (defaults) => [
+  {
+    name: 'project',
+    type: 'list',
+    message: 'Do you want to build an application, a library, or components?',
+    choices: [
+      { name: 'Application',
```
eliperelman: Some of these properties are on the same line as the brace; move them to their own line.

packages/create-neutrino-project/commands/init/generator/questions.js
```diff
@@ -0,0 +1,140 @@
...
+        value: false,
+        checked: false
+      },
+    ],
+    store: true
+  },
+  {
+    name: 'linter',
+    type: 'list',
+    message: 'What kind of linter do you want?',
```
eliperelman: "Would you like to add linting to your project?"

packages/create-neutrino-project/bin/create-neutrino-project.js
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env node
```
eliperelman: I'm thinking the syntax of this command should be:

```js
create-neutrino-project <directory>
```

So that we know the directory name to create, as well as a project name to infer from this directory. We should throw when this option is not preset. That will also save us from asking the project name question.

packages/create-neutrino-project/commands/init/generator/prompting.js
```diff
@@ -0,0 +1,23 @@
+const questions = require('./questions');
+
+module.exports = function () {
+  const defaults = {
+    author: '',
+    description: 'Get Started with Neutrino!'
+  };
+  const done = this.async();
+  const command = this.spawnCommand('bash', [
```
eliperelman: We shouldn't need this author command if we are going to spawn a command for `yarn init` or `npm init`.

packages/create-neutrino-project/commands/init/generator/questions.js
```diff
@@ -0,0 +1,140 @@
...
+      {
+        name: 'none',
+        value: false,
+        checked: false
+      }
+    ],
+    store: true
+  },
+  {
+    name: 'name',
```
eliperelman: As noted in the bin, we can skip this question if we require a directory name to be generated up front.

packages/create-neutrino-project/commands/init/generator/questions.js
```diff
@@ -0,0 +1,140 @@
...
+    message: 'What is the name of your package (i.e. npm package name)?',
+    default: path.basename(process.cwd()).replace(/\s/g, '-'),
+    store: false
+  },
+  {
+    name: 'author',
```
eliperelman: Let's skip defining the author and description as well. The user can fill this in themselves when they are ready, or it will be automatically populated by `npm init` or `yarn init`.

packages/create-neutrino-project/commands/init/generator/utils.js
```diff
@@ -0,0 +1,28 @@
+module.exports.VERSIONS = {
+  NEUTRINO: '^7.2.1',
```
eliperelman: So we definitely don't want to do this, which is what led us away from continuing with Neo. By hardcoding these versions here, we will be forced to release a new version of the project creator every time one of these needs a version bump. Instead, we can invoke the installation commands directly, and let the latest always be installed.

packages/create-neutrino-project/commands/init/generator/writing.js
```diff
@@ -0,0 +1,158 @@
...
+      config.use.push(`neutrino-preset-${data.linter}${data.projectType === PROJECTS.WEB ? '-base' : ''}`);
+      break;
+    case LINTERS.STANDARDJS:
+      config.use.push(`neutrino-middleware-${LINTERS.STANDARDJS}`);
+      break;
+    default:
+      config.use.push(`neutrino-preset-${data.linter}`);
+  }
+
+  return `module.exports = ${JSON.stringify(config, null, 2)};\n`;
```
eliperelman: Instead of `JSON.stringify`, let's try the `javascript-stringify` package that Neutrino also uses.

packages/create-neutrino-project/commands/init/generator/writing.js
```diff
@@ -0,0 +1,158 @@
...
+    if (data.testRunner === 'karma' || data.testRunner === 'mocha') {
+      devDependencies[`assert`] = VERSIONS.ASSERT;
+    }
+
+    scripts.test = 'neutrino test';
+  }
+
+  if (data.projectType === PROJECTS.REACT && data.router) {
+    dependencies['react-router-dom'] = VERSIONS.REACT_ROUTER_DOM;
+    dependencies['react-async-component'] = VERSIONS.REACT_ASYNC;
```
eliperelman: I like react-loadable, and it works super-well with Webpack. Can we use that instead?

packages/create-neutrino-project/commands/init/generator/writing.js
```diff
@@ -0,0 +1,158 @@
...
+
+module.exports = function () {
+  let templateDir = this.data.projectType;
+
+  if (this.data.router === 'react-router') {
+    templateDir = 'react-router';
+  }
+
+  this.fs.copy(this.templatePath(`${templateDir}/src/**/*`), this.destinationPath('src'));
+  this.fs.write(this.destinationPath('package.json'), makePackageJson(this.data));
```
eliperelman: We shouldn't generate this file. Instead, lets execute a child process with either `npm init --yes` or `yarn init --yes`. After that, we can execute `yarn add --dev <DEV_DEPS> && yarn add <DEPS>`. Then we can overwrite the package.json by augmenting our scripts.

packages/create-neutrino-project/commands/init/templates/react-components/src/stories.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { render } from 'react-dom';
+import { Stories, Story, Props } from 'neutrino-preset-react-components/lib';
+import HelloWorld from './components/HelloWorld';
+
+const root = document.getElementById('root');
+
```
eliperelman: This file should probably use HMR if possible.

packages/create-neutrino-project/commands/init/templates/react-router/src/App.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { BrowserRouter as Router, Route, Link } from 'react-router-dom';
+import './App.css';
+
+import AsyncFoo from './components/Foo';
```
eliperelman: I want to avoid a lot of boilerplate that is going to end up being deleted. Let's give them a single home route that is dynamically imported.

packages/create-neutrino-project/commands/init/templates/react-router/src/App.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { BrowserRouter as Router, Route, Link } from 'react-router-dom';
```
eliperelman: I'm OK with leaving this import named as `BrowserRouter`.

packages/create-neutrino-project/commands/init/templates/react-router/src/index.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { render } from 'react-dom';
+import { AppContainer } from 'react-hot-loader';
+
```
eliperelman: Remove extra blank line.

packages/create-neutrino-project/commands/init/templates/react-router/src/index.js
```diff
@@ -0,0 +1,21 @@
...
+    <AppContainer>
+      <App />
+    </AppContainer>,
+    document.getElementById('root'),
+  );
+};
+
+// This is needed for Hot Module Replacement
+if (module.hot) {
+  module.hot.accept('./App', () => renderApp());
```
eliperelman: The `renderApp` function can be passed by reference:

```js
module.hot.accept('./App', renderApp);
```

packages/create-neutrino-project/commands/init/templates/react/src/App.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import './App.css';
+
+const App = () => (
```
eliperelman: This may be controversial, but let's use a component here.

```js
export default class App extends Component {
  render() {
    return (
      <div className="App">
        <h1>{/* ••••••• insert the directory name here we got from the command line •••••••••/*}</h1>
      </div>
    );
  }
}
```

packages/create-neutrino-project/commands/init/templates/taskcluster-web-library/src/index.js
```diff
@@ -0,0 +1 @@
+export { sum, product } from './my-library';
```
eliperelman: We don't need the extra import and export dance here. Let's export directly from `src/index.js` our functionality:

```js
export default () => console.log(/* Maybe a "Welcome to <directory name>" here.*/);
```

packages/create-neutrino-project/commands/init/templates/taskcluster-web-library/src/my-library.js
```diff
@@ -0,0 +1,4 @@
+const sum = (a, b) => a + b;
```
eliperelman: Delete this file.

packages/create-neutrino-project/commands/init/templates/test/eslintrc.js
```diff
@@ -0,0 +1,5 @@
+const { Neutrino } = require('neutrino');
```
eliperelman: We can probably move this file out of the `test` directory and into a more appropriate one.

packages/create-neutrino-project/commands/init/templates/test/eslintrc.js
```diff
@@ -0,0 +1,5 @@
...
+
+const api = Neutrino();
```
eliperelman: To save a declaration, let's export directly:

```js
module.exports = Neutrino().call('eslintrc');
```

edmorley: Let's also update the main Neutrino docs to suggest doing this too :-)

helfi92:
Implemented the requested changes. A couple of points I wanted to bring up:

* We used to use have react-loadable for the react-router-starter kit but I remember reverting back because it was causing issues with HMR. I can try again and see if the issue has been fixed.

* After performing a (yarn|npm) init for the user, an attempt to extend the package json file with the `scripts` property notifies the user about a conflict with `package.json`. I'm not sure how to go around it. Yeoman seems to always ask the user for confirmation before overwriting a file:

```
conflict project/package.json
? Overwrite project/package.json? (ynaxdH)
```

* An attempt to add HMR to react-components yields an error on the client side: `module.hot is not a function`. I'm not sure why this is happening. I had `react-hot-loader` imported as well as the Stories component wrapped in AppContainer. Any idea? I'll spend more time investigating.

That's it for now, looking forward to having this reviewed :) Thanks!

barraponto:
First, we enforce neutrinorc.js, then we end up making a command line tool to scaffold and soon we end up setting configurations that can't be upgraded (just like webpack and the framework cli tools)...
  
i told you so

eliperelman:
@barraponto I am pretty much against any kind of scaffolding where we start to dictate application specifics or increasingly complex configuration. Basically, what I want to achieve here is to reduce the amount of work it takes to start a new Neutrino project. For example, here are the steps I take to create a new React app:

---

```sh
mkdir example-project && cd example-project
yarn init --yes
yarn add --dev neutrino neutrino-preset-react neutrino-preset-airbnb
yarn add react react-dom
touch .neutrinorc.js
mkdir src
touch src/index.js
```

```js
// create .neutrinorc.js with following content
module.exports = {
  use: [
    'neutrino-preset-airbnb',
    'neutrino-preset-react'
  ]
};
```

```js
// edit package.json with following changes
{
  "scripts": {
      "start": "neutrino start",
      "build": "neutrino build",
      "lint": "neutrino lint"
   }
}
```

```js
// create src/index.js with following content
import React from 'react';
import { render } from 'react-dom';
import { AppContainer } from 'react-hot-loader';
import App from './App';

const load = () => render((
  <AppContainer>
    <App />
  </AppContainer>
), document.getElementById('root'));
```

---

That's quite a few steps to just get **started**. No custom config, sane defaults, and a quick start. I think these are reasonable.

That being said, I agree that maybe we are still going a bit too far in this PR, and we should be a little more agnostic. I'll mark the down in my review.

eliperelman:
@helfi92 were you trying to run `yarn init --yes`?

packages/create-neutrino-project/commands/init/generator/writing.js
```diff
@@ -0,0 +1,150 @@
...
+    scripts.lint = 'neutrino lint';
+  }
+
+  if (this.data.testRunner) {
+    scripts.test = 'neutrino test';
+  }
+
+  mkdirp.sync(this.options.directory);
+
+  const command = this.spawnCommand('bash', [
```
eliperelman: As noted in an earlier comment, don't use bash, spawn yarn or npm directly so it still works on Windows. Instead of `cd`, use `process.chdir(this.options.directory)`. There are also Yeoman commands to help with this:

http://yeoman.io/authoring/dependencies.html

packages/create-neutrino-project/commands/init/generator/writing.js
```diff
@@ -0,0 +1,150 @@
...
+    '-c',
+    `cd ${this.options.directory} && ${packageManger} init --force --yes`
+  ], { stdio: 'ignore' });
+
+  command.on('close', () => {
+    this.fs.extendJSON(path.resolve(this.destinationPath(this.options.directory), 'package.json'), { scripts });
```
eliperelman: I'm not sure if `extendJSON` is sync or async, but it might be easier to avoid a conflict if you `require(package.json)`, then add the values, then `fs.writefileSync(JSON.stringify(content, null, 2)\n)` when you're ready.

packages/create-neutrino-project/commands/init/generator/writing.js
```diff
@@ -0,0 +1,150 @@
...
+
+    return done();
+  });
+}
+
+module.exports = function () {
+  const destinationPath =  this.destinationPath(this.options.directory);
+  let templateDir = this.data.projectType;
+
+  if (this.data.router) {
```
eliperelman: Remove react-router.

packages/create-neutrino-project/commands/init/templates/react-router/src/App.js
```diff
@@ -0,0 +1,21 @@
+import React, { Component } from 'react';
```
eliperelman: Remove this template.

packages/create-neutrino-project/commands/init/templates/react-router/src/components/Home/Home.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
```
eliperelman: Remove.

packages/create-neutrino-project/commands/init/templates/react-router/src/components/Home/index.js
```diff
@@ -0,0 +1,5 @@
+import { asyncComponent } from 'react-async-component';
```
eliperelman: Remove.

packages/create-neutrino-project/commands/init/templates/react-router/src/index.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
```
eliperelman: Remove.

barraponto:
Just to clarify: is this going to be a neutrino command? Or a yeoman generator?
I'm all in favor of the latter, I was thinking about starting up a project to save my time by running `yo neutrino`. If we go down that path, let's make it easy to plug other presets (i.e. for Vue support).

eliperelman:
@barraponto it will be a Yeoman generator that will be run through the standard yarn or npx commands, so users won't need to install Neutrino, yo, or anything else to generate the minimal data needed to bootstrap a project.

```sh
yarn create neutrino-project my-app

npx create-neutrino-project my-app
```

> If we go down that path, let's make it easy to plug other presets (i.e. for Vue support).

Totally agree! We are going to merge this into a separate branch so we can iterate on this. I want to provide an "other" option for some of these questions so users can input their own repo or package name to use. And we would certainly use help with expat ding the number of projects we can bootstrap here, minimally of course.

packages/create-neutrino-project/commands/init/generator/writing.js
helfi92: When I stopped using the yeoman file system wrapper module (i.e., `this.fs`), it stopped prompting the user for permission to override the file :) I am now using [readJsonSync](https://github.com/jprichardson/node-fs-extra/blob/master/docs/readJson-sync.md) and [writeJsonSync](https://github.com/jprichardson/node-fs-extra/blob/master/docs/writeJson-sync.md) instead.

packages/create-neutrino-project/commands/init/generator/writing.js
helfi92: Thanks, `process.chdir` makes things a lot neater!

barraponto:
`yarn create neutrino-project` and `npx create-neutrino-project` are global installations even if it doesn't look like it.

eliperelman:
> `yarn create neutrino-project` and `npx create-neutrino-project` are global installations even if it doesn't look like it.

True, but these are tools that they will already have, yarn or npm/npx. We won't be forcing them to install intermediary tools, or making them learn proprietary other systems in order to start a neutrino project. It's the power of yeoman without the upfront steps to start using it.

Also, by all research I've done, yarn and npx keep the packages up to date before running.

If there are more cons, we should certainly address those sooner rather than later so we can avoid potential pitfalls or blockers.

packages/create-neutrino-project/commands/init/generator/install.js
```diff
@@ -0,0 +1,12 @@
+const chalk = require('chalk');
+const isYarn = require('./utils').isYarn;
```
eliperelman: This can be destructured.

packages/create-neutrino-project/commands/init/generator/writing.js
```diff
@@ -0,0 +1,142 @@
+const { ensureDirSync, readJsonSync, writeJsonSync } = require('fs-extra');
+const path = require('path');
+const stringify = require('javascript-stringify');
+const utils = require('./utils');
+
+const PROJECTS = utils.PROJECTS;
```
eliperelman: These can be destructured from the require.

packages/create-neutrino-project/commands/init/generator/writing.js
```diff
@@ -0,0 +1,142 @@
...
+
+function makeRcFile(data) {
+  let config = { use: [] };
+
+  config.use.push(data.projectType);
+
+  switch (data.linter) {
+    case LINTERS.AIRBNB:
+      config.use.push(
+        data.projectType === PROJECTS.WEB ? LIBRARY.NEUTRINO_PRESET_AIRBNB_BASE : LIBRARY.NEUTRINO_PRESET_AIRBNB
```
eliperelman: I think this should probably be:

```js
data.projectType === PROJECTS.REACT ? LIBRARY.NEUTRINO_PRESET_AIRBNB : LIBRARY.NEUTRINO_PRESET_AIRBNB_BASE
```

since we may want to use Airbnb-base on more than just web projects.

packages/create-neutrino-project/commands/init/generator/writing.js
```diff
@@ -0,0 +1,142 @@
...
+  switch (this.data.projectType) {
+    case PROJECTS.REACT:
+      dependencies.push(LIBRARY.REACT, LIBRARY.REACT_DOM, LIBRARY.REACT_HOT_LOADER);
+      devDependencies.push(LIBRARY.NEUTRINO_PRESET_REACT);
+      break;
+    case PROJECTS.REACT_COMPONENTS:
+      dependencies.push(
+        LIBRARY.REACT,
+        LIBRARY.REACT_DOM,
+        LIBRARY.NEUTRINO_PRESET_REACT_COMPONENTS,
```
eliperelman: I think this one and the next are supposed to be devDependencies, right?

packages/create-neutrino-project/commands/init/generator/writing.js
```diff
@@ -0,0 +1,142 @@
...
+      dependencies.push(LIBRARY.NEUTRINO_PRESET_TASKCLUSTER_WEB_LIBRARY);
+      break;
+  }
+
+  if (this.data.testRunner) {
+    devDependencies.push(this.data.testRunner);
+  }
+
+  if (this.data.linter) {
+    if (this.data.projectType === PROJECTS.WEB && this.data.linter === LINTERS.AIRBNB) {
```
eliperelman: I think this should be:

```js
if (this.data.projectType === PROJECTS.REACT && this.data.linter === LINTERS.AIRBNB) {
  devDependencies.push(LIBRARY.NEUTRINO_PRESET_AIRBNB);
```

packages/create-neutrino-project/commands/init/generator/utils.js
```diff
@@ -0,0 +1,32 @@
+module.exports.PROJECTS = {
+  REACT: 'neutrino-preset-react',
+  REACT_COMPONENTS: 'neutrino-preset-react-components',
+  WEB: 'neutrino-preset-web',
+  NODE: 'neutrino-preset-node',
+  WEB_LIBRARY: 'neutrino-preset-taskcluster-web-library'
+};
+
+module.exports.LIBRARY = {
```
eliperelman: Make this plural like the other names: `LIBRARIES`.

packages/create-neutrino-project/commands/init/generator/writing.js
```diff
@@ -0,0 +1,142 @@
...
+  ensureDirSync(this.options.directory);
+
+  const command = this.spawnCommand(installer, ['init', '--yes'], { cwd: path.resolve(this.options.directory), stdio: 'ignore'});
+
+  command.on('close', () => {
+    const jsonPath = path.join(this.options.directory, 'package.json');
+    const json = readJsonSync(`./${jsonPath}`);
+    const packageJson = Object.assign(json, { scripts });
+
+    writeJsonSync(jsonPath, packageJson, { spaces: 2 });
```
eliperelman: Does this function still put a new line at the end of the file?

packages/create-neutrino-project/commands/init/generator/writing.js
```diff
@@ -0,0 +1,142 @@
...
+
+    return done();
+  });
+}
+
+module.exports = function () {
+  const destinationPath =  this.destinationPath(this.options.directory);
```
eliperelman: Double space after equals.

packages/create-neutrino-project/commands/init/generator/writing.js
```diff
@@ -0,0 +1,142 @@
...
+    scripts.lint = 'neutrino lint';
+  }
+
+  if (this.data.testRunner) {
+    scripts.test = 'neutrino test';
+  }
+
+  ensureDirSync(this.options.directory);
+
+  const command = this.spawnCommand(installer, ['init', '--yes'], { cwd: path.resolve(this.options.directory), stdio: 'ignore'});
```
eliperelman: Missing extra space before closing curly brace.

packages/create-neutrino-project/bin/create-neutrino-project.js
```diff
@@ -0,0 +1,18 @@
...
+const dir = path.resolve(__dirname, `../commands/init`);
+
+env.register(require.resolve(dir), `create-neutrino-project:init`);
+
+const cli = yargs.command('<project-directory>')
+  .demandCommand(1, 'Only <project-directory> is required')
+  .help()
+  .argv;
+
+env.run(`create-neutrino-project:init`, { 'directory': cli._[0]}, done);
```
eliperelman: Missing extra space before closing curly brace.

packages/create-neutrino-project/commands/init/generator/install.js
```diff
@@ -0,0 +1,12 @@
...
+
+module.exports = function() {
+  const installer = isYarn ? 'yarnInstall' : 'npmInstall';
+  const devOptions = isYarn ? { dev: true } : { 'save-dev' : true};
```
eliperelman: Missing extra space before closing curly brace.

packages/create-neutrino-project/commands/init/generator/writing.js
```diff
@@ -0,0 +1,142 @@
...
+
+    return done();
+  });
+}
+
+module.exports = function () {
+  const destinationPath =  this.destinationPath(this.options.directory);
+  const templateDir = this.data.projectType;
+
+  initialPackageJson.call(this);
```
eliperelman: Why not `this.initialPackageJson()`?

packages/create-neutrino-project/commands/init/generator/writing.js
```diff
@@ -0,0 +1,142 @@
...
+  this.allDependencies = getDependencies.call(this);
```
eliperelman: Why not `this.allDependencies = this.getDependencies();`?

packages/create-neutrino-project/commands/init/generator/writing.js
```diff
@@ -0,0 +1,142 @@
...
+    const testDestinationDir = path.join(destinationPath, 'test');
+
+    ensureDirSync(testDestinationDir);
+
+    this.fs.copy(
+      this.templatePath(path.join('test', `${this.data.testRunner}.js`)),
+      path.join(testDestinationDir, 'simple_test.js')
+    );
+
+    this.fs.copy(
```
eliperelman: I think this is supposed to be in a different condition?

```js
if (this.data.linter) {
  this.fs.copy(
    this.templatePath('eslintrc.js'),
    path.join(destinationPath, '.eslintrc.js')
  );
}
```

packages/create-neutrino-project/commands/init/index.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+const Generator = require('yeoman-generator');
+const prompting = require('./generator/prompting');
+const writing = require('./generator/writing');
+const install = require('./generator/install');
+const end = require('./generator/end');
+
+class Project extends Generator {
+  constructor(args, opts) {
```
eliperelman: This constructor is redundant, so it can be removed.

packages/create-neutrino-project/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "create-neutrino-project": "./bin/create-neutrino-project.js"
+  },
+  "scripts": {
+    "start": "create-neutrino-project init"
+  },
+  "files": [
+    "bin",
+    "commands"
+  ],
+  "repository": "https://github.com/mozilla-neutrino/create-neutrino-project",
```
eliperelman: The repository format here will need to be updated to be a format similar to what the rest of the monorepo packages define.

eliperelman:
@edmorley I'd like more of your opinion on this as well if you have time. 😃

packages/create-neutrino-project/commands/init/generator/writing.js
helfi92: Yes, the `EOL` option object property defaults to `\n`. https://github.com/jprichardson/node-fs-extra/blob/master/docs/writeJson-sync.md.

packages/create-neutrino-project/commands/init/generator/writing.js
helfi92: Good catch :)

packages/create-neutrino-project/commands/init/generator/writing.js
eliperelman: Oh, it's because that function isn't actually a part of the generator prototype. You can just pass the context as an argument: `initialPackageJson(this)`.

helfi92: I left it as `initialPackageJson.call(this);`. I think it's neater than passing `this` as a parameter since it's not really one and might confuse the reader. Let me know if you think otherwise.

packages/create-neutrino-project/commands/init/generator/writing.js
```diff
@@ -0,0 +1,150 @@
...
+    return done();
+  });
+}
+
+module.exports = function () {
+  const destinationPath = this.destinationPath(this.options.directory);
+  const templateDir = this.data.projectType;
+
+  initialPackageJson.call(this);
+  this.allDependencies = getDependencies.call(this);
```
eliperelman: Another way to resolve this would be to make these class methods on the generator class. It wpould probably be easier to maintain as well to move all of these files into that Project class.

packages/create-neutrino-project/commands/init/index.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
```
eliperelman: No strict directives.

packages/create-neutrino-project/commands/init/generator/writing.js
helfi92: The methods are all under the Generator class. Note that they are prefixed with an underscore so that the generator doesn't automatically invoke them. [source](http://yeoman.io/authoring/running-context.html)

packages/create-neutrino-project/commands/init/index.js
helfi92: Out of curiosity, what is the reasoning behind removing the strict directives :)

packages/create-neutrino-project/commands/init/index.js
```diff
@@ -0,0 +1,187 @@
...
+  PROJECTS,
+  LIBRARIES,
+  LINTERS,
+  isYarn,
+  isObjectEmpty
+} = require('./utils');
+
+module.exports = class Project extends Generator {
+  _makeRcFile(data) {
+    let config = { use: [] };
```
eliperelman: This can be `const` since you aren't overwriting the value of `config`.

helfi92: ++

packages/create-neutrino-project/bin/create-neutrino-project.js
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env node
+
+const yeoman = require('yeoman-environment');
+const yargs = require('yargs');
+const path = require('path');
+
+const env = yeoman.createEnv();
+const done = exitCode => process.exit(exitCode || 0);
+const dir = path.resolve(__dirname, `../commands/init`);
```
eliperelman: Remove template strings from strings without interpolation.

packages/create-neutrino-project/bin/create-neutrino-project.js
```diff
@@ -0,0 +1,18 @@
...
+
+env.register(require.resolve(dir), `create-neutrino-project:init`);
+
+const cli = yargs.command('<project-directory>')
+  .demandCommand(1, 'Only <project-directory> is required')
+  .help()
+  .argv;
+
+env.run(`create-neutrino-project:init`, { 'directory': cli._[0] }, done);
```
eliperelman: Remove quotes from object property.

packages/create-neutrino-project/commands/init/index.js
```diff
@@ -0,0 +1,187 @@
...
+    const devDependencies = [];
+    const dependencies = [];
+
+    switch (this.data.projectType) {
+      case PROJECTS.REACT:
+        dependencies.push(LIBRARIES.REACT, LIBRARIES.REACT_DOM, LIBRARIES.REACT_HOT_LOADER);
+        devDependencies.push(LIBRARIES.NEUTRINO_PRESET_REACT);
+        break;
+      case PROJECTS.REACT_COMPONENTS:
+        dependencies.push(
```
eliperelman: For a react-components project, I think these are all dev dependencies.

packages/create-neutrino-project/commands/init/index.js
```diff
@@ -0,0 +1,187 @@
...
+        );
+        break;
+      case PROJECTS.WEB:
+        devDependencies.push(LIBRARIES.NEUTRINO_PRESET_WEB);
+        break;
+      case PROJECTS.NODE:
+        devDependencies.push(LIBRARIES.NEUTRINO_PRESET_NODE);
+        break;
+      case PROJECTS.WEB_LIBRARY:
+        dependencies.push(LIBRARIES.NEUTRINO_PRESET_TASKCLUSTER_WEB_LIBRARY);
```
eliperelman: devDependencies.

packages/create-neutrino-project/commands/init/index.js
```diff
@@ -0,0 +1,187 @@
...
+      {},
+      !isObjectEmpty(dependencies) ? { dependencies } : null,
+      !isObjectEmpty(devDependencies) ? { devDependencies } : null
+    );
+  }
+
+  _initialPackageJson() {
+    const done = this.async();
+    const installer = isYarn ? 'yarn' : 'npm';
+    const scripts = { 'build': 'neutrino build' };
```
eliperelman: Remove quotes from object property key.

packages/create-neutrino-project/commands/init/index.js
helfi92: @eliperelman How come? the [installation](https://github.com/eliperelman/neutrino-preset-react-components#installation) of neutrino-preset-react-components suggests that `react`, `react-dom`, `react-addons-css-transition-group` are added without the dev flag. Is that a mistake in the README? Hm, maybe I should only change `LIBRARIES.NEUTRINO_PRESET_REACT_COMPONENTS` to be a dev dependency?

packages/create-neutrino-project/commands/init/index.js
```diff
@@ -0,0 +1,187 @@
...
+        path.join(destinationPath, '.eslintrc.js')
+      );
+    }
+  }
+
+  install() {
+    const installer = isYarn ? 'yarn' : 'npm';
+    const argument = isYarn ? 'add' : 'install';
+    const development = isYarn ? '--dev' : '--save-dev';
+    const { dependencies , devDependencies } = this._getDependencies();
```
eliperelman: Remove extra space before destructuring comma.

packages/create-neutrino-project/commands/init/index.js
```diff
@@ -0,0 +1,187 @@
...
+
+    this.log(`${chalk.green('success')} Saved package.json`);
+    process.chdir(this.options.directory);
+    this.spawnCommandSync(installer, [argument, ...dependencies], { stdio: 'inherit' });
+    this.spawnCommandSync(installer, [argument, development, ...devDependencies], { stdio: 'inherit' });
+
+    if (this.data.linter) {
+      this.spawnCommandSync('neutrino', ['lint', '--fix'], { stdio: 'ignore' });
```
eliperelman: Did you test this out?

packages/create-neutrino-project/commands/init/index.js
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  _getDependencies() {
+    const devDependencies = [];
+    const dependencies = [];
+
+    switch (this.data.projectType) {
+      case PROJECTS.REACT:
+        dependencies.push(LIBRARIES.REACT, LIBRARIES.REACT_DOM, LIBRARIES.REACT_HOT_LOADER);
+        devDependencies.push(LIBRARIES.NEUTRINO_PRESET_REACT);
```
edmorley: I think all of these for React should be the same (either `dependencies` or `devDependencies`). Once built, the react packages are redundant. It's worth noting that Heroku only installs `dependencies` rather than `devDependencies` by default, so it really depends on the user's setup here.

packages/create-neutrino-project/yarn.lock
```diff
@@ -0,0 +1,1621 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
edmorley: This `yarn.lock` shouldn't exist. Are you using an old version of yarn that doesn't support workspaces? Please remove from the PR :-)

eliperelman: Good catch, yeah this is from moving the project from a standalone repo to be included in the monorepo.

packages/create-neutrino-project/commands/init/index.js
eliperelman: Agreed. I try to make the distinction about app dependencies vs. project/build/dev dependencies. What do you import in the app vs. what do you need to build the app?

packages/create-neutrino-project/commands/init/index.js
eliperelman: Yeah, I think that's a mistake in the readme. For example, if you look at the dependencies of react-lazylog and react-fout-stager:

https://github.com/mozilla-rpweb/react-lazylog/blob/6a10676edacedcf0fb84e6165ef75e33aa7fbc80/package.json#L27-L34

https://github.com/eliperelman/react-fout-stager/blob/624befe3fc7c7f019164fc71d975de2557e396da/package.json#L17-L24

You'll notice everything is in devDependencies that aren't direct for the component.

packages/create-neutrino-project/commands/init/index.js
edmorley: Going off on a bit of a tangent here - but IMO the real problem is that there are three buckets of dependencies (run-time, build-time, dev/testing) and only two places to put them (`dependencies`, `devDependencies`). And even 'build-time' is a bit nebulous, since that differs for people who commit the built assets to source control vs those who generate at deploy time.

eliperelman: Good point! Too bad you cant have user defined dependencies:

```json
{
  "dependencies": {
    "default": {
      "react": "*"
    },
    "build": {
      "neutrino": "*"
    },
    "test": {
      "neutrino-preset-react": "*"
    }
  }
}
```

Anyway, probably a pipe dream, but would be neat. :smile:

packages/create-neutrino-project/commands/init/index.js
helfi92: Yep.

packages/create-neutrino-project/bin/create-neutrino-project.js
```diff
@@ -0,0 +1,18 @@
...
+
+const yeoman = require('yeoman-environment');
+const yargs = require('yargs');
+const path = require('path');
+
+const env = yeoman.createEnv();
+const done = exitCode => process.exit(exitCode || 0);
+const dir = path.resolve(__dirname, '../commands/init');
+
+env.register(require.resolve(dir), `create-neutrino-project:init`);
```
eliperelman: This template string still needs to be a normal string. :)

packages/create-neutrino-project/bin/create-neutrino-project.js
```diff
@@ -0,0 +1,18 @@
...
+
+const cli = yargs.command('<project-directory>')
+  .demandCommand(1, 'Only <project-directory> is required')
+  .help()
+  .argv;
+
+env.run(`create-neutrino-project:init`, { directory: cli._[0] }, done);
```
eliperelman: This template string too.

packages/create-neutrino-project/commands/init/index.js
edmorley: This still needs updating

eliperelman: If we move neutrino deps to normal deps, it would go against our installation process:

https://www.npmjs.com/package/neutrino-preset-react#installation

eliperelman: Most projects out there still split their deps into app code vs. non-app code.

eliperelman: (Like Node.js projects)

edmorley: I would argue that those are wrong. `neutrino` and `neutrino-preset-react` are only ever needed at the same time as react-dom, there isn't a scenario when one is needed and the other not. For `neutrino-preset-airbnb-base`, that does belong in `devDependencies` on the other hand.

Though given that this PR is consistent with the docs, it seems fine to ignore for now, and potentially discuss alternatives in the future :-)

packages/create-neutrino-project/commands/init/templates/eslintrc.js
```diff
@@ -0,0 +1,3 @@
+const { Neutrino } = require('neutrino');
+
+module.exports = Neutrino().call('eslintrc');
```
edmorley: Could you update the docs to use this shorter `.eslintrc.js` template too?

packages/create-neutrino-project/commands/init/index.js
eliperelman: Yes, from a logical standpoint, you are correct. These are idioms that are borrowed from Node.js, but aren't quite congruent to the nature of building a front-end project, since the dependencies get bundled at build time, unlike building Node.js projects. We try to give *the appearance* of similarity by keeping app dependencies separate from non-app dependencies, but it's merely a facade. It's some semblance of consistency between our projects.

packages/create-neutrino-project/commands/init/templates/eslintrc.js
helfi92: Addressed in https://github.com/mozilla-neutrino/neutrino-dev/pull/412 :)

eliperelman:
:tada:

helfi92:
@eliperelman Want me to write some docs for it?

eliperelman:
Sure! Just submit a PR to the create-neutrino-project branch.

packages/create-neutrino-project/commands/init/index.js
barraponto: IMO, `neutrino-preset-react` is a development dependency, since it is only needed at build time. Although the preset itself could have `react-dom` as a peer dependency, indicating the user he/she should install that as well. That's what I do in https://github.com/barraponto/neutrino-preset-vue/blob/master/package.json

barraponto: (then, when the user installs it, it should go into actual dependencies)

PR #378, opened 2017-10-25
Output logging messages when using watch instead of devServer

### LINKED ISSUE #375: Incremental building is not performed on basic config
Test suite: https://mega.nz/#!HdMnVbRI!3s6VchrK3EFxzsUTF-DUVb5G5PvW4SGNYzepzI7gLCs

Here is a very simple custom preset for building a library in UMD format. I used `neutrino.config.watch(true)` and expect build process to restart on changes in File System. The goal is to rebuild library on any changes in a source code and write it to the hard drive.

Build task just is finished after successful build

```
> neutrino build

✔ Building project completed
Hash: 3ff1ed2317d535f3f013
Version: webpack 3.8.1
Time: 1355ms
     Asset       Size  Chunks             Chunk Names
    lib.js  877 bytes       0  [emitted]  index
lib.js.map    3.47 kB       0  [emitted]  index
```

Start task builds one time and doesn't react on any changes. You have to end process with Ctrl+C
```
> neutrino start

✔ Build completed
```

But none of these task listens for changes and incrementally rebuilds the project.

Is there a way to perform a `neutrino start` with saving to a hard drive and incremental building on changes?

eliperelman:
Fixes #375.

So, incremental building actually works, it just doesn't show a new console message every time like the dev server currently does. This PR should fix that.

### DIFF
--- a/packages/neutrino/bin/start.js
+++ b/packages/neutrino/bin/start.js
@@ -33,6 +33,13 @@ module.exports = (middleware, args) => {
 
       if (!compiler.options.devServer) {
         spinner.succeed('Build completed');
+        const building = ora('Performing initial build');
+
+        compiler.plugin('done', () => building.succeed('Build completed'));
+        compiler.plugin('compile', () => {
+          building.text = 'Source changed, re-compiling';
+          building.start();
+        });
       } else {
         const { devServer } = compiler.options;
         const url = `${devServer.https ? 'https' : 'http'}://${devServer.public}:${devServer.port}`;


packages/neutrino/bin/start.js
```diff
@@ -33,6 +33,13 @@ module.exports = (middleware, args) => {
 
       if (!compiler.options.devServer) {
         spinner.succeed('Build completed');
+        const building = ora('Source changed, re-compiling');
```
edmorley: Should this say "Waiting for initial build to finish" or similar, instead?

edmorley: Or "Performing initial build" etc

eliperelman: Oops, good catch.

eliperelman:
Updated.

edmorley (approved):

PR #385, opened 2017-10-27
Support multiple configurations from neutrinorc

### LINKED ISSUE #308: Support multiple Webpack configurations
In order to support multiple Webpack configurations, I think we need to re-think how we are allowing configurations to be composed. I still think the right approach is to let middleware modify configuration that is passed to it as an argument, but I think that this should not be a single config object directly accessed on the Neutrino object.

We can allow the user to name configuration objects using the API or `.neutrinorc.js`, and then Neutrino would be able to pass an array of configs to Webpack. The caveat here is that this is a breaking change, as we would probably need to change the signature of the middleware function to take a particular configuration object as a parameter as the user composes them.

Let me try and break down what I mean via some reduced examples:

`simple-middleware`

```js
module.exports = (config, options) => {
  config.devtool('source-map');
};
```

`.neutrinorc.js`

```js
module.exports = {
  react: {
    use: [
      // all the middleware to use for a react config
    ]
  },

  express: {
    use: [
      // all the middleware to use for an express config
    ]
  }
};
```

We could store these named configs on the Neutrino object, and remove `neutrino.config`. It could look something like:

```js
api.configs = {
  react: new Config(),
  express: new Config()
};
```

Then it would still be possible to target them by name, but we would definitely prefer users to compose their middleware via `.neutrinorc.js`. Maybe we even create a `default` middleware when a named one isn't specified.

Thoughts?

eliperelman:
**This is a WIP!**

Breaking change! This will allow us to finally support multiple configs from `.neutrinorc.js` by exporting an array.

```js
module.exports = [
  {
  },
  {
    options: {
      source: 'server'
    },
    use: ['neutrino-preset-node']
  }
];
```

This works by spawning a new process for each configuration. I tried going down 2 other routes but faced difficulties that were near impossible to overcome in a good way.

Wanted to get this out there for feedback.

Fixes #308.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,5 @@
 const { Neutrino } = require('./packages/neutrino');
 
-const api = Neutrino();
-
-module.exports = api.call('eslintrc');
+module.exports = Neutrino()
+  .use('.neutrinorc.js') // eslint-disable-line global-require
+  .call('eslintrc');
--- a/packages/neutrino-middleware-eslint/eslintrc.js
+++ b/packages/neutrino-middleware-eslint/eslintrc.js
@@ -1,6 +1,5 @@
 const { Neutrino } = require('../neutrino');
 
-const api = Neutrino({ cwd: __dirname });
-
-// eslint-disable-next-line global-require
-module.exports = api.call('eslintrc', [require('.')]);
+module.exports = Neutrino({ root: __dirname })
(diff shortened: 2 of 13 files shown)


edmorley:
> Fixes #385.

s/385/308/ ? :-)

eliperelman:
@edmorley lol oops.

eliperelman:
Closing this in favor of a different solution.

Issue #363, opened 2017-10-19
Problems testing React component


zeropaper:
I am trying to setup tests for a React component with the following dependencies
(npm version is 5.5.1, node is 8.3.0)

```
  "devDependencies": {
    "autoprefixer": "^7.1.5",
    "enzyme": "^3.1.0",
    "enzyme-adapter-react-16": "^1.0.2",
    "neutrino": "^7.1.1",
    "neutrino-middleware-postcss": "^3.0.0",
    "neutrino-middleware-pwa": "^7.1.1",
    "neutrino-preset-airbnb-base": "^7.1.1",
    "neutrino-preset-jest": "^7.1.3",
    "neutrino-preset-react": "^7.1.1",
    "react-test-renderer": "^16.0.0"
  },
  "dependencies": {
    "moment": "^2.19.1",
    "normalize.css": "^7.0.0",
    "open-sans-fontface": "^1.4.0",
    "react": "^16.0.0",
    "react-dom": "^16.0.0",
    "react-hot-loader": "^3.0.0"
  }
```

with the following test file content:

```js
import React from 'react';
import { shallow, mount, render } from 'enzyme';
import Calendar from './../src/components/calendar.jsx';

describe('simple', () => {
  it('should be sane', () => {
    expect(false).not.toBe(true);
  });
});

describe('<Calendar>', () => {
  it('renders 1 <Calendar /> component', () => {
    const wrapper = mount(<Calendar />);
  });
});
```

but get the following error:

```
$ npm test

> x-react-calendar@0.0.1 test .../x-react-calendar
> neutrino test

 FAIL  test/simple_test.js
  simple
    ✓ should be sane (4ms)
  <Calendar>
    ✕ renders 1 <Calendar /> component (2ms)

  ● <Calendar> › renders 1 <Calendar /> component

    
          Enzyme Internal Error: Enzyme expects an adapter to be configured, but found none. To
          configure an adapter, you should call `Enzyme.configure({ adapter: new Adapter() })`
          before using any of Enzyme's top level APIs, where `Adapter` is the adapter
          corresponding to the library currently being tested. For example:
    
          import Adapter from 'enzyme-adapter-react-15';
    
          To find out more about this, see http://airbnb.io/enzyme/docs/installation/index.html
      
          
      at validateAdapter (.../x-react-calendar/node_modules/enzyme/build/validateAdapter.js:14:11)
[...]

PR #388, opened 2017-10-29
add notes about setup scripts

zeropaper:
Related to #363


zeropaper:
My pleasure!
Thanks for the help :)

Issue #389, opened 2017-10-29
Connect more things to --debug


edmorley:
Neutrino has a `--debug` CLI option, which currently only affects two things:
* it enables `babel-preset-env` debug mode
* it causes `--inspect` to be passed to `start-server-webpack-plugin`

I've had a bit of a look, and there are a few more places where we could choose to vary behaviour when `--debug` is passed:

- [X] #391 copy-webpack-plugin: `debug` defaults to `warning`, but can also be set to `info` or `debug` ([docs](https://github.com/webpack-contrib/copy-webpack-plugin#available-options))
- [x] #392 clean-webpack-plugin: `verbose` defaults to `false`, but could be set to `true` ([docs](https://github.com/johnagan/clean-webpack-plugin#options-and-defaults-optional))
- [X] #754 neutrino-middleware-eslint: could enable eslint debug output [like this](https://github.com/eslint/eslint/blob/v4.10.0/bin/eslint.js#L20-L23)
- [x] #394 neutrino-preset-karma: `logLevel` defaults to `LOG_INFO`, but could be set to `LOG_DEBUG` ([docs](https://karma-runner.github.io/1.0/dev/public-api.html))
- [x] #523 neutrino-preset-jest: the jest API has `verbose` and `debug` options that could be enabled ([docs](https://facebook.github.io/jest/docs/en/cli.html#debug))
- [x] #485 ModuleConcatenationPlugin: the `stats` optimizationBailout option could be set ([docs](https://webpack.js.org/plugins/module-concatenation-plugin/#debugging-optimization-bailouts))
- [ ] neutrino-middleware-dev-server: The `noInfo` and `stats` configs could be adjusted ([docs](https://webpack.js.org/configuration/dev-server/#devserver-noinfo-) and [docs](https://webpack.js.org/configuration/stats/))
- [ ] neutrino build command: the `stats` config could be adjusted ([docs](https://webpack.js.org/configuration/stats/))

I'll open PRs for some of the above. If anyone wants to help out, or can think of others to add to the list, please say :-)

ahulab:
Hey I'd like to jump in and tackle as a first contribution if that's cool 
- neutrino-preset-jest: the jest API has verbose and debug options that could be enabled (docs)

helfi92:
@ahulab Go for it :)

edmorley:
I think we're mostly done here; anything else we can file separate issues for.

PR #394, opened 2017-10-30
Make --debug increase karma's logLevel from info to debug

edmorley:
This gives output of form:
https://emorley.pastebin.mozilla.org/9071511

See:
https://karma-runner.github.io/1.0/config/configuration-file.html
https://karma-runner.github.io/1.0/dev/public-api.html

Refs #389.

### DIFF
--- a/packages/neutrino-preset-karma/index.js
+++ b/packages/neutrino-preset-karma/index.js
@@ -1,4 +1,4 @@
-const { Server } = require('karma');
+const { Server, constants: { LOG_DEBUG, LOG_INFO } } = require('karma');
 const merge = require('deepmerge');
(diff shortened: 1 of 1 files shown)


edmorley:
So the extra output for this PR is a bit more than some of the others, however I think that's ok, since it only affects the `test` command, and is likely exactly what the user would expect after running `neutrino test --debug`.

edmorley:
Agreed - I've just been keeping an eye on what the output difference is after my first go at the eslint part to this resulted in tens of thousands of lines of output 😆

eliperelman:
@edmorley haha yeah, I've made that mistake before. 😆

PR #419, opened 2017-11-08
Docs: Remove leftover babel-polyfill references

edmorley:
Since the option was removed in #315.


eliperelman:
Looks good!

eliperelman:
Deployed docs, but not publishing new packages right now. We can wait until we have more commits.

PR #420, opened 2017-11-09
Return future when running a command

### LINKED ISSUE #415: `neutrino lint` has broken after upgrade to neutrino@7.3.0
I got different result with the same code.

## v7.2.3
image

## v7.3.0
image

ludat:
This should fix #415


ludat:
It looks like CI is failing because of linting errors on a file I didn't touch

ludat:
@eliperelman yup, you were exactly right.I just updated the PR

PR #426, opened 2017-11-14
Regression: allow defining neutrinorc as directory

### LINKED ISSUE #370: Allow .neutrinorc.js to be a directory
Right now we do a file-exists check on `.neutrinorc.js` to make sure we have the file before requiring it, but this prevents it from being a directory, which can be nice sometimes when you want to split it up into multiple files.

Just need to switch this check to be non-file-specific.

eliperelman:
Fixes #370.


packages/neutrino/src/utils.js
```diff
@@ -11,6 +11,16 @@ const createPaths = curry((root, moduleId) => List.of(
   moduleId
 ));
 
+// hasRc :: (String -> String) -> Boolean
```
helfi92: // `exists :: ...`

eliperelman: lol oops

PR #427, opened 2017-11-14
Import webpack-chain to neutrino-dev

### LINKED ISSUE #414: Migrate webpack-chain into monorepo
This would let us consolidate this API's development with the rest of Neutrino, and keep its versioning in sync. GitHub unfortunately won't let us carry over stars, but that's OK.

We can consider renaming as part of v8 as well, like `@neutrinojs/chain`, if we move forward with scoped packages.

eliperelman:
Fixes #414. This also imports the commit history from webpack-chain to neutrino-dev. Importing this also means we will probably start tracking with Neutrino's common versioning.


eliperelman:
Do you think merging this into the monorepo is worth it considering the version tracking against the rest of the packages? The webpack-chain project is different than many of the other packages, since Neutrino relies on it, but it does not rely on Neutrino.

helfi92:
I'm cool having it live outside by itself. Webpack-chain is a tool that is independent of Neutrino and can be used by anyone without having to deal with Neutrino. People using `webpack-chain` will definitely get confused by the constant version bumps it may get as a consequence of being inside Neutrino packages.

timkelty:
Agreed, I prefer it separate.

edmorley:
Yeah I could go either way - perhaps 60-40 to not merging.

eliperelman:
OK, I think right now we can not do this.

PR #434, opened 2017-11-15
Support options for server-side html

timkelty:
@eliperelman Starting this PR just for a spot for discussion

This initially started as https://github.com/timkelty/neutrino-packages/tree/master/packages/neutrino-preset-web-ssr, which extended `web` and reconfigured.

### Global changes
- [x] Always clean `output` dir, not just on build.
            This is an easy way for a server-side app to know if it should serve the revved assets (from manifest), or if you are running the dev-server.
- [x] Add `options.publicPath` and pass it to output.publicPath and devServer.publicPath
- [x] Normalize `devServer.publicPath` as a root-relative url
- [x] change `options.html` to `options.htmlTemplate` and make `options.html` a bool option
- [x] Make all paths predictable (non-revved) when `neutrino.options.command === 'start'` Applicable to anything that uses url-loader (https://github.com/mozilla-neutrino/neutrino-dev/pull/435)

### If `options.html === false`:
- [x] do not use `neutrinojs/html-template` and associated plugins
- [x] use `webpack-manifest-plugin`
- [x] If `options.devServer.proxy` is a string (eg `http://localhost:8080`), set defaults for `http-proxy-middleware` assuming user wants to proxy all requests but webpack assets


timkelty:
@eliperelman questions for you:

>>> Always clean output dir, not just on build.

Is there any drawback to this?

>>>  Make all paths predictable (non-revved) when neutrino.options.command === 'start' 

This is already the case with .js files. The problem is with anything that uses url/file-loader (@neutrino/image-loader, @neutrino/font-loader), as the default `name` for that is `[hash].[ext]`: https://github.com/webpack-contrib/file-loader#options

We want to do something like this:

```
  const urlLoaderOptions = {
    name: isBuild ? '[name].[hash].[ext]' : '[name].[ext]',
  };
```

My question is _where_ to make the changes. Here, or in @neutrino/image-loader and @neutrino/font-loader. I'm thinking the latter might make sense and be a better default for most people?

eliperelman:
> You think this is ok, is there any drawback to this?

I **think** this sounds OK, and would be more useful for the Node.js preset definitely.

packages/web/index.js
```diff
@@ -23,16 +25,30 @@ module.exports = (neutrino, opts = {}) => {
   const options = merge({
     env: [],
     hot: true,
-    html: {},
-    devServer: {
-      hot: opts.hot !== false
-    },
+    html: true,
```
eliperelman: Can you explain the rationale for using 2 flags of `html` and `htmlTemplate`, instead of `html: false` vs. `html: {}`?

packages/web/index.js
```diff
@@ -143,14 +163,19 @@ module.exports = (neutrino, opts = {}) => {
...
-      neutrino.use(clean, { paths: [neutrino.options.output] });
       neutrino.use(copy, {
         patterns: [{
           context: staticDir,
           from: '**/*',
           to: basename(staticDir)
         }]
       });
+
+      if (!options.html) {
```
eliperelman: We already have a `.when` condition for `options.html`, so we can use that as well:

```js
.when(options.html
  (config) => {
    neutrino.use(htmlTemplate, options.htmlTemplate);
    config.plugin('script-ext').use(ScriptExtHtmlPlugin, [{ defaultAttribute: 'defer' }]);
  },
  (config) => config.plugin('manifest').use(ManifestPlugin)
)
```

packages/web/index.js
timkelty: Yeah…I think you're right, it isn't necessary.

To be clear: You're suggesting a single `html` setting where if object, it would pass options to @neutrinojs/html-template and if false, would do "the rest" (exclude htmlTemplate, script-ext, include manifest), yes?

packages/web/index.js
timkelty: @eliperelman I had it there as to only include it when `neutrino.options.command === 'build'`

timkelty:
Ok, I refactored to just use a single `html` option.

Do you think this condition is good enough, or do we need to test if `options.html` is explicitly an object?

```
    .when(options.html, (config) => {
      neutrino.use(htmlTemplate, options.html);
```

Currently I've left the `ManifestPlugin` in the `.when(neutrino.options.command === 'build'`.
If you think we should include it regardless of the command, I'll put it in the other `.when(options.html)...` as you suggested.

packages/web/index.js
eliperelman: Yeah, I think that makes sense to me.

eliperelman:
@timkelty I think what you have is good enough, I'll see about another review.

timkelty:
> I'd have to look in more detail, but some of the changes in the other PR seem borderline breaking, depending on the weird and wonderful things other people are doing :-)

These changes _should_ be non-breaking, but I did a quick review of things that might, in some world, be considered breaking:

### Output is now _always_ cleared, not just on `build`

Not exactly sure how this might break for someone, but if it did, it would be isolated to dev.

### When building, image-loader/font-loader names will now be `[name].[hash].[ext]` instead of the default (`[hash].[ext]`). 

I think `[name].[hash].[ext]` is probably a better default for most, but if we wanted to be extra careful, we could keep it at the default (`[name].[ext]`) and change it in the next major release.

### Conditions for  `options.html === false`

This is the core of this PR, and should be non-breaking, as it would only affect someone that was already passing `false`, which isn't a valid option for `@neutrinojs/html-template` (though no errors are thown by passing it)

## https://github.com/mozilla-neutrino/neutrino-dev/pull/434/commits/489851f47c41c11d461e2e5fe20b110ee1c75e5d

This should be non-breaking, as the default is already `/`: https://webpack.js.org/configuration/dev-server/#devserver-publicpath- (which is what `resolve('/','./')` will resolve to

eliperelman:
Cleaning on everything would be a breaking change. This may not be wanted for test or other commands. For example, if someone is using the `lint` command, I'm pretty sure they don't want their build directory wiped out. I'm thinking we should leave clean for production only, just so it stays predictable.

Changing the names would also be breaking, which I'm fine with, as long as it goes into v8, and not v7.

timkelty:
> Cleaning on everything would be a breaking change. This may not be wanted for test or other commands. For example, if someone is using the lint command, I'm pretty sure they don't want their build directory wiped out. I'm thinking we should leave clean for production only, just so it stays predictable.
> 
> Changing the names would also be breaking, which I'm fine with, as long as it goes into v8, and not v7.

Good point. I can wrap those things up if we want to ship it with v7, otherwise it can just go with v8. Thoughts?

eliperelman:
Let's focus on v8, and if there is demand, we can backport to v7.

edmorley:
@timkelty - great summary of the changes, thank you! 

The one that made me slightly uneasy was the cleaning the `dist` directory one -- would make certain workflows (like "why is the production build broken but `neutrino start` working?") hard, since each time you switch back and forth the dist directory would be cleared.

timkelty:
Cleaning output on every command seemed drastic and not for everyone.
I put it back to just cleaning on `build`.

It's easy enough for users to clean on the `prestart` event if they need this (as I did): https://github.com/timkelty/neutrino-packages/blob/master/packages/neutrino-preset-web-ssr/index.js#L36-L42

timkelty:
@eliperelman think this should be ready for merge.

Also, now that I reverted back to cleaning output only on build, this should be non-breaking.

edmorley:
>     * resolve devServer.publicPath to be root relative

@timkelty - can you remember why devServer's `publicPath` is set to `resolve('/', publicPath)` rather than matching `config.output.publicPath`?

timkelty:
Yes…let me look over my notes and get back to you :)

timkelty:
Per the [dev-server docs](https://webpack.js.org/configuration/dev-server/#devserver-publicpath-):

> Make sure publicPath always starts and ends with a forward slash.

@edmorley Since our default `config.output.publicPath` is `./`, it was making sure it was _always_ root relative. Seems like it always should make sense to do so…are you running into something?

edmorley:
I'm reviewing the devServer options (it's one of the few areas I haven't yet fully spring-cleaned) and was wondering if we could simplify `publicPath` at all.

The [devServer docs](https://webpack.js.org/configuration/dev-server/#devserver-publicpath-) say that it defaults to `/`, though webpack's [output.publicPath](https://webpack.js.org/configuration/output/#output-publicpath) defaults to `""`.

edmorley:
Ah ok so reading up more on these two again:
https://webpack.js.org/configuration/output/#output-publicpath
https://webpack.js.org/configuration/dev-server/#devserver-publicpath-

...I believe what had me thinking there was more simplification possible, was this from the devserver docsL

> It is recommended that `devServer.publicPath` is the same as `output.publicPath`. 

But it fails to mention the CDN case, where people have to set `output.publicPath` to an absolute URL, which therefore needs to have the domain removed before it's suitable for use as `devServer.publicPath`.

I think we could:
* see if we can use the empty string instead of `'./'` in `@neutrinojs/web` (as you mentioned)
* add a comment explaining the need for the `url.resolve()` in `@neutrinojs/web`
* remove the unnecessary default `publicPath: '/'` in `@neutrinojs/dev-server`

I'll open a PR :-)

timkelty:
@edmorley Yep, that all sounds right to me.

PR #437, opened 2017-11-16
Cache rebuilds of compile-loader

### LINKED ISSUE #433: Use cache-loader with compile-loader
https://github.com/webpack-contrib/cache-loader

eliperelman:
Fixes #433.

### DIFF
--- a/docs/packages/compile-loader/README.md
+++ b/docs/packages/compile-loader/README.md
@@ -41,7 +41,8 @@ const compile = require('@neutrinojs/compile-loader');
 neutrino.use(compile, {
   include: [],
   exclude: [],
-  babel: {}
+  babel: {},
+  cache: {}
 });
 ```
 
@@ -53,7 +54,8 @@ module.exports = {
     ['@neutrinojs/compile-loader', {
       include: [],
       exclude: [],
-      babel: {}
+      babel: {},
+      cache: {}
     }]
   ]
 };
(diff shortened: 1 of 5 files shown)


edmorley (changes requested):
The cache-loader docs say that it is intended for use prior to other loaders, but in the same `use` block. I don't know if I'm just tired and misreading here, but I think this PR is adding it in it's own `use` block instead?

Also the webpack docs say:
> Note that there is an overhead for saving the reading and saving the cache file, so only use this loader to cache expensive loaders.

...so we should probably only use this for babel-loader.

It would also be good to have some rough timings before/after.

edmorley:
Oh and I noticed babel-loader has a `cacheDirectory` option - should that be used in addition to/instead of this?
https://github.com/babel/babel-loader#options

eliperelman:
We already have cacheDirectory in use for babel-loader, but yeah maybe this isn't needed. From my testing I thought I saw an improvement but I guess I should put together something more definitive.

Rwgarding the uses, they can be confusing to read sometimes, but these are all uses that belong to compile rule, so this single rule would now expose 2 uses, cache before babel. You can't use 2 loaders in a single use in webpack-chain since they are named, so the organization happens at the use level. This was a source of issue with webpack since configuration wasn't deterministic.

eliperelman:
Going to close this for now.

PR #438, opened 2017-11-16
web/node: Easier targeting via options.targets

timkelty:
It is currently rather verbose to override the `targets.browsers` option for `babel-preset-env`, and it requires you to overwrite the whole options object rather than just specifying browsers: https://neutrino.js.org/presets/neutrino-preset-web/#preset-options

This PR adds `options.browsers` to be able to specify directly.

Furthermore, when babel-preset-env@7.0 is released (currently at beta.32), this will make it easier for users that want to use a browserslist config file, as the new version of babel-preset-env will look for those automatically. In that case, we can support passing `false` or `undefined` to have babel-preset-env/browserslist do its thing.

### DIFF
--- a/docs/packages/node/README.md
+++ b/docs/packages/node/README.md
@@ -240,14 +240,16 @@ module.exports = {
       // Add additional Babel plugins, presets, or env options
       babel: {
         // Override options for babel-preset-env, showing defaults:
-        presetEnv: {
-          targets: { node: '6.10' },
-          modules: false,
-          useBuiltIns: true,
-          // These are excluded when using polyfills.async. Disabling the async polyfill
-          // will remove these from the exclusion list
-          exclude: ['transform-regenerator', 'transform-async-to-generator']
-        }
+        presets: [
+          ['babel-preset-env', {
+            targets: { node: '6.10' },
+            modules: false,
+            useBuiltIns: true,
+            // These are excluded when using polyfills.async. Disabling the async polyfill
+            // will remove these from the exclusion list
+            exclude: ['transform-regenerator', 'transform-async-to-generator']
+          }]
+        ]
       }
     }]
   ]
--- a/docs/packages/node/README.md
+++ b/docs/packages/node/README.md
@@ -232,18 +232,21 @@ module.exports = {
         async: true
       },
 
+      // Target specific versions via babel-preset-env
+      targets: {
+        node: '6.10'
+      }
+
       // Add additional Babel plugins, presets, or env options
       babel: {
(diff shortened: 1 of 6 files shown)


eliperelman:
The reason I have shied away from doing something like this has to do with parity of the Node.js preset. The phrase `targets` applies to both. Maybe I'm over-thinking this too, and it's OK to have a `browsers` option for web and a `node` option for node.

Using `targets` gives a common configuration, with the tradeoff being more verbose nesting/more complex nesting:

```js
['@neutrinojs/web', {
  targets: {
    browsers: [/* ... */]
  }
}]

['@neutrinojs/node', {
  targets: {
    node: /* ... */
  }
}]
```

Using `browsers`/`node` gives a simpler and more targeted configuration per preset, with the tradeoff being a divergent configuration, which isn't really a problem:

```js
['@neutrinojs/web', {
  browsers: [/* ... */]
}]

['@neutrinojs/node', {
  node: /* ... */
}]
```

timkelty:
Yeah I'd be fine with just exposing `options.targets` instead. In which case, we should add the option to both node and web preset. Do you agree?

eliperelman:
Let's get some more opinions. @mozilla-neutrino/core-contributors any thoughts?

timkelty:
I like the idea of maintaining parity with `@neutrinojs/node`.

I updated my PR to pass `targets` instead - to both node and web.

eliperelman:
@timkelty could you also add this option to the docs?

timkelty:
Will do.

timkelty:
Docs updated.

docs/packages/node/README.md
```diff
@@ -232,18 +232,21 @@ module.exports = {
...
 
+      // Target specific versions via babel-preset-env
+      targets: {
+        node: '6.10'
+      }
+
       // Add additional Babel plugins, presets, or env options
       babel: {
         // Override options for babel-preset-env, showing defaults:
         presetEnv: {
```
eliperelman: Hmm, there is no more `presetEnv` here, must be a typo. Can you update this to:

```js
babel: {
  presets: [
    ['babel-preset-env', {
```

?

edmorley:
Good idea! I was only thinking the other day that our configs were pretty verbose just to specify a reduced set of targets.

So long as the `docs/` and `packages/` copy of the docs are in sync this looks great :-)

docs/packages/node/README.md
timkelty: Ah yep - I wondered about that.

timkelty:
Docs are fixed!

PR #441, opened 2017-11-16
Web: Remove console, debugger in production

timkelty:
Since we only use minify in production, removing console and debugger seems like good defaults in a web context.

### DIFF
--- a/packages/web/index.js
+++ b/packages/web/index.js
@@ -30,7 +30,13 @@ module.exports = (neutrino, opts = {}) => {
     polyfills: {
       async: true
     },
-    babel: {}
+    babel: {},
+    minify: {
+      minify: {
+        removeConsole: true,
+        removeDebugger: true
+      }
+    }
   }, opts);
 
   Object.assign(options, {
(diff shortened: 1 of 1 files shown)


packages/web/index.js
```diff
@@ -30,7 +30,13 @@ module.exports = (neutrino, opts = {}) => {
     polyfills: {
       async: true
     },
-    babel: {}
+    babel: {},
+    minify: {
+      minify: {
```
eliperelman: There's a double nesting here.

packages/web/index.js
```diff
@@ -30,7 +30,13 @@ module.exports = (neutrino, opts = {}) => {
...
+        removeConsole: true,
```
eliperelman: Removing these would be a breaking change, just FYI.

packages/web/index.js
timkelty: This is intentional, as options.minify is what gets passed to the preset, which has options { minify, overrides}

packages/web/index.js
timkelty: Sure. As a first step, we could just allow the `minify` option to be passed, and not set the defaults.

Otherwise changing it looks something like:

```
neutrino.use('neutrino-preset-web);
neutrino.config
.when(isProduction, (config) => {
  config.plugin('minify').tap(() => [{
    removeConsole: true,
    removeDebugger: true,
  }]);
});
```

packages/web/index.js
eliperelman: Maybe we shouldn't allow overrides here to keep things simple? I could definitely see users getting confused about the double-word nesting here.

PR #443, opened 2017-11-16
Support extract-text for styles via options.extract

timkelty:
Notes:

- This extracts _by default_, so if you don't want to, you need to pass `options.extract = false`
- Currently, other style rule loaders (e.g. https://github.com/barraponto/neutrino-middleware-postcss) would need to come _before_ this one for the extraction to work.

### DIFF
--- a/packages/style-loader/index.js
+++ b/packages/style-loader/index.js
@@ -2,13 +2,12 @@ const ExtractTextPlugin = require('extract-text-webpack-plugin');
 const merge = require('deepmerge');
 
 module.exports = (neutrino, opts = {}) => {
-  const isBuild = neutrino.options.command === 'start';
   const options = merge({
     styleUseId: 'style',
     hot: true,
     extract: {
       plugin: {
-        filename: isBuild ? '[name].css' : '[name].[contenthash].css'
+        filename: neutrino.options.command === 'build' ? '[name].[contenthash].css' : '[name].css'
       }
     }
   }, opts);
--- a/packages/style-loader/index.js
+++ b/packages/style-loader/index.js
@@ -1,10 +1,66 @@
-module.exports = ({ config }, options = {}) => config.module
-  .rule(options.ruleId || 'style')
-    .test(/\.css$/)
-      .use(options.styleUseId || 'style')
-        .loader(require.resolve('style-loader'))
-        .when(options.style, use => use.options(options.style))
-        .end()
+const ExtractTextPlugin = require('extract-text-webpack-plugin');
+const merge = require('deepmerge');
+
+module.exports = (neutrino, opts = {}) => {
+  const isBuild = neutrino.options.command === 'start';
+  const options = merge({
+    styleUseId: 'style',
+    hot: true,
+    extract: {
+      plugin: {
+        filename: isBuild ? '[name].css' : '[name].[contenthash].css'
+      }
+    }
+  }, opts);
+
+  neutrino.config.module
+    .rule(options.ruleId || 'style')
+      .test(/\.css$/)
+      .use(options.styleUseId)
+         .loader(require.resolve('style-loader'))
+         .when(options.style, use => use.options(options.style))
+         .end()
       .use(options.cssUseId || 'css')
         .loader(require.resolve('css-loader'))
         .when(options.css, use => use.options(options.css));
+
+  if (options.extract) {
+    const styleRule = neutrino.config.module.rule('style');
+    const styleEntries = styleRule.uses.entries();
+    const useKeys = Object.keys(styleEntries).filter(key => key !== options.styleUseId);
+
+    options.extract.loader = Object.assign({
+      use: useKeys.map(key => ({
+        loader: styleEntries[key].get('loader'),
+        options: styleEntries[key].get('options')
+      })),
+      fallback: {
+        loader: styleEntries[options.styleUseId].get('loader'),
+        options: styleEntries[options.styleUseId].get('options')
+      }
+    }, options.extract.loader || {});
+
+    styleRule
+      .uses
+        .clear()
+        .end()
+      .when(options.hot, (rule) => {
+        rule.use(options.hotUseId || 'hot')
+          .loader(require.resolve('css-hot-loader'))
+          .when(options.hot !== true, use => use.options(options.hot));
+      });
+
+    ExtractTextPlugin
+      .extract(options.extract.loader)
+      .forEach(({ loader, options }) => {
+        styleRule
+          .use(loader)
+            .loader(loader)
+            .options(options);
+      });
+
+    neutrino.config
+      .plugin('extract')
+        .use(ExtractTextPlugin, [options.extract.plugin]);
+  }
+};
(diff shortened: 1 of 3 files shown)


packages/style-loader/index.js
```diff
@@ -1,10 +1,66 @@
-module.exports = ({ config }, options = {}) => config.module
+const ExtractTextPlugin = require('extract-text-webpack-plugin');
+const merge = require('deepmerge');
+const { pick, mapValues } = require('lodash');
```
eliperelman: We typically use ramda in place of lodash, so let's swap that out here, or use native implementations where possible.

packages/style-loader/index.js
```diff
@@ -1,10 +1,66 @@
...
+    const keys = Object.keys(styleEntries);
+    const useKeys = keys.filter(key => key !== options.styleUseId);
+    const fallbackKeys = keys.filter(key => key === options.styleUseId);
+    const loadersByKey = mapValues(styleEntries, use => ({
+      loader: use.get('loader'),
+      options: use.get('options')
+    }));
+
+    options.extract.loader = Object.assign({
+      use: Object.values(pick(loadersByKey, useKeys)),
```
eliperelman: Object.values isn't available in Node.js v6, so you'll have to map the values out:

```js
Object.keys(object).map(key => object[key])
```

packages/style-loader/index.js
```diff
@@ -1,10 +1,66 @@
...
+      fallback: Object.values(pick(loadersByKey, fallbackKeys))
+    }, options.extract.loader || {});
+
+    styleRule
+    .uses
```
eliperelman: Add 2 more spaces of indentation.

packages/style-loader/index.js
```diff
@@ -1,10 +1,66 @@
...
+      .clear()
+      .end()
+    .when(options.hot, (rule) => {
+      rule.use(options.cssHotLoaderUseId)
+        .loader(require.resolve('css-hot-loader'))
+        .when(options.cssHotLoader, use => use.options(options.cssHotLoader));
+    });
+
+    ExtractTextPlugin.extract(options.extract.loader)
```
eliperelman: Move `.extract` to the next line, and increase its indent 2 spaces.

packages/style-loader/index.js
```diff
@@ -1,10 +1,66 @@
...
+    .forEach(({ loader, options }) => {
+      styleRule
+      .use(loader)
```
eliperelman: Indent 2 more spaces.

packages/style-loader/index.js
```diff
@@ -1,10 +1,66 @@
...
+        .loader(loader)
+        .options(options);
+    });
+
+    neutrino.config.plugin('extract')
```
eliperelman: Move `.plugin` to the next line and increase indent 2 spaces from the line above it.

packages/style-loader/index.js
```diff
@@ -1,10 +1,66 @@
-module.exports = ({ config }, options = {}) => config.module
+const ExtractTextPlugin = require('extract-text-webpack-plugin');
+const merge = require('deepmerge');
+const { pick, mapValues } = require('lodash');
+
+module.exports = (neutrino, opts = {}) => {
+  const isBuild = neutrino.options.command === 'start';
+  const options = merge({
+    styleUseId: 'style',
+    cssHotLoaderUseId: 'cssHotLoader',
```
eliperelman: How would you feel about this just being `hot`, similar to the `hot` plugin?

timkelty: Sure. Do you mean the the option name (`cssHotLoaderUseId`) or the value `'cssHotLoader'`?

Not sure if we even need it configurable as an option, was just following suit with `options.styleUseId`, `options.cssUseId`

timkelty:
I think I've tidied up all the indentation from the review…the only one I wasn't sure about was > Move .extract to the next line, and increase its indent 2 spaces.

This isn't a `ChainedMap`, so I wasn't sure why we'd break and indent…`ExtractTextPlugin.extract() → array`

timkelty:
I refactored without lodash or ramda. If we end up wanting ramda, we'll probably just want `{ pick, map }`;

timkelty:
I'm now passing `options.hot` directly to `css-hot-loader` (if it isn't just `true`), which is I _think_ what you were getting at.

I'm also now passing `options.hot` from  preset-web.

packages/style-loader/index.js
```diff
@@ -1,10 +1,66 @@
...
+        options: styleEntries[options.styleUseId].get('options')
+      }
+    }, options.extract.loader || {});
+
+    styleRule
+      .uses
+        .clear()
+        .end()
+      .when(options.hot, (rule) => {
+        rule.use(options.cssHotLoaderUseId || 'cssHotLoader')
```
eliperelman: Yeah, my comment before wasn't very clear, my apologies. I was asking to name this ID from `cssHotLoader` to just `hot`, just the HotPlugin does in other middleware.

packages/web/index.js
```diff
@@ -24,6 +24,9 @@ module.exports = (neutrino, opts = {}) => {
     env: [],
     hot: true,
     html: {},
+    style: {
+      hot: opts.hot !== false
```
eliperelman: :heart_eyes:

packages/image-loader/index.js
```diff
@@ -13,7 +13,7 @@ module.exports = ({ config }, options = {}) => {
 
   config.module
     .rule('img')
-    .test(/\.(png|jpg|jpeg|gif)(\?v=\d+\.\d+\.\d+)?$/)
+    .test(/\.(png|jpg|jpeg|gif|webp)(\?v=\d+\.\d+\.\d+)?$/)
```
eliperelman: Looks like the master commit got re-included here. You may want to:

```sh
git checkout master
git pull upstream master
git checkout style-loader-extract
git checkout HEAD~1
git rebase master
git commit -m "Pass options.hot instead of options.cssHotLoader" --amend
git push -f origin style-loader-extract
```

timkelty: Whups - yeah I rebased from master

eliperelman:
@timkelty looks like the same thing happened to yargs 🤣

timkelty:
@eliperelman ok, should just have commits for this PR now.

Will this be for 8.x then?

eliperelman:
Yep! Extraction itself is also breaking, since it would modify hashes of file output.

timkelty:
Last thing, just for future reference - is this how you wanted this bit indented? https://github.com/timkelty/neutrino-dev/blob/10ca3bca6615bb8dae7bafe85dcdcd9a435d3586/packages/style-loader/index.js#L53-L60

I didn't expect you to want the `.extract` indented as it isn't a `ChainedMap` changing contexts like the others.

eliperelman:
@timkelty yes, my indentation preferences just rely on readability, method chaining, and context.

timkelty:
Yikes, that fell through the cracks! https://github.com/mozilla-neutrino/neutrino-dev/pull/443/commits/b777bb59ead5d539552706b2ac79dbf994368e50

PR #446, opened 2017-11-17
Minimize images in production via img-loader (imagemin)

timkelty:


edmorley:
Do we know what kind of impact this might have on build times?

Should people instead be minimising images once in the repo rather than at build time? (I wonder if a new "build size analyser" feature of Neutrino could include a "look for images that could be minimised" check)

timkelty:
@edmorley ↓

> Do we know what kind of impact this might have on build times?

In my experience it isn't bad, but could certainly be significant with lots of large assets.

> Should people instead be minimizing images once in the repo rather than at build time?

good question. All my experiences with Neutrino involve gitignoring the `build` folder, and building as part of the deploy process, so that doesn't really apply for me.

Furthermore, I always want my `source` assets pre-optimized, so I wouldn't want to commit optimized versions there.

Wonder if most people un-source `neutrino.options.output`?

timkelty:
@eliperelman Docs added.

timkelty:
I wonder if we should also include this as an option for `neutrino-middleware-copy`?

That way if you had [static assets](https://neutrino.js.org/presets/neutrino-preset-web/#static-assets) you could optimize those as well.

timkelty:
Maybe this is a case for adding a new bool API option, `neutrino.options.optimize`?

That way we could imagemin, babel-minify, cssnano - all based on that, rather than `process.env.NODE_ENV === 'production'` like we do now (of course, that would still be the default value).

eliperelman:
> Maybe this is a case for adding a new bool API option, `neutrino.options.optimize`?

I'm not sure, but I think it's worthy of discussion. @mozilla-neutrino/core-contributors @SeeThruHead any thoughts?

SeeThruHead:
seems like a good middleware idea no? with an option to turn it on in web preset?, and pass options through?

eliperelman:
@SeeThruHead it seems like several pieces of middleware already have some sort of optimization settings, or could have:

- minify plugin
- style
- sourcemaps
- copy

And I'm sure there's more. The question is: should we be able to shut off all optimization with an option, or continue to control is based on the command/environment you're using?

SeeThruHead:
ah i can see it being an option that defaults to true during production, unless you pass a value

edmorley:
What do we think the use-cases are for enabling/disabling these features? Off the top of my head:
* to speed up the build by disabling unwanted features
* to debug/workaround optimise/minify-only breakage

For debugging disabling all features in one go seems fine, but for speed-up/workaround I imagine people will want to disable piecemeal.

eliperelman:
> and set that option based on user input or node env

Agreed, e.g.:

```sh
neutrino build --options.optimize false
```

```js
module.exports = {
  options: {
    optimize: false
  }
};
```

> For debugging disabling all features in one go seems fine, but for speed-up/workaround I imagine people will want to disable piecemeal.

Also agree, and I don't see exposing an option as preventing that from happening. For example, if you wanted to disable everything, you could use the previous options, otherwise middleware should still take its own options:

```js
// middleware
  const options = merge({ optimize: neutrino.options.optimize }, opts);

  if (options.optimize) {
    neutrino.use(something);
  }
};

// usage
neutrino.use(middleware, { optimize: false });
```

Or you could manually remove/disable things:

```js
neutrino.config.plugins.delete('minify');
```

timkelty:
Ok, I've updated the PR:

- `neutrino.options.optimize` is now an api option, which defaults to true in production
- node, web, and image-loader now use `neutrino.options.optimize`

timkelty:
So the more I think about this, I think this might be better suited for a dedicated preset instead of being part of `image-loader`.

https://github.com/itgalaxy/imagemin-webpack provides not only a loader, but also a webpack plugin to optimize any images added by any _other_ plugins (e.g. favicons-webpack-plugin,  `copy-webpack-plugin`).

A dedicated preset would allow us to add the loader and the plugin - and include the preset only in production a bit easier. For example, in `@neutrinojs/web`, we need to make sure it is added _after_ `neutrino.use(copy)`, so it can optimize images in the static dirs.

eliperelman:
@timkelty I'm happy with whatever direction you want to take this.

timkelty:
OK - I think a separate MW is best, then.
I'll open up a new PR.

PR #448, opened 2017-11-17
Better terminal output when using the scaffolding app

helfi92:


helfi92:
When running the scaffolding app, users may have the tendency to run `yarn` or `npm install` after changing into the directory when in fact, it is not needed. This pull-requests makes it explicit, hopefully making people happier and less confused :)

helfi92:
screen shot 2017-11-17 at 10 01 08 am

packages/create-neutrino-project/commands/init/index.js
```diff
@@ -181,7 +191,7 @@ module.exports = class Project extends Generator {
   }
 
   end() {
-    this.log(`Success! Created ${this.options.directory} at ${process.cwd()}`);
+    this.log(`${chalk.green('Success')}! Created ${this.options.directory} at ${process.cwd()}`);
```
eliperelman: Put the exclamation point inside the green too. :)

helfi92:
Pushed a commit to it then rebased.

PR #453, opened 2017-11-17
Introduce official standardjs preset for consumption by create CLI

eliperelman:


jaridmargolin:
supports react?

eliperelman:
@jaridmargolin yep! tested it out using the react preset with JSX.

jaridmargolin:
Interesting... In the package I had originally set up (which I am happy to deprecate 👻), I remember needing to include `standard-react`: https://github.com/jaridmargolin/neutrino-contrib/blob/master/packages/neutrino-middleware-standardreact/src/index.js#L19

Issue #442, opened 2017-11-16
Use preact and vue in create-neutrino-project


eliperelman:
The preact preset seems pretty good as-is to add to the list, just need the scaffold bits.

I'd like the vue addition to be a single addition to neutrinorc instead of 2 with web+vue.

@fa7ad @barraponto I'm going to extend an offer if you so choose: you can add your presets to core if you like, under `@neutrinojs/preact` and `@neutrinojs/vue`. If that interests you, or you want to discuss more about it, let's do it!

helfi92:
Today I'll work on the scaffolding bit. I agree it's a good idea to make the vue preset a single addition to be consistent with the other presets neutrino offers.

fa7ad:
@eliperelman I'd be happy with the arrangement, just tell me what to do.

eliperelman:
@fa7ad nice! just create a directory in packages named `preact` with the contents of your current package, and we can go from there!

fa7ad:
@eliperelman check #454

eliperelman:
Completed with #484.

PR #454, opened 2017-11-18
Add preact preset to packages

fa7ad:
> just create a directory in packages named preact with the contents of your current package, and we can go from there!

#442 [(comment)](https://github.com/mozilla-neutrino/neutrino-dev/issues/442#issuecomment-345011749)


helfi92:
Thanks @fa7ad. @eliperelman, I will look into it probably tomorrow.

PR #455, opened 2017-11-19
Add official scoped preact preset, convert to latest Neutrino API

### LINKED ISSUE #260: add neutrino-preset-react-require to react preset
Any interest in adding https://github.com/tj/neutrino-preset-react-require out of the box? Shouldn't do any harm having it there even if people ignore it, but then they can get rid of the "code smell" of importing React each time which is nice!

eliperelman:
This PR merges in the preact preset to packages, with a little cleanup to match the latest Neutrino API. This also adds the automatic import of `h` so users don't need to do it:

```js
// not needed, this happens automatically
import { h } from 'preact';
```

In order to make the react preset consistent, I've also implemented the same auto-import of `createElement` from react:

```js
// not needed, happens automatically
import React from 'react';
```

This would fix #260.


edmorley:
I see a few small differences between this imported preset and the existing react one - I've filed #473 to sync them up at some point.

PR #457, opened 2017-11-19
Add vue preset

helfi92:


helfi92:
The first commit is the original code of the exported preset. I will send patches to this pull-request.

packages/vue/README.md
```diff
@@ -0,0 +1,253 @@
...
+```
+
+You can either serve or deploy the contents of this `build` directory as a static site.
+
+## Static assets
+
+If you wish to copy files to the build directory that are not imported from application code, you can place
+them in a directory within `src` called `static`. All files in this directory will be copied from `src/static`
+to `build/static`. To change this behavior, specify your own patterns with
+[@neutrinojs/copy](../../packages/copy/README.md).
```
eliperelman: This should be an absolute URL when in the package README.

packages/vue/src/index.js
```diff
@@ -0,0 +1,79 @@
...
+        },
+        rules: {
+          'vue/jsx-uses-vars': 2
+        }
+      }));
+  }
+
+  if (neutrino.config.plugins.has('stylelint')) {
+    neutrino.config.plugin('stylelint')
+      .tap(args => [
```
eliperelman: This seems like it ignores all other arguments. If there aren't any, we can be explicit about extending the first arg using destructuring, otherwise we could spread the rest of the args:

```js
.tap(([options, ...args]) => [
  merge(options, { ... }),
  ...args
]);
```

packages/vue/src/index.js
```diff
@@ -0,0 +1,79 @@
...
+  const lintRule = neutrino.config.module.rules.get('lint');
+  const compileRule = neutrino.config.module.rules.get('compile');
+
+  neutrino.config.module.rule('vue')
+    .test(LOADER_EXTENSIONS)
+    .use('vue')
+    .loader(require.resolve('vue-loader'))
+    .options(options);
+
+  if (compileRule && compileRule.uses.has('babel')) {
```
eliperelman: I'm confused about what this does. These isn't a compile loader here until `neutrino.use(web, options)`.

packages/vue/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@neutrinojs/vue",
+  "version": "7.3.2",
+  "description": "Neutrino preset adding Vue.js support.",
+  "main": "src/index.js",
+  "author": "Capi Etheriel <barraponto@gmail.com> (https://barraponto.blog.br)",
+  "license": "GPL-3.0",
```
eliperelman: In order for us to use the code as-is from the original repo, we need to re-license this new version as MPL-2.0. @barraponto do we have your permission for that?

helfi92:
Closing. Will continue in https://github.com/mozilla-neutrino/neutrino-dev/pull/484.

PR #459, opened 2017-11-21
Imagemin middleware

timkelty:
Started here: https://github.com/mozilla-neutrino/neutrino-dev/pull/446

This seemed better suited for a middleware, so that we can apply a single set of imagemin options to both loader assets, as well as any assets written from other plugins (eg copy-webpack-plugin).


packages/imagemin/index.js
```diff
@@ -0,0 +1,46 @@
+const merge = require('deepmerge');
+
+const imageminLoader = require.resolve('imagemin-webpack/imagemin-loader');
```
eliperelman: Group the `require`s together, and move the `resolve` to its own grouping:

```js
const ImageminWebpackPlugin = require('imagemin-webpack/ImageminWebpackPlugin');
const gifsicle = require('imagemin-gifsicle');
const svgo = require('imagemin-svgo');
const pngquant = require('imagemin-pngquant');
const mozjpeg = require('imagemin-mozjpeg');
const webp = require('imagemin-webp');

const imageminLoader = require.resolve('imagemin-webpack/imagemin-loader');
```

packages/web/index.js
```diff
@@ -155,5 +151,16 @@ module.exports = (neutrino, opts = {}) => {
         }]
       });
       config.output.filename('[name].[chunkhash].js');
+    })
+    .when(process.env.NODE_ENV === 'production', () => {
+      neutrino.use(chunk);
+      neutrino.use(minify);
+
+      if (options.imagemin) {
```
eliperelman: Is the goal to switch this to `neutrino.options.optimize` when it's available?

timkelty: Yes, but also to pass options if desired.

timkelty:
@eliperelman I rebased this off master.

Note: When this is merged into master, there will be a conflict that needs to resolved, as I've moved the `.when(process.env.NODE_ENV === 'production', () => {` below the `.when(neutrino.options.command === 'build', (config) => {`.

This is because we need the imagemin plugin to be applied _after_ the `@neutrinojs/copy` stuff, so those images are optimized as well.

packages/imagemin/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "version": "7.3.2",
+  "description": "Neutrino middleware for optimizing images with imagemin",
+  "main": "index.js",
+  "keywords": [
+    "neutrino",
+    "neutrino-middleware",
+    "image",
+    "imagemin"
+  ],
+  "author": "Eli Perelman <eli@eliperelman.com>",
```
eliperelman: Feel free to put your information here since you wrote this package. 😃

packages/imagemin/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "license": "MPL-2.0",
+  "repository": "https://github.com/mozilla-neutrino/neutrino-dev/tree/master/packages/imagemin",
+  "homepage": "https://neutrino.js.org",
+  "bugs": "https://github.com/mozilla-neutrino/neutrino-dev/issues",
+  "dependencies": {
+    "deepmerge": "^1.5.2",
```
eliperelman: Update this package to latest.

timkelty:
Ok, made those change, plus rebased off master again to mitigate the above mentioned conflicts.

packages/imagemin/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "main": "index.js",
+  "keywords": [
+    "neutrino",
+    "neutrino-middleware",
+    "image",
+    "imagemin"
+  ],
+  "author": "Tim Kelty <timkelty@gmail.com>",
+  "license": "MPL-2.0",
+  "repository": "https://github.com/mozilla-neutrino/neutrino-dev/tree/master/packages/imagemin",
```
eliperelman: Does this repo format actually work on npm?

timkelty: Hmmm not sure - I was just following suit with the rest of the middlewares: https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/web/package.json#L14

edmorley:
Do we know how deterministic the output images files are after being run though minification? It occurred to me that perhaps the process might not be deterministic and thus cause constant hash churn.

Continuing on from #446:

> > Should people instead be minimizing images once in the repo rather than at build time?
> 
> good question. All my experiences with Neutrino involve gitignoring the build folder, and building as part of the deploy process, so that doesn't really apply for me.
> 
> Furthermore, I always want my source assets pre-optimized, so I wouldn't want to commit optimized versions there.

To clarify, I don't mean store the Neutrino build directory in the repo. I mean outside the Neutrino build optimise your source directory assets once and commit them, to save the build time penalty every time. I agree one wouldn't want to lose the original Photoshop document or whatever, but normally that would be separate from the file being minimised anyway?

packages/imagemin/package.json
eliperelman: wtf. When I was viewing this diff, it showed a different format...don't mind me.

edmorley:
I think this is a great middleware, and I'll probably end up using it myself in certain projects. I'm just trying to decide whether it would be best for it to be opt-in or opt-out. Knowing impact on build times and image sizes would with calculating a cost-benefit.

timkelty:
good points, @edmorley.

> Do we know how deterministic the output images files are after being run though minification? It occurred to me that perhaps the process might not be deterministic and thus cause constant hash churn.

Not certain. In my experiences the hash's have been the same, but I don't know for sure.

> I'm just trying to decide whether it would be best for it to be opt-in or opt-out

I'd be fine with defaulting `options.imagemin` to false. This might also mitigate the potential heft added if people had huge images and weren't expecting it to exponentially affect their build time.

eliperelman:
If we end up making this opt-in instead of opt-out, then there is no value in embedding this into the web middleware. They should just manually add it to their `use`s:

```js
module.exports = {
  use: [
    '@neutrinojs/react',
    '@neutrinojs/imagemin'
  ]
};
```

timkelty:
@eliperelman good point…

I'm fine either way. Maybe leave it out of web until we have a better idea how it will impact builds?

timkelty:
@eliperelman @edmorley my most recent commit on this PR removes the middleware from web.

Reasoning:
- there are some fairly heavy deps involved (imagemin-mozjpeg, etc), that I could see people being annoyed with having to install if they didn't want neutrino to optimize images
- The plugin needs to be included AFTER any other plugins that add images to the manifest. Having it separate allows users to manage that themselves.
- The above mentioned concerns about build times and deterministic file names

edmorley:
Sounds good - we can always re-evaluate in the future depending on popularity :-)

packages/imagemin/README.md
```diff
@@ -0,0 +1,108 @@
...
+// Use with default options
+module.exports = {
+  use: ['@neutrinojs/imagemin']
+};
+
+// Usage showing default options
+module.exports = {
+  use: [
+    ['@neutrinojs/imagemin', {
+	imagemin: {},
```
eliperelman: Indentation here seems off.

packages/imagemin/README.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+- `imagemin`: Set options for [imagemin](https://github.com/imagemin/imagemin#options).
+- `plugin`: Set options for [imagemin-webpack](https://github.com/itgalaxy/imagemin-webpack#standalone-plugin)'s ImageminWebpackPlugin.
+- `rules`: Specify rules to apply imagemin to
+
+## Customization
+
+`@neutrinojs/imagemin` creates some conventions to make overriding the configuration easier once you are
+ready to make changes.
```
eliperelman: Looks like we are missing the table of conventions here.

packages/web/index.js
```diff
@@ -192,5 +182,15 @@ module.exports = (neutrino, opts = {}) => {
       }
 
       config.output.filename('[name].[chunkhash].js');
+    })
```
eliperelman: Looks like this was moved without modification, so we should probably reset this file to what is in master.

packages/imagemin/README.md
timkelty: Added a Rules and Plugins table.

packages/imagemin/README.md
```diff
@@ -0,0 +1,125 @@
...
+// Use with default options
+module.exports = {
+  use: ['@neutrinojs/imagemin']
+};
+
+// Usage showing default options
+module.exports = {
+  use: [
+    ['@neutrinojs/imagemin', {
+			imagemin: {},
```
eliperelman: Heh, looks like we may have some mixed tabs in here. Make each indentation here 2 spaces.

packages/imagemin/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "homepage": "https://neutrino.js.org",
+  "bugs": "https://github.com/mozilla-neutrino/neutrino-dev/issues",
+  "dependencies": {
+    "deepmerge": "^2.0.0",
```
eliperelman: Use deepmerge `^1.5.2`.

packages/imagemin/README.md
timkelty: Oh dear…sorry in the middle of changing-editors-hell 🥇

eliperelman:
:tada:

PR #462, opened 2017-11-21
fix(deps): update dependency html-webpack-template to ^6.1.0

eliperelman:


edmorley:
I was puzzled to see 6.1.0 given the package's repo says 6.0.2 is the latest, but looks like the Git tag is missing. Have filed jaketrent/html-webpack-template#64.

PR #467, opened 2017-11-21
Add font, image options.

timkelty:
Adds a couple more options, allows exclusion with `false`.

Related: what do you think about grouping these under `options.loaders`?

```
  const options = merge({
    env: [],
    hot: true,
    html: {},
    publicPath: './',
    devServer: {
      hot: opts.hot !== false
    },
    polyfills: {
      async: true
    },
    minify: {},
    manifest: {},
    babel: {},
    targets: {},
    loaders: {
      html: {},
      font: {},
      image: {},
      style: {
        hot: opts.hot !== false
      }
    }
  }, opts);
```

Might not be necessary, but could clear up confusion (e.g. `options.html` vs `options.loaders.html`)

### DIFF
--- a/docs/packages/web/README.md
+++ b/docs/packages/web/README.md
@@ -160,6 +160,29 @@ module.exports = {
         async: true
       },
 
+      // Sets Webpack's `output.publicPath` and
+      // `devServer.publicPath` settings. Useful if you want to
+      // serve assets from a non-root location (e.g. `/assets/`)
+      publicPath: './',
+
+      // Change options for @neutrinojs/style-loader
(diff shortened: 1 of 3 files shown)


eliperelman:
Using the term `loaders` alludes to a structure that is used internally by the preset, but which the user may not necessarily care about. I'd like to keep these top-level if possible.

edmorley:
LGTM with docs additions :-)

timkelty:
Yep, I'll add docs.

timkelty:
Ok docs added.

PR #476, opened 2017-11-21
Add yarn script helper for linking all packages

eliperelman:


package.json
```diff
@@ -14,6 +14,8 @@
     "packages/*"
   ],
   "scripts": {
+    "bootstrap": "lerna bootstrap",
```
edmorley: The `bootstrap` step was removed in #372 since my understanding was that lerna's bootstrap itself just runs `yarn install`, which will have already occurred. Does it do something else that I've missed?

eliperelman: It's possible they do the same thing, `yarn` and `yarn lerna bootstrap`, but I've been using both. I'll try it out and see what happens.

eliperelman: LOL you're right. I just nuked everything with `yarn lerna clean`, then `yarn` and `yarn test` and everything passes. I'll revise.

PR #478, opened 2017-11-21
Turn the migration guide into a page that covers all versions

helfi92:


helfi92:
@eliperelman This doesn't include v7-> v8. Let's merge what we have now then we can open a new PR about migrating neutrino from v7 -> v8.

helfi92:
@eliperelman @edmorley I used something like `git show HEAD~225:docs/upgrading-neutrino.md` to fetch old documentation. Is that what you folks typically use? Curious to know if there exists a better options :)

edmorley:
Another option is the latest of each set of major version tags?

PR #483, opened 2017-11-21
Fix inconsistencies in the react and preact docs

### LINKED ISSUE #473: Sync up differences between the react and preact packages
Diffing `packages/react/` and `packages/preact/` shows a few differences between the two, some of which are necessary but it would be good to sync the rest up.

eliperelman:
Fixes #473.


edmorley:
> Fixes #473.

This will close that issue, but there were also a few non-docs changes too (eg missing `.npmignore`, differences in the way the babel options are merged)? :-)

eliperelman:
@edmorley I've added the npmignore, but the merging of babel options is different because of the ordering of react-hot-loader, which is different from the way the preact preset works.

PR #484, opened 2017-11-22
Vue preset

helfi92:


packages/neutrino/src/api.js
```diff
@@ -26,14 +26,26 @@ const pathOptions = [
...
     env: {
       NODE_ENV: 'development'
     },
     debug: false,
-    quiet: false
+    quiet: false,
+    extensions
   }, opts);
 
+  Object.defineProperty(options, 'extensions', {
```
helfi92: @eliperelman Let me know if this is the way you wanted it to be done. Once we've agreed on a way, I could update the rest neutrino to use it.

eliperelman: As discussed in IRC, this can just be a normal property, and consumers can access the Set directly. We can also expose a method for generating a regex on the fly, but we don't want to limit this property to only creating regexes.

packages/vue/index.js
```diff
@@ -0,0 +1,56 @@
...
+
+  neutrino.config.module.rule('vue')
+    .test(neutrino.options.extensions)
+    .use('vue')
+    .loader(require.resolve('vue-loader'))
+    .options(options);
+
+  neutrino.config.when(neutrino.config.module.rules.has('lint'), () => neutrino.config.module
+    .rule('lint')
+    .test(neutrino.options.extensions)
```
eliperelman: Use the loader merge here:

```js
neutrino.config.when(neutrino.config.module.rules.has('lint'), () => {
  neutrino.use(loaderMerge('lint', 'eslint'), {
    plugins: ['vue'],
    envs: ['node'],
    parserOptions: {
      ecmaFeatures: {
        jsx: true
      }
    },
    rules: {
      'vue/jsx-uses-vars': 2
    }
  });
});
```

packages/vue/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "neutrino",
+    "neutrino-preset",
+    "vue",
+    "web",
+    "hot module replacement"
+  ],
+  "dependencies": {
+    "@neutrinojs/web": "^7.3.2",
+    "babel-preset-vue": "^1.2.1",
+    "deepmerge": "^1.4.4",
```
eliperelman: Not sure where this version came from, but everywhere else it's at `^1.5.2`.

packages/vue/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "eslint-plugin-vue": "^2.1.0",
+    "stylelint-processor-html": "^1.0.0",
+    "vue-loader": "^13.5.0",
+    "vue-template-compiler": "^2.5.6"
```
eliperelman: Is this a peerDependency of something? I don't see this used anywhere.

packages/vue/index.js
```diff
@@ -0,0 +1,56 @@
+const web = require('@neutrinojs/web');
+const path = require('path');
+const merge = require('deepmerge');
+
+const MODULES = path.join(__dirname, 'node_modules');
+
+module.exports = (neutrino, options) => {
+  neutrino.use(web, options);
+
+  neutrino.config.module.rule('vue')
```
eliperelman: Use indented chaining when contexts change:

```js
neutrino.config.module
  .rule('vue')
    .test(neutrino.options.extensions)
    .use('vue')
      .loader(require.resolve('vue-loader'))
      .options(options);
```

packages/neutrino/src/api.js
eliperelman: Another thing, a user should be able to pass this as an array to Neutrino `options`, which when merged using `mergeOptions` or whatever, can be merged with the Set values:

```js
module.exports = {
  options: {
    extensions: ['vue']
  }
};

// ...

new Set([...this.options.extensions, ...extensions])
```

When merging we should also probably normalize the values to ensure they don't contain dots.

packages/vue/package.json
helfi92: Yeah, it's a peerDependency of `vue-loader` https://github.com/vuejs/vue-loader/blob/master/package.json#L60.

helfi92:
@eliperelman You mention [here](https://github.com/mozilla-neutrino/neutrino-dev/pull/484#discussion_r153276722) to have extensions merged i.e., `new Set([...this.options.extensions, ...extensions])`. I'm not sure why we need to merge the extensions. Sometimes you want to overwrite the entries instead of appending to the set. Let me know what you think about my last commit.

eliperelman:
> I'm not sure why we need to merge the extensions. Sometimes you want to overwrite the entries instead of appending to the set.

I see this as being the primary use case:

```js
module.exports = {
  options: {
    extensions: ['esm']
  }
};
```

Basically, a user can come in and add their own extensions, without having to replicate the whole list of other extensions that we store. Hence needing to merge instead of replace.

A replacement can still happen though if you're using the API, which seems good because I think it's a less common need, and only happens in more advanced cases:

```js
neutrino.extensions = new Set(['js', 'jsx']);
```

helfi92:
@eliperelman, In my last commit, something like the following:

```js
module.exports = {
  options: {
    extensions: ['.vue', 'vue'],
  },
  ...
};
```

would make `neutrino.options.extensions` equal `[ 'js', 'jsx', 'vue', 'ts', 'mjs', 'vue' ]`.

Whereas something like `neutrino.options.extensions = ['vue']` overwrites the default ones. Does it make more sense now? Thanks for your patience.

packages/neutrino/src/api.js
```diff
@@ -26,14 +27,28 @@ const pathOptions = [
...
+    enumerable: true,
+    get() {
+      return [...fileExtensions];
+    },
+    set(extensions) {
+      const newExtensions = extensions.map(ext => equals(ext[0], '.') ? tail(ext) : ext);
+
+      fileExtensions = [...new Set(newExtensions)];
```
eliperelman: Looks like the original value of `fileExtensions` is a Set, but here we spread it into an array. We can keep this as a Set here, and only spread to array during `get`:

```js
fileExtensions = new Set(newExtensions);
```

packages/neutrino/src/api.js
```diff
@@ -26,14 +27,28 @@ const pathOptions = [
...
+    extensions: fileExtensions
   }, opts);
 
+  Object.defineProperty(options, 'extensions', {
+    enumerable: true,
+    get() {
+      return [...fileExtensions];
+    },
+    set(extensions) {
+      const newExtensions = extensions.map(ext => equals(ext[0], '.') ? tail(ext) : ext);
```
eliperelman: This could be simplified to:

```js
const newExtensions = extensions.map(ext => ext.replace('.', ''));
```

packages/neutrino/src/api.js
eliperelman: Combining this with my previous comment, this can be one-lined into:

```js
fileExtensions = new Set(extensions.map(ext => ext.replace('.', '')));
```

packages/neutrino/src/api.js
```diff
@@ -26,14 +27,28 @@ const pathOptions = [
 
 // getOptions :: Object? -> IO Object
 const getOptions = (opts = {}) => {
+  let fileExtensions = new Set(['js', 'jsx', 'vue', 'ts', 'mjs']);
```
eliperelman: Let's rename this to `moduleExtensions` so it's clear we are only dealing with module code instead of things like images and stylesheets.

packages/neutrino/src/api.js
```diff
@@ -87,6 +102,10 @@ class Api {
     this.config = new Config();
   }
 
+  regexFromExtensions(extensions = this.options.extensions) {
+    return new RegExp(`.(${extensions.join('|')})$`);
```
eliperelman: :heart:

packages/vue/index.js
```diff
@@ -0,0 +1,57 @@
+const web = require('@neutrinojs/web');
+const path = require('path');
+const merge = require('deepmerge');
+const loaderMerge = require('@neutrinojs/loader-merge');
+
+const MODULES = path.join(__dirname, 'node_modules');
+
+module.exports = (neutrino, options) => {
+  neutrino.use(web, options);
+  neutrino.options.extensions = ['vue']; // eslint-disable-line no-param-reassign
```
eliperelman: We now have `vue` in the extensions list, so we don't need this line now.

packages/vue/index.js
```diff
@@ -0,0 +1,57 @@
+const web = require('@neutrinojs/web');
+const path = require('path');
+const merge = require('deepmerge');
+const loaderMerge = require('@neutrinojs/loader-merge');
```
eliperelman: nit: move this require to be above the `web` one.

packages/vue/index.js
```diff
@@ -0,0 +1,57 @@
...
+          }
+        }),
+        ...args
+      ]);
+  }
+
+  neutrino.config
+    .resolve
+      .modules.add(MODULES).end()
+      .extensions.add('.vue').end()
```
eliperelman: We should make this automatic from the `web` preset:

```js
neutrino.config.resolve.extensions.merge(neutrino.options.extensions.map(ext => `.${ext}`);
```

Something like that.

packages/vue/index.js
helfi92: If I include the `js` extension, I get an error:

```
ERROR in   Error: The loader "/Users/haali/Documents/Mozilla/projects/neutrino-dev/node_modules/html-webpack-plugin/lib/loader.js!/Use  rs/haali/Documents/Mozilla/projects/neutrino-dev/node_modules/html-webpack-template/index.ejs" didn't return html.
```

If extensions is something like `['jsx', 'vue', 'ts', 'mjs' ]` (without the js extension), then it works. Do you know why this is happening?

helfi92: Also, should we include `.json` in the list of extensions? :)

eliperelman: Looks like this:

https://github.com/jantimon/html-webpack-plugin/issues/602

Basically, I think the vue-loader is only supposed to accept `vue` extensions:

```js
neutrino.config.module
    .rule('vue')
      .test(/\.vue$/) // instead of .test(neutrino.regexFromExtensions())
      .use('vue')
        .loader(require.resolve('vue-loader'))
        .options(options);
```

eliperelman: > Also, should we include .json in the list of extensions? :)

I...don't know. Give it a shot and see if the tests continue to pass, but I'm not sure what effect it may have.

helfi92: That's what I figured, sounds good!

Btw, instead of hardcoding `/\.vue$/`, we could just do `neutrino.regexFromExtensions(['vue'])`.

eliperelman: :heart_eyes:

helfi92:
@eliperelman Adding `.json` to the list of extensions seems to work. Added the requested changes. Hopefully, things are neater :) Once that is done, I could clean the other packages to use the list of extensions. Also, to answer your question, I have tested merging extensions from `.neutrinorc.js`. The [example shown earlier](https://github.com/mozilla-neutrino/neutrino-dev/pull/484#issuecomment-347727223) was the one I tried on my machine.

packages/neutrino/src/api.js
```diff
@@ -26,14 +26,26 @@ const pathOptions = [
...
 const getOptions = (opts = {}) => {
+  let moduleExtensions = new Set(['js', 'jsx', 'vue', 'ts', 'mjs', 'json']);
   const options = merge({
     env: {
       NODE_ENV: 'development'
     },
     debug: false,
-    quiet: false
+    quiet: false,
+    extensions: moduleExtensions
```
eliperelman: Remove this line.

packages/vue/README.md
```diff
@@ -0,0 +1,256 @@
...
+
+[![NPM version][npm-image]][npm-url]
+[![NPM downloads][npm-downloads]][npm-url]
+[![Join the Neutrino community on Spectrum][spectrum-image]][spectrum-url]
+
+## Features
+
+- Zero upfront configuration necessary to start developing and building a Vue web app
+- Modern Babel compilation.
+- Extends from [@neutrinojs/web](https://neutrino.js.org/packages/web)
```
eliperelman: Copy this file to `docs/packages/vue/README.md` then replace these links with relative ones in that file, e.g.

`https://neutrino.js.org/packages/web` -> `../web`
`https://neutrino.js.org/project-layout` -> `../../project-layout`

PR #485, opened 2017-11-23
Debug optimization bailouts

helfi92:
Add helpful debug output when using ModuleConcatenationPlugin.

Relevant documentation: https://webpack.js.org/plugins/module-concatenation-plugin/#debugging-optimization-bailouts

### DIFF
--- a/packages/library/index.js
+++ b/packages/library/index.js
@@ -139,11 +139,18 @@ module.exports = (neutrino, opts = {}) => {
         neutrino.use(loaderMerge('lint', 'eslint'), { envs: ['browser', 'commonjs'] });
       }
     })
-    .when(process.env.NODE_ENV === 'production', () => {
+    .when(process.env.NODE_ENV === 'production', (config) => {
       neutrino.use(minify);
-      neutrino.config
-        .plugin('module-concat')
-          .use(optimize.ModuleConcatenationPlugin);
+      config.plugin('module-concat').use(optimize.ModuleConcatenationPlugin);
+
+      if (neutrino.options.debug) {
+        config.merge({
+          stats: {
+            maxModules: Infinity,
+            optimizationBailout: true
+          }
+        });
+      }
     })
     .when(neutrino.options.command === 'build', () => {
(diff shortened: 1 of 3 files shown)


edmorley:
Many thanks for doing this! I'd run out of time in #389 to try this one :-)

packages/library/index.js
```diff
@@ -139,11 +139,18 @@ module.exports = (neutrino, opts = {}) => {
...
     })
-    .when(process.env.NODE_ENV === 'production', () => {
+    .when(process.env.NODE_ENV === 'production', (config) => {
       neutrino.use(minify);
-      neutrino.config
-        .plugin('module-concat')
-          .use(optimize.ModuleConcatenationPlugin);
+      config.plugin('module-concat').use(optimize.ModuleConcatenationPlugin);
+
+      if (neutrino.options.debug) {
```
eliperelman: I think we'd want to do this regardless of being in `production`, right? If they are using `debug`, we should also use these I think...

helfi92: Sure, we can remove the conditional. I'll send a patch now.

eliperelman: Sorry, I think you may have misread my comment. We want these options to be set always when using debug, not always using production. It needs to come out of this `when` and go into its own when using `debug`.

helfi92: I definitely misread the comment. Sorry for the confusion.

eliperelman: No worries at all. 😃

packages/library/index.js
```diff
@@ -131,6 +131,12 @@ module.exports = (neutrino, opts = {}) => {
           .end()
         .end()
       .end()
+    .merge({
```
eliperelman: I think these are still supposed to be wrapped in a debug:

```js
.when(neutrino.options.debug, (config) => {
  config.merge({
    stats: {
      maxModules: Infinity,
      optimizationBailout: true
    }
  });
})
```

helfi92: My bad...

PR #486, opened 2017-11-23
Add webpack bundle information when running in debug

helfi92:
Relevant documentation: https://webpack.js.org/configuration/dev-server/#devserver-noinfo-


packages/neutrino/src/devServer.js
```diff
@@ -3,8 +3,8 @@ const Future = require('fluture');
 const { serve, validate } = require('./webpack');
 
 // server :: Object config -> Future () Function
```
eliperelman: The signature of this function needs to change to:

```js
// devServer :: (Object config -> Object api) -> Future () Function
```

PR #487, opened 2017-11-23
Support multiple entry points via new mains options, remove entry option

### LINKED ISSUE #367: Add support for multiple entry-points
In #326 the ability to manually use multiple instances of certain types of middleware (including `neutrino-middleware-html-template`) was added, which helped reduce the boilerplate mentioned there.

However configuring Neutrino to use multiple entry-points is still pretty painful, due to the number of other parts of the build that reference `neutrino.options.entry`.

For example one has to:
1. Remove or modify the existing `index` entrypoint
2. Remove or modify the existing `html` plugin (from `neutrino-middleware-html-template`)
3. Create entrypoints and HTML plugins for each desired page, remembering to set the `chunks` option to `html-webpack-plugin` correctly (ie: using the same hardcoded vendor/runtimes names as in `neutrino-middleware-chunk`. These names have changed between Neutrino versions, so having to keep them in sync is a pain).
4. If using `neutrino-middleware-dev-server` (such as via `neutrino-preset-{react,web}`, remember to prepend `webpack/hot/dev-server` (if `options.hot`) or else `webpack-dev-server/client` to each entrypoint.
5. If using `neutrino-preset-react`, remember to prepend `react-hot-loader/patch` to each entrypoint.
6. If using `neutrino-middleware-pwa` remember to append the relevant PWA script.

...plus account for any other customisations that third-party middleware might make.

This results in a lot of boilerplate, and increases the potential for it to get out of sync with Neutrino.

I was thinking perhaps the Neutrino API could be made to support `neutrino.options.entry` being either a string or an object, with the string case behaving as it does now, and the object being used like so:
```js
  options: {
    entry: {
      index: './index.js',
      foo: './foo.js',
      bar: './bar.js'
    }
  }
```

The middleware would then be made to iterate the entrypoints when making modifications. For example `neutrino-middleware-html-template` would act as it does now, except use plugin names of `html-index`, `html-foo` etc, and also set the `chunks` property correctly. And `neutrino-preset-react` would add the react hot loader to all entrypoints etc.

One question would be whether all consumers of `neutrino.options.entry` would have to handle both forms, or whether internally the string shorthand form should get converted to the object form with just one `index` key.

Thoughts?

eliperelman:
Fixes #367. Still a WIP, and still need docs.

### DIFF
--- a/packages/dev-server/index.js
+++ b/packages/dev-server/index.js
@@ -19,11 +19,13 @@ const getPublic = (neutrino, options) => {
 
 
 module.exports = (neutrino, opts = {}) => {
+  const port = opts.port || 5000;
   const publicHost = getPublic(neutrino, opts);
   const host = getHost(publicHost);
+
   const options = merge.all([
     {
-      port: 5000,
+      port,
       https: false,
       contentBase: neutrino.options.source,
       open: false,
@@ -31,7 +33,7 @@ module.exports = (neutrino, opts = {}) => {
       historyApiFallback: true,
       publicPath: '/',
       headers: {
-        host: publicHost
+        host: `${publicHost}:${port}`
       },
       stats: {
         assets: false,
@@ -49,7 +51,7 @@ module.exports = (neutrino, opts = {}) => {
       }
     },
     opts,
-    { host, public: publicHost },
+    { host, public: `${publicHost}:${port}` },
     neutrino.options.port ? { port: neutrino.options.port } : {},
     neutrino.options.https ? { https: neutrino.options.https } : {}
   ]);
@@ -59,9 +61,7 @@ module.exports = (neutrino, opts = {}) => {
   neutrino.config
     .devServer
       .merge(options)
-      .end()
-    .when(options.open, () => neutrino.on('start', () => open(url, { wait: false })))
-    .entry('index')
-      .when(options.hot, entry => entry.prepend(require.resolve('webpack/hot/dev-server')))
-      .prepend(`${require.resolve('webpack-dev-server/client')}?${url}`);
+      .when(options.open, () => {
+        neutrino.on('start', () => open(url, { wait: false }));
+      });
 };
--- a/packages/neutrino/bin/start.js
+++ b/packages/neutrino/bin/start.js
@@ -1,4 +1,5 @@
 const ora = require('ora');
+const debounce = require('lodash.debounce');
 const { start } = require('../src');
 const base = require('./base');
 
@@ -42,7 +43,7 @@ module.exports = (middleware, args) => {
         });
       } else {
         const { devServer } = compiler.options;
-        const url = `${devServer.https ? 'https' : 'http'}://${devServer.public}:${devServer.port}`;
+        const url = `${devServer.https ? 'https' : 'http'}://${devServer.public}`;
         const building = ora({
           text: 'Waiting for initial build to finish',
           enabled: global.interactive
@@ -53,10 +54,10 @@ module.exports = (middleware, args) => {
         compiler.plugin('done', () => {
           building.succeed('Build completed');
         });
-        compiler.plugin('compile', () => {
+        compiler.plugin('compile', debounce(() => {
           building.text = 'Source changed, re-compiling';
           building.start();
-        });
+        }, 1000, { leading: true, maxWait: 1000 }));
       }
     }
   });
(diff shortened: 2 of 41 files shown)


packages/dev-server/index.js
```diff
@@ -19,19 +19,21 @@ const getPublic = (neutrino, options) => {
...
+      port,
       https: false,
       contentBase: neutrino.options.source,
       open: false,
       hot: true,
       historyApiFallback: true,
       publicPath: '/',
       headers: {
-        host: publicHost
+        host: `${publicHost}:${port}`
```
helfi92: Is there a reason why we include it in the host if it's defined in `port`?

eliperelman: This: https://github.com/mozilla-neutrino/neutrino-dev/pull/487/files#diff-e157c45007fb7ba6df1afdf15e78279eR46

I wouldn't work everywhere without doing it like this. Or maybe it should only be set for public?

helfi92: Maybe just `public`.

eliperelman: I'll try it out and see what happens.

eliperelman: My bad, I thought this was editing `devServer.host`, but this is actually the `HOST` header we are modifying. The host header needs the port as well, not just the domain/host portion.

eliperelman: @edmorley could you give this a sanity check?

edmorley:
Yeah definitely - I'll do so within the next 24 hours :-)

packages/preact/index.js
```diff
@@ -53,4 +54,12 @@ module.exports = (neutrino, opts = {}) => {
         .set('react-dom', 'preact-compat')
         .set('create-react-class', 'preact-compat/lib/create-react-class')
         .set('react-addons-css-transition-group', 'preact-css-transition-group');
+
+  neutrino.config.when(neutrino.config.module.rules.has('lint'), () => {
```
edmorley: Was this change related to the rest of the PR? I couldn't see anything explaining it in the commit message or review comments?

packages/react/index.js
```diff
@@ -49,18 +50,22 @@ module.exports = (neutrino, opts = {}) => {
 
   neutrino.config.when(neutrino.config.module.rules.has('lint'), () => {
     neutrino.use(loaderMerge('lint', 'eslint'), {
-      plugins: ['react']
+      plugins: ['react'],
+      rules: {
+        'react/react-in-jsx-scope': 'off'
```
edmorley: Same here

packages/neutrino/bin/start.js
```diff
@@ -53,10 +54,10 @@ module.exports = (middleware, args) => {
         compiler.plugin('done', () => {
           building.succeed('Build completed');
         });
-        compiler.plugin('compile', () => {
+        compiler.plugin('compile', debounce(() => {
```
edmorley: Could you add a comment explaining why this is needed? (I thought Ora did debouncing? https://github.com/sindresorhus/ora#interval)

packages/preact/index.js
edmorley: Was this change related to the rest of the PR? I couldn't see anything explaining it in the commit message or review comments? 

Also could you add a comment explaining why this is needed? :-)

edmorley (commented):
This is amazing, thank you!

The only other place that needed to be adjusted for multi-entrypoint that I could find was html-template, but that's been fixed since in:
https://github.com/mozilla-neutrino/neutrino-dev/pull/509/files#diff-f7eeb689dea27b680d5b7fd733b51ec9R167

packages/neutrino/bin/start.js
eliperelman: This isn't debouncing the animation, rather this is debouncing the `compile` event. If `devServer` gets several updates in a row, like an editor writes multiple files at different times, the compiler will send several `compile` events, which causes a lot of noise in the terminal. This tries to reduce that logging noise to just once per second.

eliperelman: I'll add comment. :)

packages/preact/index.js
eliperelman: This makes it so we can use JSX without having to manually say:

```js
import React from 'react';
```

This was supposed to be part of the create-project fixes, but I did that work in tandem with this one because they were interlinked, and I tried to split them out, but unfortunately missed a few things.

PR #488, opened 2017-11-23
Refactor create-project to support testing the cli and middleware integration

### LINKED ISSUE #472: Implement e2e testing
Neutrino is made up of many moving parts. It is paramount that these moving parts operate well with each other.

With the introduction of the new scaffolding tools, we believe there may be an opportunity to run high level overarching e2e tests that confirm common tools are able to start, build, lint, and test.

Note: I do not believe this is a replacement for adding more thorough package level tests. This is a quick win that gives us immediately more confidence and ensures presets well work well with each other.

eliperelman:
Fixes #472. The tests take a little while, but it runs pretty well!

We attempt to run each preset, along with tests, and linting in the project and make sure they all pass for each project type. There's more that can be done here as well, such as making sure the package.json and .neutrinorc.js have the right contents.


packages/create-project/commands/init/index.js
```diff
@@ -112,61 +113,90 @@ module.exports = class Project extends Generator {
...
+          stdio: this.options.stdio,
+          env: process.env,
+          cwd: this.options.directory
+        });
+      }
     }
 
     if (this.data.linter) {
-      this.spawnCommandSync('neutrino', ['lint', '--fix'], { stdio: 'ignore' });
+      this.spawnCommandSync('neutrino', ['lint', '--fix'], { stdio: this.options.stdio, cwd: this.options.directory });
```
helfi92: Would this output to the user when linting fails to fix? I think we shouldn't output any linting errors when we're in the installing phase.

packages/create-project/commands/init/index.js
```diff
@@ -112,61 +113,90 @@ module.exports = class Project extends Generator {
...
     }
   }
 
   end() {
-    this.log(`\n${chalk.green('Success!')} Created ${chalk.cyan(this.options.directory)} at ${chalk.yellow(process.cwd())}`);
-    this.log(`To get started, change your current working directory to: ${chalk.cyan(this.options.directory)}`);
+    this.log(`\n${chalk.green('Hooray, I successfully created your project!')}`);
+    this.log(`To get started, change your current working directory to:\n  ${chalk.cyan(this.options.directory)}`);
```
helfi92: "Hooray" :) 

If the user is installing an app, what do you think if we log something like:

"To start your application, run `yarn start`" (or `npm start` if `isYarn` evaluates to false)

packages/create-project/commands/init/questions.js
```diff
@@ -6,18 +6,15 @@ module.exports = () => [
     choices: [
       {
         name: 'A web or Node.js application',
-        value: 'application',
-        checked: true
```
helfi92: Would omitting the `checked` property default to having the first option checked? Not having at least one checked property would maybe require the user to press the keyboard down arrow before seeing one of the options selected.

packages/create-project/commands/init/utils.js
```diff
@@ -1,25 +1,30 @@
 const projects = {
   '@neutrinojs/airbnb': {
+    type: 'linting',
```
helfi92: What do you think if we store the types in a constant.

```
const TYPES = {
  LINTING: 'linting',
  PROJECT: 'project',
  TESTING: 'testing'
}
```

helfi92 (approved):
Lovely! See comments regarding a couple of concerns. Otherwise, I'm really happy with the changes :)

packages/create-project/commands/init/index.js
eliperelman: This is normally set to ignore, but during testing we want to see the output so it fails the test.

packages/create-project/commands/init/index.js
eliperelman: I like it.

packages/create-project/commands/init/questions.js
eliperelman: Yes, this causes the first option to be selected. I think checked only works for checkable items, which these weren't.

packages/create-project/commands/init/utils.js
eliperelman: 👍

packages/create-project/commands/init/index.js
helfi92: Agreed. What you said makes a lot of sense :)

eliperelman:
Full output image:

terminal output

eliperelman:
(Using the `test` env makes it use links for local deps instead of npm/yarn lookups.)

helfi92:
Looks great! I assume the directory name you chose when taking the screenshot was `db408292ac69f703`. Just making sure 😅

eliperelman:
Haha yeah, I use a random directory every time. I use an alias to generate the directory:

```bash
alias random="node -e \"process.stdout.write(require('crypto').randomBytes(8).toString('hex'))"\"
```

helfi92:
I love it! Thanks a lot for sharing 🙏 Added it to my list of aliases!

eliperelman:
Revised PR again, and will also need to update when @helfi92's extensions PR lands. I've done some work to the Vue preset to make it pass the tests, and have also made some changes to the tests themselves to get them to run better in parallel, making it faster.

edmorley:
Included in this PR were some additions of:
`const NEUTRINO_MODULES = join(__dirname, '../../node_modules');`

I'm not really sure what that path if trying to reference? It seems like it might not be taking into account hoisting properly. It might be worth adding some comments alongside them?

eliperelman:
When running create-project from the tests, packages are linked instead of installed, since we don't want non-existent versions installed in the repo, but rather to test the current local versions of the packages. But when linking the vue package, the vue-loader had a lot of trouble trying to find its deps because of hoisting combined with a different linked structure, and adding `NEUTRINO_MODULES` fixed that.

eliperelman:
Either way, adding more explanatory in comments would be a good thing, and I'll start doing that.

edmorley:
Ah ok. I'm concerned that it will now mean projects using Neutrino have `node_modules` outside their project directory though? (Given hoisting will mean the directory is close to the project root already, even before the `../../`)

eliperelman:
@edmorley oooooooo great point. 🤔 

Maybe we should instead this only when working in the monorepo, maybe with an environment variable or Neutrino option?

jaridmargolin:
Is this specific to working with lerna or monorepos in general? I wonder if you can just sniff for `lerna.json`.

eliperelman:
@jaridmargolin you make an **excellent point**, and one which highlights a greater need for #516. If the package.json gets injected into options and is accessible by middleware, it then becomes super easy to determine if this is the monorepo, because we can just look at the name!

```js
.add(MODULES)
.when(neutrino.options.packageJson.name === 'neutrino-dev', (modules) => {
  modules.add(NEUTRINO_MODULES);
})
```

Issue #383, opened 2017-10-27
See if any of the @neutrinojs/eslint defaults are unnecessary


edmorley:
neutrino-middleware-eslint sets a number of ESLint defaults:
https://github.com/mozilla-neutrino/neutrino-dev/blob/6ea0728c95ac1a7c061428a8dd7a33db7dfd6192/packages/neutrino-middleware-eslint/index.js#L59-L84

As seen in #380 and then discussed in #382, the more defaults there are, the more things people have to override/keep in sync, particularly for confusing rules like [strict](https://eslint.org/docs/rules/strict#rule-details), whose behaviour is affected by several other settings.

Some settings we clearly do need (eg `sourceType: 'module'`), however I wonder if setting that makes eg `impliedStrict: true` redundant?

I've had a look at the ESLint docs, but it still wasn't that clear - so I suspect we'll need to do some digging in the ESLint source (followed by some tickets asking for clarification in their docs).

eirikurn:
IMO emitWarning and emitError should have default values. Both force all eslint issues as warnings and errors respectively. When both are set, emitError takes precedence, causing eslint warnings to be reported as errors and breaking the build.

The default logic is better (emitError: false, emitWarning: false). Then eslint issues are reported as warnings unless one of them is an error.

eliperelman:
Fixed emitting, evaluated settings in #493.

edmorley:
Reopening, since I think at least `impliedStrict` is unnecessary.

eliperelman:
> `impliedStrict` - enable global strict mode (if `ecmaVersion` is 5 or greater)

PR #493, opened 2017-11-28
Use eslint-loader default behavior for emitting warnings and errors

eliperelman:
Addresses #383.

Need a strong look at the removal of `emitWarning` and `emitError` properties to see if there are any implications I may have missed.

cc: @eirikurn

### DIFF
--- a/packages/eslint/index.js
+++ b/packages/eslint/index.js
@@ -31,7 +31,6 @@ const getEslintRcConfig = pipe(
 );
 
 module.exports = (neutrino, opts = {}) => {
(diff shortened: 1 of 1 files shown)


eliperelman:
For reference from eslint-loader:

https://github.com/MoOx/eslint-loader#errors-and-warning

eliperelman:
@edmorley I took a look at the properties we were setting and honestly couldn't find any values were I thought they should be excluded to their default values

eirikurn:
LGTM.

PR #495, opened 2017-11-29
Enable CSS Modules by default in *.module.css files

### LINKED ISSUE #440: Enable css modules for local project files
For the style-loader, I would like to see a default added for enabling CSS modules. I could see two potential solutions:

- Enable CSS modules for files ending in `.module.css`
- Enable CSS modules for all project CSS files

My gut tells me the first option may be the better one for keeping as a default.

eliperelman:
Fixes #440.

@timkelty would you be interested in having a look at this?

### DIFF
--- a/packages/style-loader/index.js
+++ b/packages/style-loader/index.js
@@ -29,7 +29,10 @@ module.exports = (neutrino, opts = {}) => {
       styleUseId: `${options.styleUseId}${options.modulesSuffix}`,
       cssUseId: `${options.cssUseId}${options.modulesSuffix}`,
       hotUseId: `${options.hotUseId}${options.modulesSuffix}`,
-      extractId: `${options.extractId}${options.modulesSuffix}`
+      extractId: `${options.extractId}${options.modulesSuffix}`,
+      css: {
+        modules: options.modules
+      }
     }));
   }
 
--- a/packages/style-loader/index.js
+++ b/packages/style-loader/index.js
@@ -3,8 +3,16 @@ const merge = require('deepmerge');
 
 module.exports = (neutrino, opts = {}) => {
   const options = merge({
+    test: neutrino.regexFromExtensions(['css']),
+    ruleId: 'style',
     styleUseId: 'style',
+    cssUseId: 'css',
     hot: true,
+    hotUseId: 'hot',
+    modules: true,
+    modulesSuffix: '-modules',
+    modulesTest: /\.module.css$/,
+    extractId: 'extract',
     extract: {
       plugin: {
         filename: neutrino.options.command === 'build' ? '[name].[contenthash].css' : '[name].css'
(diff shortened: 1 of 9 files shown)


timkelty:
@eliperelman not sure what I'm missing…but can't get your PR working:

As soon as I add a `.module.css` import, I pukes:

>>>
$ neutrino build
✖ Building project failed
../neutrino-dev/node_modules/css-loader!../neutrino-dev/node_modules/style-loader!../neutrino-dev/node_modules/css-loader!./src/test.module.css
Module build failed: Unknown word (5:1)

  3 | // load the styles
  4 | var content = require("!!../../neutrino-dev/node_modules/css-loader/index.js!./test.module.css");
> 5 | if(typeof content === 'string') content = [[module.id, content, '']];
    | ^
  6 | // Prepare cssTransformation
  7 | var transform;
  8 |

timkelty (commented):
Shouldn't we be setting `options.css.modules` = `options.modules` to actually enable modules in css-loader? https://github.com/webpack-contrib/css-loader#options

eliperelman:
lol yup. I'll revise.

eliperelman:
@timkelty OK, should be fixed.

packages/style-loader/index.js
```diff
@@ -3,63 +3,92 @@ const merge = require('deepmerge');
 
 module.exports = (neutrino, opts = {}) => {
   const options = merge({
+    test: opts.modules !== false
+      ? /(?<!\.module)\.css$/
```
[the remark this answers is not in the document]
eliperelman: omg 🤣 

OK, lemme get that fixed.

eliperelman:
@timkelty OK, give that a shot.

packages/style-loader/index.js
```diff
@@ -2,64 +2,103 @@ const ExtractTextPlugin = require('extract-text-webpack-plugin');
...
 module.exports = (neutrino, opts = {}) => {
+  const cssTest = neutrino.regexFromExtensions(['css']);
+  const cssModulesTest = neutrino.regexFromExtensions(['module.css']);
+
   const options = merge({
+    test: (input) => {
+      const isCssModule = cssModulesTest.test(input);
+      const isRegularCss = cssTest.test(input);
+
+      if (opts.modules !== false && isCssModule) {
```
timkelty: Since `opt.modules` is always just a bool, why not just do `opts.modules && isCssModule`?

eliperelman: Hmm, for some reason I had it in my head that it could be an object, but I guess that's not the case.

eliperelman: Ah, I know why: `opts.modules` could be undefined, and in that case, it should be treated the same as `true`, since that would be the default value.

timkelty: Ah gotcha.

packages/style-loader/index.js
```diff
@@ -2,64 +2,103 @@ const ExtractTextPlugin = require('extract-text-webpack-plugin');
...
 
-    options.extract.loader = Object.assign({
-      use: useKeys.map(key => ({
-        loader: styleEntries[key].get('loader'),
-        options: styleEntries[key].get('options')
-      })),
-      fallback: {
-        loader: styleEntries[options.styleUseId].get('loader'),
-        options: styleEntries[options.styleUseId].get('options')
+  if (options.modules) {
```
timkelty: Something to consider here might be to set `options.extract.plugin.ignoreOrder: options.modules`, just as a nice default: https://github.com/webpack-contrib/extract-text-webpack-plugin#options

timkelty (approved):
LGTM!

See note about potentially setting ignoreOrder. Could go either way on it - just an idea.

packages/style-loader/index.js
eliperelman: Ooo nice!

eliperelman:
Thanks @timkelty for the great review!

packages/style-loader/index.js
timkelty: Hrmm - `ignoreOrder` is an option for `options.extract.plugin`, not `options.css`.

eliperelman: Yikes, sorry. :(

timkelty: I think we should be able to just do it up in the initial `options` merge, eg
```
    extract: {
      plugin: {
        filename: neutrino.options.command === 'build' ? '[name].[contenthash].css' : '[name].css',
        ignoreOrder: opts.modules !== false
      }
    }
```

PR #497, opened 2017-11-29
Use Neutrino's extensions in the rest of packages

helfi92:


eliperelman:
I grepped the following usages of "extensions":

```
packages/eslint/index.js:            extensions: ['js', 'jsx'],
packages/jest/src/index.js:    ...neutrino.config.resolve.extensions.values().map(e => e.replace('.', ''))
```

helfi92:
Webstorm didn't do a good job at grepping. @eliperelman I started using [ag](https://github.com/ggreer/the_silver_searcher) for searching files. Amazing tool. It ignores everything listed in `.gitignore` which is pretty neat.

eliperelman:
:shipit:

packages/eslint/README.md
```diff
@@ -46,7 +46,7 @@ neutrino.use(eslint, {
     cwd: neutrino.options.root,
     useEslintrc: false,
     root: true,
-    extensions: ['js', 'jsx'],
+    extensions: neutrino.options.extensions,
```
edmorley: This means eslint will now be run against `.json` files too. Is this intended? (For example if people have the "enforce trailing commas" eslint rule enabled, eslint will complain unless the .json is changed to be invalid)

eliperelman: Have we tested that this really happens? That would be weird! It would also mean single quote properties and values would also break...I guess we should find out. If it does, we can filter `json` from the list. Good catch.

PR #500, opened 2017-11-30
Remove non-functioning script-ext plugin

### LINKED ISSUE #470: Problems with the use of script-ext-html-webpack-plugin
The Neutrino web preset uses [script-ext-html-webpack-plugin](https://github.com/numical/script-ext-html-webpack-plugin) however at first glance I think it doesn't work with `html-webpack-template` (since that requires disabling injection mode, see jaketrent/html-webpack-template#61).

In addition, the docs for `script-ext-html-webpack-plugin` say:

> The order is important - the plugin must come after HtmlWebpackPlugin.

However the web preset doesn't use `.before()`, which is important if people end up customising the `html` plugin.

Also, do we really need to use `script-ext-html-webpack-plugin` at all?

eliperelman:
Fixes #470.

### DIFF
--- a/packages/web/index.js
+++ b/packages/web/index.js
@@ -15,7 +15,6 @@ const devServer = require('@neutrinojs/dev-server');
 const { join, basename } = require('path');
 const { resolve } = require('url');
 const merge = require('deepmerge');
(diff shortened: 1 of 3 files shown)


edmorley:
Looks good, many thanks!

PR #504, opened 2017-12-01
Use allChunks CSS options when doing extraction

### LINKED ISSUE #503: style-loader: allChunks: true for ExtractTextPlugin?
I'm wondering if we should be setting `allChunks: true` for `ExtractTextPlugin` in `style-loader`: https://github.com/webpack-contrib/extract-text-webpack-plugin#options

Does anybody have experience with this option? Reading the description, I'm still not 💯  I grasp if and when to use it, but it does mention combined use with `CommonsChunkPlugin`:

> Extract from all additional chunks too (by default it extracts only from the initial chunk(s))
> When using CommonsChunkPlugin and there are extracted chunks (from ExtractTextPlugin.extract) in the commons chunk, allChunks must be set to true

Any ideas @mozilla-neutrino/core-contributors?

eliperelman:
Fixes #503.

### DIFF
--- a/packages/web/index.js
+++ b/packages/web/index.js
@@ -23,7 +23,6 @@ const NEUTRINO_MODULES = join(__dirname, '../../node_modules');
 
 module.exports = (neutrino, opts = {}) => {
   const publicPath = './';
-  const extract = opts.extract || process.env.NODE_ENV === 'production';
   const options = merge({
     publicPath,
     env: [],
@@ -39,8 +38,7 @@ module.exports = (neutrino, opts = {}) => {
     },
     style: {
       hot: opts.hot !== false,
-      extract: extract && {},
-      css: extract && { allChunks: true }
+      extract: opts.extract || (process.env.NODE_ENV === 'production' && {})
     },
     manifest: opts.html === false ? {} : false,
     minify: {},
--- a/packages/web/index.js
+++ b/packages/web/index.js
@@ -23,6 +23,7 @@ const NEUTRINO_MODULES = join(__dirname, '../../node_modules');
 
 module.exports = (neutrino, opts = {}) => {
   const publicPath = './';
+  const extract = opts.extract || process.env.NODE_ENV === 'production';
   const options = merge({
     publicPath,
     env: [],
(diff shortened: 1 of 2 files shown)


eliperelman:
@timkelty I moved `ignoreOrder` in here.

packages/web/index.js
```diff
@@ -23,6 +23,7 @@ const NEUTRINO_MODULES = join(__dirname, '../../node_modules');
 
 module.exports = (neutrino, opts = {}) => {
   const publicPath = './';
+  const extract = opts.extract || process.env.NODE_ENV === 'production';
```
timkelty: I'm not sure I like `extract` being an option for the web middleware, as it is completely specific to the style-loader. I think I'd rather enable/disable this via `opts.style.extract`.

Also, why are we only extracting in production? That seems off to me.
Either way, the way we have style-loader configured, HMR works, and the files aren't actually output.

If anything, I would the test should be `neutrino.options.command === 'build'`, but I kinda think we shouldn't do it conditionally. What do you think?

packages/web/index.js
```diff
@@ -38,7 +39,8 @@ module.exports = (neutrino, opts = {}) => {
     },
     style: {
       hot: opts.hot !== false,
-      extract: opts.extract || process.env.NODE_ENV === 'production' && {}
+      extract: extract && {},
+      css: extract && { allChunks: true }
```
timkelty: Like `ignoreOrder`, `allChunks` is an option for `opts.style.extract.plugin`, not `opts.style.css`.

eliperelman: 🤦‍♂️

packages/web/index.js
eliperelman: > I'm not sure I like `extract` being an option for the web middleware, as it is completely specific to the style-loader. I think I'd rather enable/disable this via `opts.style.extract`.

Makes sense to me.

> Also, why are we only extracting in production?

It's unnecessary overhead during development. We can produce a build faster if we don't have to do extraction, and save extraction for when it really matters: when creating a production build.

> If anything, I would the test should be `neutrino.options.command === 'build'`, but I kinda think we shouldn't do it conditionally. What do you think?

I think it should be based on same condition of minification and chunking, which occur when in production mode, not for the build command. Right now production mode is synonymous with optimized mode, which I think extraction should fall under.

timkelty (approved):
LGTM!

eliperelman:
Thank you @timkelty!

PR #509, opened 2017-12-02
Allow usage of multiple same plugins, map mains to html plugins

### LINKED ISSUE #326: Provide way to use neutrino-middleware-html-template multiple times
For [Treeherder](https://github.com/mozilla/treeherder), we have a multi-page app that needs to use different custom templates for each page, rather than the stock template generated by html-webpack-plugin / html-webpack-template.

In our Neutrino v4 config there's a lot of unnecessary duplication:
https://github.com/mozilla/treeherder/blob/5c0c7eb5e241638bc1d636cce2468d568e374ce2/neutrino-custom/base.js#L96-L145

At the same time as upgrading to Neutrino v6, this has been reduced down to roughly:
```js
    (neutrino) => {
      // Remove the default entry point and html plugin
      neutrino.config.entry('index').delete(neutrino.options.entry);
      neutrino.config.plugins.delete('html');

      const entries = ['index', 'perf', 'logviewer', 'failureviewer', 'userguide'];

      entries.forEach(entry => neutrino.config
        .entry(entry)
          .add(join(neutrino.options.source, `entry-${entry}.js`))
          .end()
        .plugin(`html-${entry}`)
          .use(HtmlPlugin, [{
              filename: `${entry}.html`,
              template: join(neutrino.options.source, `${entry}.html`),
              // The vendor/runtime chunk names here must match those in neutrino-middleware-chunk.
              chunks: [entry, 'vendor', 'runtime'],
              // Copied from neutrino-middleware-html-template
              minify: {
                useShortDoctype: true,
                keepClosingSlash: true,
                collapseWhitespace: true,
                preserveLineBreaks: true
              },
          }])
      );
    },
```

Whilst this is an improvement on how it was before, it still has to duplicate the minify options from [neutrino-middleware-html-template](https://github.com/mozilla-neutrino/neutrino-dev/blob/v6.2.0/packages/neutrino-middleware-html-template/index.js).

Ideally I'd be able to just `.use()` the middleware multiple times (passing the options I need to override for each), and be able to override the plugin entry name for each, to avoid the `html` name collision.

Is there a way to do this that I'm missing or else would adding support for customising the names be possible?

eliperelman:
Fixes #326.

This is a breaking change because it removes the `html` plugin from the web preset and replaces it with `html-index`.


edmorley:
Oh good spot!

packages/web/index.js
```diff
@@ -165,6 +164,12 @@ module.exports = (neutrino, opts = {}) => {
           Object
             .keys(neutrino.options.mains)
             .forEach(key => {
+              neutrino.use(htmlTemplate, merge({
+                pluginId: `html-${key}`,
+                filename: `${key}.html`,
+                chunks: [key, 'vendor', 'runtime']
```
edmorley: Could you add a comment here and to `packages/chunk/index.js` saying the two sets of chunk names need to be kept in sync?

packages/web/README.md
```diff
@@ -340,7 +339,7 @@ _Note: Some plugins are only available in certain environments. To override them
 | `env` | Inject environment variables into source code at `process.env`, defaults to only inject `NODE_ENV`. From `@neutrinojs/env`. | all |
 | `extract` | Extracts CSS from JS bundle into a separate stylesheet file. From `@neutrinojs/style-loader`. | all |
 | `extract-modules` | Extracts CSS from JS bundle into a separate stylesheet file. From `@neutrinojs/style-loader`. | all |
-| `html` | Automatically generates HTML files for configured entry points. From `@neutrinojs/html-template` | all |
+| `html-{MAIN_NAME}` | Automatically generates HTML files for configured entry points. `{MAIN_NAME}` corresponds to the entry point of each page. By default, there is only a single `index` main, so this would generate a plugin named `html-index`. From `@neutrinojs/html-template` | all |
```
edmorley: Good idea on making this breaking change, I think it will result in much less confusion long term for different naming forms for single and multi `mains` projects.

packages/web/index.js
eliperelman: Sure thing!

timkelty:
@eliperelman something's not right here…in `master`'s current state, `neutrino build` is not creating any html files.

Looking at the changes here, it indeed appears that `htmlTemplate` is only getting added in `start`: https://github.com/eliperelman/neutrino-dev/blob/420bdac667b03798f358f49c54479ab69d6b1039/packages/web/index.js#L156-L171

eliperelman:
@timkelty yikes, great catch.

PR #511, opened 2017-12-02
pass clean options to webpack plugin

### LINKED ISSUE #502: clean middleware options are not passed to the webpack plugin
What version of Neutrino are you using?
7.3.3

Are you trying to use any presets? If so, which ones, and what versions?
neutrino-preset-react (but I believe the problem lies in the neutrino-preset-web)

Are you using the Yarn client or the npm client? What version?
npm 5.5.1

What version of Node.js are you using?
8.3.0

What operating system are you using?
Kubuntu

I was struggling a bit to tweak the "clean" setup (in the neutrino-preset-react) when I discovered that my options were not passed to the clean-webpack-plugin has I would have expected.

if my .neutrinorc.js I have something like:
```js
[
      'neutrino-preset-react',
      {
        html: {
          // ...
        },

        babel: {
          //...
        },

        clean: {
          exclude: ['.git']
        }
      }
    ],
```

After digging a bit I found the possible issue here:
https://github.com/mozilla-neutrino/neutrino-dev/blob/6c7e840fb79cd612a64cae66c27acd132fc3e5f9/packages/web/index.js#L189

I could actually get around my problem by using the following script (don't know if it is really elegant though)
```js
(neutrino) => {
      const { config } = neutrino;

      config.plugins.delete('clean');
      config.plugin('clean')
        .use(CleanPlugin, [[neutrino.options.output], {
          exclude: ['.git']
        }]);
// ...
```

I've tried to fix this and send a PR but I now have issues about packages not being found (probably related to the '@neutrino/...' (i tried to "npm link" the package in order to try the changes)

zeropaper:
I hope this is the right way to fix #502 :)

### DIFF
--- a/packages/neutrino-preset-web/index.js
+++ b/packages/neutrino-preset-web/index.js
@@ -31,7 +31,7 @@ module.exports = (neutrino, opts = {}) => {
       async: true
     },
     babel: {},
-    clean: {
+    clean: opts.clean !== false && {
       paths: [neutrino.options.output]
     }
   }, opts);
@@ -55,11 +55,6 @@ module.exports = (neutrino, opts = {}) => {
       ]
     }, options.babel)
   });
-
-  Object.assign(options.clean, opts.clean || {});
-  if (!options.clean.paths.includes(neutrino.options.output)) {
-    options.clean.paths.push(neutrino.options.output);
-  }
 
   const staticDir = join(neutrino.options.source, 'static');
   const presetEnvOptions = options.babel.presets[0][1];
@@ -151,7 +146,12 @@ module.exports = (neutrino, opts = {}) => {
         .use(optimize.ModuleConcatenationPlugin);
     })
     .when(neutrino.options.command === 'build', (config) => {
-      neutrino.use(clean, options.clean);
+      if (options.clean) {
+        if (!options.clean.paths.includes(neutrino.options.output)) {
+          options.clean.paths.push(neutrino.options.output);
+        }
+        neutrino.use(clean, options.clean);
+      }
       neutrino.use(copy, {
         patterns: [{
           context: staticDir,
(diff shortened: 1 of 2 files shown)


packages/neutrino-preset-web/index.js
```diff
@@ -53,6 +56,11 @@ module.exports = (neutrino, opts = {}) => {
     }, options.babel)
   });
 
+  Object.assign(options.clean, opts.clean || {});
```
eliperelman: I don't think you need these lines since you are already setting defaults on line 34. You should be able to safely delete L59-63.

packages/neutrino-preset-web/index.js
```diff
@@ -30,7 +30,10 @@ module.exports = (neutrino, opts = {}) => {
     polyfills: {
       async: true
     },
-    babel: {}
+    babel: {},
+    clean: {
```
eliperelman: It would also be nice to disable cleaning by passing `false` to `options.clean`. You should be able to achieve this by changing this to:

```js
clean: opts.clean !== false && {
  paths: [neutrino.options.output]
}
```

packages/neutrino-preset-web/index.js
```diff
@@ -143,7 +146,12 @@ module.exports = (neutrino, opts = {}) => {
         .use(optimize.ModuleConcatenationPlugin);
     })
     .when(neutrino.options.command === 'build', (config) => {
-      neutrino.use(clean, { paths: [neutrino.options.output] });
+      if (options.clean) {
```
eliperelman: For consistency with the rest of the middleware, could you switch this to `.when`? Also, I don't think we need to include the `output` condition here since you have that set above in the options with `paths: [neutrino.options.clean]`:

```js
config.when(options.clean, () => neutrino.use(clean, options.clean));
```

zeropaper:
I discovered that the clean middleware was also not passing all options, so I corrected it on the fly.

Issue #357, opened 2017-10-12
Provide option to render CLI using webpack-dashboard


eliperelman:
I think it would be nice to provide the webpack-dashboard as a visualization option using something like:

```bash
neutrino start --dashboard
```

https://github.com/FormidableLabs/webpack-dashboard

eliperelman:
Another interesting project:

https://github.com/webpackmonitor/webpackmonitor

edmorley:
> https://github.com/webpackmonitor/webpackmonitor

I tried this one since it looked superior from the screenshots, however with it the builds for Treeherder would hang or OOM and not complete. I've not had a chance to write a reduced testcase for them (and realistically might not in the foreseeable future).

eliperelman:
The repo for https://github.com/FormidableLabs/electron-webpack-dashboard looked pretty good as well, with the caveat that it required you to install an Electron app prior to being able to run. If this could spin up a browser window instead that would be nice, or even if it installed itself when installing deps, that would be nice.

@kenwheeler would you happen to know if it's possible to consume the electron-webpack-dashboard in this manner, by either spinning up a browser window, or installing the app upon package install? Thanks!

collinklippel:
Hey guys! Do you know if this is still open to work on? Thanks!

eliperelman:
@collinklippel Sure is, thank you! I'll try and post my thoughts on how this would probably work:

First, you'll need a CLI option to decide to show the dashboard, for now let's pretend it's `--dashboard`. All of Neutrino's CLI interface is defined in https://github.com/mozilla-neutrino/neutrino-dev/tree/master/packages/neutrino/bin, and I figure this flag is only useful when using the `start` command. To add a new option for the start command, you need to modify this yargs command:

https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/neutrino/bin/neutrino.js#L61

Once you have the flag for the command, you'll need to modify the command to actually use it. If you are implementing webpack-dashboard, this means you'll need to inject some custom middleware to add it to plugins. The code for the start command is here:

https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/neutrino/bin/start.js

Notice that the start command receives two arguments: `middleware` and `args`. The middleware is an array, so you can either push a new function on the end, or tack one onto the beginning.

```js
middleware.push((neutrino) => {
  neutrino.config.plugins // ...
});
```

[...]

PR #519, opened 2017-12-03
Adding webpack-dashboard support

gerherz:
Adding webpack-dashboard support as stated in #357


docs/cli/README.md
```diff
@@ -150,6 +150,12 @@ Using the command `neutrino start` builds a project in development mode, also st
 watcher depending on the middleware or configuration used. This command sets the `NODE_ENV` environment variable to
 `development` by default.
 
+This command also receives a flag `--dashboard` to add the [webpack-dashboard](https://github.com/FormidableLabs/webpack-dashboard) plugin. To use the plugin, the start script in package.json can be modified:
+
+```json
+"start": "webpack-dashboard -- neutrino start --dashboard"
```
eliperelman: Is the `webpack-dashboard` command really needed here? Is there a way we can make this work with **only** `neutrino start --dashboard`?

edmorley:
Any thoughts about the concerns in:
https://github.com/mozilla-neutrino/neutrino-dev/issues/357#issuecomment-339033267

I'm not sure how reliable this package is.

eliperelman:
@edmorley your comment is referencing `webpackmonitor`, and this is for `webpack-dashboard`, right?

edmorley:
Oh indeed! haha. Sorry ignore me! :-)

docs/cli/README.md
gerherz: I couldn't find a way to make it work that way.

> In the latest version, you can either run your app, and run webpack-dashboard independently (by installing with npm install webpack-dashboard -g) or run webpack-dashboard from your package.json`

eliperelman: Apologies, this issue used to be simpler. The only thing I can think of then to make this a reality would be to `fork` the process by calling `webpack-dashboard` with the starting script. If you're up for it, we can try experimenting with it, but it will be much more complex.

gerherz: I'm working on it. I've spawned a new child process, but I need to silence the main process' stdout. I mean, messages like these:

>✔ Development server running on: http://localhost:5000
>✔ Build completed

Is there a way to silence the output and let webpack-dashboard take over stdout?

eliperelman: The CLI supports a `quiet` flag. Maybe you can set `args.quiet` to `true` when `args.dashboard` is `true`?

gerherz: Done

packages/neutrino/bin/start.js
```diff
@@ -2,16 +2,33 @@ const ora = require('ora');
 const debounce = require('lodash.debounce');
 const { start } = require('../src');
 const base = require('./base');
+const DashboardPlugin = require("webpack-dashboard/plugin");
```
eliperelman: Let's move this `require` to just under the `debounce` require.

packages/neutrino/bin/start.js
```diff
@@ -2,16 +2,33 @@ const ora = require('ora');
...
+const { spawn } = require('child_process');
```
eliperelman: Let's move this to be under `DashboardPlugin` once you move that.

packages/neutrino/bin/start.js
```diff
@@ -2,16 +2,33 @@ const ora = require('ora');
...
 
 module.exports = (middleware, args) => {
   const spinner = ora({ text: 'Building project' });
+  const argsClone = Object.assign({}, args);
```
eliperelman: Do we need this to be cloned rather than just overriding the value?

```js
if (args.dashboard) {
  args.quiet = true;
 // ...
}
```

packages/neutrino/bin/start.js
```diff
@@ -2,16 +2,33 @@ const ora = require('ora');
...
+
+  if (argsClone.dashboard) {
+    argsClone.quiet = true;
+
+    middleware.push(neutrino => {
+      neutrino.config.plugin('dashboard').use(DashboardPlugin);
+    });
+
+    const child = spawn(`${__dirname  }/../../../node_modules/.bin/webpack-dashboard`, [], { stdio: 'inherit' });
```
eliperelman: Since `webpack-dashboard` is a dependency of `neutrino` now, we can simplify this path to:

```js
const child = spawn(require.resolve('webpack-dashboard/bin/webpack-dashboard.js'), [], { stdio: 'inherit' });
```

This will let it work no matter how webpack-dashboard gets linked or hoisted.

packages/neutrino/bin/start.js
```diff
@@ -2,16 +2,33 @@ const ora = require('ora');
...
+
+    process.on('exit', () => {
+      child.kill();
+    })
```
eliperelman: Missing semicolon here.

packages/neutrino/bin/start.js
```diff
@@ -2,16 +2,33 @@ const ora = require('ora');
...
+  }
 
   return base({
     middleware,
-    args,
+    argsClone,
```
eliperelman: These are named args, so if we do decide we need a clone of `args`, we need to pass this as:

```js
args: argsClone,
```

packages/neutrino/bin/start.js
```diff
@@ -2,16 +2,33 @@ const ora = require('ora');
...
     NODE_ENV: 'development',
     commandHandler(config, neutrino) {
-      if (!args.start) {
+      if (!argsClone.start && !argsClone.dashboard) {
```
eliperelman: Weird, I think there was a bug here, because that **should** say `args.quiet` instead of `args.start`! Can you change that to:

```js
if (!argsClone.quiet && !argsClone.dashboard) {
```

packages/neutrino/bin/start.js
gerherz: The linter complains when reassigning a value of an argument.

packages/neutrino/bin/start.js
```diff
@@ -1,17 +1,34 @@
 const ora = require('ora');
 const debounce = require('lodash.debounce');
+const DashboardPlugin = require('webpack-dashboard/plugin');
+const { spawn } = require('child_process');
 const { start } = require('../src');
 const base = require('./base');
 
 module.exports = (middleware, args) => {
   const spinner = ora({ text: 'Building project' });
+  const argsClone = Object.assign({}, args);
```
eliperelman: Yeah, I think you can avoid the `argsClone` altogether here with either:

```js
if (args.dashboard) {
  Object.assign(args, { quiet: true });
}
```

or:

```js
if (args.dashboard) {
  args.quiet = true; // eslint-disable-line no-param-reassign
};
```

gerherz: Sure, I agree.

eliperelman:
Since Neutrino will no longer have a CLI, this idea has been obsoleted. Thank you for your effort here, it was appreciated.

Issue #518, opened 2017-12-03
neutrino-preset-jest says "no tests found"


jefo:
I have two test files, but preset says "No tests found". In same time karma preset works well if i replace jest in `--use`.
Windows 10, Node 8.9.0
```
PS C:\Users\j4\Projects\rxsupp.admin> yarn test
yarn run v1.2.1
$ neutrino test --use neutrino-preset-react neutrino-preset-jest
No tests found
In C:\Users\j4\Projects\rxsupp.admin
  21 files checked.
  testMatch:  - 21 matches
  testPathIgnorePatterns: \\node_modules\\ - 21 matches
  testRegex: test\.*(_test|_spec|\.test|\.spec)\.jsx?$ - 0 matches
Pattern:  - 0 matches
Done in 6.76s.
PS C:\Users\j4\Projects\rxsupp.admin> ls test


    Каталог: C:\Users\j4\Projects\rxsupp.admin\test


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----       03.12.2017     18:30            115 users-component_test.jsx
-a----       03.12.2017     18:27           9185 users_test.js

```

eliperelman:
Sorry about your issue. I don't currently have a Windows machine, but I will see if I can get someone with Windows to replicate tomorrow. Thanks!

sgillespie:
The issue on windows is that `path.join` uses a backslash as the separator, and the backslash escapes the following `.`. Therefore the regex won't match the path correctly. Here's the bit:

```
const testRegex = join(
    basename(tests),
    `.*(_test|_spec|\\.test|\\.spec)\\.(${extensions.join('|')})$`
);
```
One possible solution would be to join the two strings with `/` as the separator. I've tested this by overriding the `testRegex` and fixed the issue.

constgen:
Guys, I have created this PR more then a week ago https://github.com/mozilla-neutrino/neutrino-dev/pull/679. Please merge

constgen:
Is somebody going to merge fixes? My team have to modify `node_modules` manually to make it work. It is insane.

sgillespie:
@constgen: You don't actually have to modify `node_modules`, just override `testRegex` in `neutrinorc.js`

```
    ['@neutrinojs/jest', {
      coverageReporters: ['lcov', 'text'],
      coveragePathIgnorePatterns: [
        '/node_modules/',
        '<rootDir>/src/index.js',
      ],
[...]

PR #526, opened 2017-12-04
fix Jest testRegex param to match neutrino test dir

jefo:
Quickfix for #518.

### DIFF
--- a/packages/jest/src/index.js
+++ b/packages/jest/src/index.js
@@ -52,7 +52,7 @@ function normalizeJestOptions(opts, neutrino, usingBabel) {
       bail: true,
       coveragePathIgnorePatterns: [neutrino.options.node_modules],
       collectCoverageFrom: [join(basename(neutrino.options.source), '**/*.js')],
-      testRegex: join(basename(neutrino.options.tests), '.*(_test|_spec|\\.test|\\.spec)\\.jsx?$'),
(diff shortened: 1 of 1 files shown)


packages/jest/src/index.js
```diff
@@ -52,7 +52,7 @@ function normalizeJestOptions(opts, neutrino, usingBabel) {
       bail: true,
       coveragePathIgnorePatterns: [neutrino.options.node_modules],
       collectCoverageFrom: [join(basename(neutrino.options.source), '**/*.js')],
```
eliperelman: @helfi92 looks like there are a few outstanding extensions here.

packages/jest/src/index.js
```diff
@@ -52,7 +52,7 @@ function normalizeJestOptions(opts, neutrino, usingBabel) {
...
-      testRegex: join(basename(neutrino.options.tests), '.*(_test|_spec|\\.test|\\.spec)\\.jsx?$'),
+      testRegex: join(basename(neutrino.options.tests), '.*(_test|_spec|\\.test|\\.spec)\\.jsx?$').replace('test\\', '(test\\)'),
```
eliperelman: @jefo if you log the value of `testRegex` before and after this change, what does it log?

eliperelman:
Adding @okonet as he may have some insight here. 😃

sgillespie:
I do not think this is necessary anymore, since #679 has been merged

edmorley:
> I do not think this is necessary anymore, since #679 has been merged

Ah ok. Thank you for the PR anyway, much appreciated :-)

PR #529, opened 2017-12-04
Restructure minify to aggregate *-minify middlewares

timkelty:
- Move imagemin to image-minify
- Move minify to babel-minify
- Make minify an aggregate


packages/babel-minify/README.md
```diff
@@ -0,0 +1,97 @@
+# Neutrino Minify Middleware
```
eliperelman: Need to update this title to "Neutrino Babel Minify Middleware".

packages/babel-minify/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+`@neutrinojs/minify` is Neutrino middleware for minifying source code using
```
eliperelman: `@neutrinojs/babel-minify`. I won't make any more comments for this, you can just do a find and replace for these.

packages/babel-minify/README.md
```diff
@@ -0,0 +1,97 @@
...
+`@neutrinojs/minify` creates some conventions to make overriding the configuration easier once you are ready to
+make changes.
+
+### Plugins
+
+The following is a list of plugins and their identifiers which can be overridden:
+
+| Name | Description | Environments and Commands |
+| --- | --- | --- |
+| `babel-minify` | Minifies source code using `BabelMinifyWebpackPlugin`. | all |
```
eliperelman: Ah ok, here's the breaking change. ++

packages/babel-minify/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Contributing
+
+This middleware is part of the [neutrino-dev](https://github.com/mozilla-neutrino/neutrino-dev) repository, a monorepo
+containing all resources for developing Neutrino and its core presets and middleware. Follow the
+[contributing guide](https://neutrino.js.org/contributing) for details.
+
+[npm-image]: https://img.shields.io/npm/v/@neutrinojs/minify.svg
```
eliperelman: These URLs will need to be updated as well.

packages/babel-minify/index.js
```diff
@@ -0,0 +1,5 @@
+const BabelMinify = require('babel-minify-webpack-plugin');
```
eliperelman: For consistency with other middleware, let's name this variable `BabelMinifyPlugin`.

packages/minify/index.js
```diff
@@ -1,5 +1,6 @@
-const BabelMinify = require('babel-minify-webpack-plugin');
+const babelMinify = require('@neutrinojs/babel-minify');
+const imageMinify = require('@neutrinojs/image-minify');
 
-module.exports = ({ config }, options = {}) => config
-  .plugin('minify')
-  .use(BabelMinify, [options.minify, options.overrides]);
+module.exports = (neutrino, options = {}) => neutrino
```
eliperelman: Let's make this a non-returning arrow function:

```js
module.exports = (neutrino, options = {}) => {
  neutrino // ...
};
```

timkelty:
This puts us back in the position of deciding wether to use `imagemin` in production (from `@neutrinojs/web`)…

Personally I'm ok with just ditching `minify` and just keeping them all separate. Curious to hear what others think.

timkelty:
More food for thought: should we just take `minify` out of `web` all together?

eliperelman:
> More food for thought: should we just take `minify` out of `web` all-together? Guessing not, since `web` is more of a ready-to-go preset…

Correct, the point of the high-level presets is to give users the best out-of-the-box experience we can give them.

> Also - does the `module-concat` stuff belong in `minify`, or should it remain in `web`?

I think if anything this would be its own middleware, since it's more related to scope hoisting than to minification.

eliperelman:
> dd93341 includes changes to allow passing false to exclude a minify preset. For now, it also passes false by default from web, as we weren't minifying images by default before

Perfect!

packages/minify/index.js
```diff
@@ -1,5 +1,12 @@
...
+const imageMinify = require('@neutrinojs/image-minify');
 
-module.exports = ({ config }, options = {}) => config
-  .plugin('minify')
-  .use(BabelMinify, [options.minify, options.overrides]);
+module.exports = (neutrino, options = {
+  babel: {},
+  image: {}
+}) => {
+
```
eliperelman: Remove this extra line.

timkelty: yep already did!

timkelty:
> dd93341 includes changes to allow passing false to exclude a minify preset. For now, it also passes false by default from web, as we weren't minifying images by default before

The only weird part is now `web` users will have to pass `{ minify: image: {} }`  to enable image minify. Think thats ok, or should we coerce `true` to `{}` in `minify`?

timkelty:
Sorry I got confused - thought I was PRing from timkelty.

Anyway - review changes should be made.

packages/minify/index.js
```diff
@@ -1,11 +1,8 @@
...
-  babel: {},
-  image: {}
-}) => {
+module.exports = (neutrino, options = {}) => {
   neutrino.config
-    .when(options.babel, () => neutrino.use(babelMinify, options.babel))
-    .when(options.image, () => neutrino.use(imageMinify, options.image));
+    .when(options.babel, () => neutrino.use(babelMinify, options.babel !== true && options.babel || {}))
+    .when(options.image, () => neutrino.use(imageMinify, options.image !== true && options.image || {}));
 }
```
timkelty: Is this weird? I added this so users (or other preset) could pass `true` instead of `{}` to enable with defaults. Maybe not needed.

packages/minify/index.js
```diff
@@ -1,5 +1,8 @@
-const BabelMinify = require('babel-minify-webpack-plugin');
+const babelMinify = require('@neutrinojs/babel-minify');
+const imageMinify = require('@neutrinojs/image-minify');
 
-module.exports = ({ config }, options = {}) => config
-  .plugin('minify')
-  .use(BabelMinify, [options.minify, options.overrides]);
+module.exports = (neutrino, options = {}) => {
+  neutrino.config
```
eliperelman: Maybe this would be easier by breaking it into merged options:

```js
const options = merge({
  babel: opts.babel ? {} : false,
  image: opts.image ? {} : false
}, opts);

neutrino.config
  .when(options.babel, () => neutrino.use(babelMinify, options.babel))
  .when(options.image, () => neutrino.use(imageMinify, options.image));
```

Thoughts?

timkelty: Probably. Easier to read.

timkelty: Hmmm that would still just end up with values as `true` though

eliperelman: Oh derp.

eliperelman: *nothing to see here*

timkelty: Do you think we should care about this? Do we want to avoid users "enabling" by passing `{}`?

eliperelman: No, passing an empty object should be a legit way to enable this. I think you were on the right track, passing `true` should coerce to object, and then merge the options.

timkelty: > Do we want to avoid users "enabling" by passing {}?

Sorry I meant do we want to avoid that being the _only_ way of enabling (aside from actually passing options).

eliperelman: Could you elaborate?

timkelty: `{}` will always be a legit way of enabling. Just verifying that we _want_ to bother coercing `true` values to `{}`.

eliperelman: Ah I see. Yeah, when you think about it from an enabled standpoint, using `true` doesn't make a whole lot of sense when used with an item that starts off in an enabled state. If `minify` already has both items enabled, then passing true doesn't seem relevant. So I guess my answer would be don't bother with `true` from the `minify` middleware.

From the perspective of the `web` middleware though, `images` would start off disabled, so maybe it's `web` that allows a `true` option instead.

timkelty: Sounds good - I'll coerce there instead.

packages/web/index.js
```diff
@@ -72,6 +73,10 @@ module.exports = (neutrino, opts = {}) => {
   }
 
   Object.assign(options, {
+    minify: {
+      babel: options.minify.babel !== true && options.minify.babel || {},
+      image: options.minify.image !== true && options.minify.image || {}
+    },
```
timkelty: @eliperelman how's this?

packages/web/index.js
```diff
@@ -184,7 +191,7 @@ module.exports = (neutrino, opts = {}) => {
     })
     .when(process.env.NODE_ENV === 'production', (config) => {
       neutrino.use(chunk);
-
+      console.log(options.minify);
```
eliperelman: Looks like this crept in. Remove this line.

timkelty:
Ok, removed that console and rebased off master.

timkelty:
@eliperelman Also, I have `style-minify` ready to go once this is merged.

eliperelman:
@timkelty I'm still showing conflicts...

timkelty:
@eliperelman weird, after a successful rebase of origin/master, it still showed 1 conflict…

I merged master in instead and it seems resolved.

packages/minify/README.md
```diff
@@ -1,8 +1,11 @@
 # Neutrino Minify Middleware
 
-`@neutrinojs/minify` is Neutrino middleware for minifying source code using
-[`BabelMinifyWebpackPlugin`](https://www.npmjs.com/package/babel-minify-webpack-plugin). This middleware is usually only
-added during production builds.
+`@neutrinojs/minify` is Neutrino middleware for minifying assets. It is an aggregate of the following middlewares:
+
+- [@neutrinojs/babel-minify](../../packages/babel-minify/README.md)
```
eliperelman: For the `packages` directory these should be absolute URLs to `https://neutrino.js.org`. They are relative when in `docs/packages`.

packages/minify/README.md
```diff
@@ -57,32 +60,26 @@ module.exports = {
...
 
-### Plugins
-
-The following is a list of plugins and their identifiers which can be overridden:
+See the following middlewares for added plugins and rules:
 
-| Name | Description | Environments and Commands |
-| --- | --- | --- |
-| `minify` | Minifies source code using `BabelMinifyWebpackPlugin`. | all |
+- [@neutrinojs/babel-minify](../../packages/babel-minify/README.md)
```
eliperelman: Absolute URLs here.

timkelty:
@eliperelman Shall we merge?

eliperelman:
dance

PR #530, opened 2017-12-04
Migrate clean options from v7 branch to master to allow all options for clean middleware

eliperelman:
See #511 for original implementation.

### DIFF
--- a/packages/web/README.md
+++ b/packages/web/README.md
@@ -245,6 +245,12 @@ module.exports = {
         removeDebugger: true,
       },
 
+      // Remove the contents of the output directory prior to building.
+      // Set to false to disable cleaning this directory
+      clean: {
+        paths: [neutrino.options.output]
+      },
+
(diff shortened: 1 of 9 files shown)


packages/web/README.md
```diff
@@ -245,6 +245,12 @@ module.exports = {
         removeDebugger: true,
       },
 
+      // Remove the contents of the output directory prior to building.
+      // Set to false to disable cleaning this directory
+      clean: {
+        paths: [neutrino.options.output]
```
edmorley: Ah - this addition is to a different part of the file than that in `docs/packages/web/README.md`. I don't suppose you could open a PR to sync them up? (The more things are in sync, the easier it will be when I come to land the docs work)

eliperelman: Definitely!

eliperelman: Addressed in #544.

PR #532, opened 2017-12-04
Use extensions in outstanding package

helfi92:


helfi92:
A couple of points to address:

1. How do you feel if we expose a `mediaExtensions` property on `neutrino.options` and rename the `extensions` propery to `moduleExtensions`? `mediaExtensions` would have `['png', 'jpg', 'jpeg', gif', 'webp']` as the default value so we could also use it in the image loader.

2. Adding `join(basename(neutrino.options.source), '**/*.jsx'` to the `collectCoverageFrom` jest option throws when when doing `yarn test --coverage` on a react project. Is that a bug? Is there a reason why we use different babel presets when testing? Error:

screen shot 2017-12-04 at 11 14 19 am

eliperelman:
> How do you feel if we expose a `mediaExtensions` property on `neutrino.options` and rename the `extensions` propery to `moduleExtensions`? `mediaExtensions` would have `['png', 'jpg', 'jpeg', gif', 'webp']` as the default value so we could also use it in the image loader.

This is only useful to a single middleware package, vs. the current extensions which is useful to most middleware, so I'm not sure what value this would add.

> Adding `join(basename(neutrino.options.source), '**/*.jsx'` to the `collectCoverageFrom` jest option throws when when doing `yarn test --coverage` on a react project. Is that a bug? Is there a reason why we use different babel presets when testing?

This would be a bug, not sure when this crept up. They use different babel presets because we don't want to inject jest globals into application code, and also because jest **does not use webpack** to compile application code, only babel, so we need to enable transforming modules, which normally webpack takes care of.

packages/jest/src/index.js
```diff
@@ -23,6 +21,9 @@ function getFinalPath(path) {
 }
 
 function normalizeJestOptions(opts, neutrino, usingBabel) {
+  const mediaNames = `\\.(${mediaExtensions.join('|')})`;
```
eliperelman: why not `regexFromExtensions` on these ones?

packages/jest/src/index.js
```diff
@@ -42,6 +43,10 @@ function normalizeJestOptions(opts, neutrino, usingBabel) {
     ...(opts.moduleFileExtensions || []),
     ...neutrino.config.resolve.extensions.values().map(e => e.replace('.', ''))
   ])];
+  const testRegex = join(
+    basename(neutrino.options.tests),
+    `.*(_test|_spec|\\.test|\\.spec)\\.(${neutrino.options.extensions.join('|')})$`
```
eliperelman: Why not `regexFromExtensions` here?

packages/neutrino/src/api.js
```diff
@@ -18,7 +18,7 @@ class Api {
 
   // getOptions :: Object? -> IO Object
   getOptions(opts = {}) {
-    let moduleExtensions = new Set(['js', 'jsx', 'vue', 'ts', 'mjs', 'json']);
+    let moduleExtensions = new Set(['js', 'jsx', 'vue', 'ts', 'tsx', 'mjs', 'json']);
```
eliperelman: What is `tsx`? TypeScript with JSX?

helfi92: Yep. https://www.typescriptlang.org/docs/handbook/jsx.html

packages/jest/src/index.js
helfi92: regexFromExtensions uses RegExp which returns an object. Below in the code, we have:

```js
  const moduleNameMapper = merge({
    [mediaNames]: require.resolve('./file-mock'),
    [styleNames]: require.resolve('./style-mock')
  }, moduleNames);
```

I thought that adding a RegExp object as a property would blow up (now that i think about it, i think it will just try to call its `toString` method). Also, RegExp adds a leading and a trailing slash to every expression (e.g., `/.jsx/`) which I'm not sure how Jest handles it. The options showed in the documentation in `[moduleNameMapper](https://facebook.github.io/jest/docs/en/configuration.html#modulenamemapper-object-string-string)` didn't have those leading and trailing slashes. Would it be okay to have them?

eliperelman: I'm OK with leaving it how you already have it, I just wasn't for sure. Thanks for looking into that!

packages/jest/src/index.js
eliperelman: I see why now. I keep thinking `regexFromExtensions` returns a string regex representation, not an actual regex. My bad.

eliperelman:
There's a conflict here, most likely from me removing the `json` extension. Could you rebase?

helfi92:
Rebased.

PR #533, opened 2017-12-04
Remove json from being in options.extensions, update docs on extensions accordingly

eliperelman:
This fixes issues with the babel compiler trying to parse JSON as JS, and prevents ESLint from trying to lint JSON files. As a result, we need to add `json` to webpack extensions where relevant, which this PR also does. I've updated the docs accordingly.


eliperelman:
This also fixes the comment: https://github.com/mozilla-neutrino/neutrino-dev/pull/497#discussion_r154629773

helfi92:
👍

Issue #516, opened 2017-12-03
Extend neutrino options with project package.json


eliperelman:
There are some places where we access the package.json in the cwd. We should add this as an option, something like `neutrino.options.package` or `neutrino.options.packageJson`. I'm open to names.

Then we can utilize this in the presets that use this.

eliperelman:
Fixed with #534.

PR #534, opened 2017-12-04
Add option to get package.json content

helfi92:
References https://github.com/mozilla-neutrino/neutrino-dev/issues/516.

### DIFF
--- a/packages/library/index.js
+++ b/packages/library/index.js
@@ -64,7 +64,8 @@ module.exports = (neutrino, opts = {}) => {
   }
 
   try {
-    const pkg = require(join(neutrino.options.root, 'package.json')); // eslint-disable-line global-require
+    const pkg = neutrino.options.packageJson;
+
     const hasSourceMap = (pkg.dependencies && 'source-map-support' in pkg.dependencies) ||
       (pkg.devDependencies && 'source-map-support' in pkg.devDependencies);
 
--- a/packages/library/index.js
+++ b/packages/library/index.js
@@ -64,7 +64,7 @@ module.exports = (neutrino, opts = {}) => {
   }
 
   try {
-    const pkg = neutrino.options.pkgJson;
+    const pkg = neutrino.options.packageJson;
 
     const hasSourceMap = (pkg.dependencies && 'source-map-support' in pkg.dependencies) ||
       (pkg.devDependencies && 'source-map-support' in pkg.devDependencies);
--- a/packages/library/index.js
+++ b/packages/library/index.js
@@ -64,7 +64,7 @@ module.exports = (neutrino, opts = {}) => {
   }
 
   try {
-    const pkg = neutrino.options.packageJson;
+    const pkg = neutrino.options.pkgJson;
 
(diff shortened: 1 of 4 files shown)


packages/neutrino/src/api.js
```diff
@@ -49,6 +50,8 @@ class Api {
       });
     });
 
+    options.packageJson = require(path.join(options.root, 'package.json')); // eslint-disable-line global-require
```
eliperelman: The section you removed from the node preset is essentially what we need here. We need to try to require the file, but not blow up if it's not there. Maybe we should consider setting this to an empty object? Not sure...

eliperelman: Meh, we can leave it undefined.

eliperelman:
Any thoughts on the option name? While `packageJson` works, it seems kinda awkward. cc: @edmorley

helfi92:
I hear you, I think it looks awkward too. `pkgJson` is slightly better. I don't like `package` just because it's hard to know what kind of package it is by simply looking at the name.

eliperelman:
I'm just annoyed at the `Json` part. I think I prefer `package` over `packageJson`, but neither make me happy. Maybe there's another word we can use?

eliperelman:
Meh, if anything we should probably roll with `packageJson` since it's the most descriptive and relevant. Let's see what @edmorley thinks.

edmorley:
`packageMetadata`, `projectMetadata`, ...?

I can't think of a great name either really! :-)

helfi92:
OK, reverting back to `packageJson` 🤢 hehe

packages/neutrino/src/api.js
```diff
@@ -49,6 +50,12 @@ class Api {
       });
     });
 
+    try {
+      options.packageJson = require(path.join(options.root, 'package.json')); // eslint-disable-line global-require
+    } catch (err) {
+      options.packageJson = undefined;
```
eliperelman: Let's use `null` here instead.

packages/neutrino/src/api.js
```diff
@@ -2,6 +2,7 @@ const Config = require('webpack-chain');
 const merge = require('deepmerge');
 const Future = require('fluture');
 const mitt = require('mitt');
+const path = require('path');
```
eliperelman: Destructure out `join`.

packages/library/index.js
```diff
@@ -64,7 +64,8 @@ module.exports = (neutrino, opts = {}) => {
   }
 
   try {
-    const pkg = require(join(neutrino.options.root, 'package.json')); // eslint-disable-line global-require
+    const pkg = neutrino.options.packageJson;
+
```
eliperelman: Remove extra blank line.

eliperelman:
Also need to add blurbs to the docs about this option.

eliperelman:
We can do that in another bug if you like.

helfi92:
Yes, will open up a different PR for this. Thanks for the reminder.

PR #535, opened 2017-12-05
Udate CHANGELOG.md

rodmoreno:


rodmoreno:
What do you think about adding the following script to the package.json?

```json
"prerelease": "yarn version",
```

eliperelman:
@rodmoreno I *think* that is probably the right move. I don't think auto-changelog is automatically updating it when we release via lerna. @cookpete could you confirm it should work this way, and whether we have the right command in place in our package.json?

CookPete:
Updating the changelog manually like this shouldn't be necessary if things are set up correctly. I don't know much about lerna but if there is an [`npm version`](https://docs.npmjs.com/cli/version) equivalent that runs every time you create a version commit, you should be able to add `yarn changelog` and never worry about doing it manually.

Also, since https://github.com/mozilla-neutrino/neutrino-dev/pull/524 the changelog will be very different to this, with many more correctly parsed merge commits.

eliperelman:
I did some digging into lerna, and it looks like they are **supposed** to run the `preversion`, `version`, and `postversion` scripts when using the `lerna publish` command. Unfortunately, looking at our last tag, it appears to not have worked, as no CHANGELOG was generated:

https://github.com/mozilla-neutrino/neutrino-dev/commit/b5f9d68af42f663a63ee06c581b77c2041d018af

@CookPete maybe since we used the `--package` flag, it's possible that nothing was updated since this could have happened prior to the tag actually being created?

CookPete:
Yep it looks like there is some work to be done for `auto-changelog` to support lerna properly. I'll look into it.

eliperelman:
Wow, thanks! Let me know if there's anything you need support-wise from me.

CookPete:
It looks like Lerna doesn't have an `npm version` equivalent that we can hook in to. There is a small discussion [here](https://github.com/lerna/lerna/issues/643) but no real solution yet.

You can add `"version": "auto-changelog -p && git add CHANGELOG.md"` to each package, which works, but will generate the same `CHANGELOG.md` inside each package directory, not in the root folder.

Annoyingly, Lerna has built in support for generating a [`conventional-changelog`](https://github.com/conventional-changelog/conventional-changelog) if you are using [conventional commits](https://conventionalcommits.org/), but no way of hooking in your own scripts to that process instead.

eliperelman:
Hmm ok, that's unfortunate. We will keep an eye on this then.

CookPete:
Note that the changelog still needs updating with the latest `auto-changelog` update since #524. The current changelog still contains unparsed merges.

eliperelman:
@CookPete running the changelog script in master just now produces the following changes:

https://github.com/mozilla-neutrino/neutrino-dev/compare/master...eliperelman:changelog-discrepancies

Which removed the v7.3.3 and v7.4.0 tags.

CookPete:
Looks good but yeah the missing versions is a bit weird. Do you have the latest commits/tags pulled down locally? The tags need to exist on the machine that is generating the changelog.

CookPete:
@eliperelman Did you have the right branch checked out when you ran it, ie the branch with `v7.4.0` in? I just checked out and ran `auto-changelog` (on branch `release/v7`) and it seemed fine.

eliperelman:
I was in master when I ran it...

eliperelman:
But yeah, it does work correctly when in the release branch.

PR #536, opened 2017-12-05
Revert broken changes, do not join.

timkelty:
@eliperelman changes introduced here broke things a bit: https://github.com/mozilla-neutrino/neutrino-dev/commit/9380f02e983e05e2780bb0a5a14e490d4a45891a#diff-3813a5588840cf846f7dfcf2cc0bc30b

This MW basically assumes you're using `image-loader`, or something else that would define image rules. Those pre-existing rule ids are what is configurable via `options.rules`.

For that reason, I was intentionally omitting the `.test`, as we can assume it already exists in that rule.

I'm then mapping the value of the actual test, so I can pass it all to the plugin's `test` option.

Your changes were using `options.rules` as extensions, and calling `.test`, which ended up passing something like `/(\.svg|\.img)$/`

Note:
I _was_ `join`ing those rules on a pipe for a regex string, which I didn't need to because `ImageminWebpackPlugin`'s test option can be an array.

**TL;DR**

This PR reverts those changes with the exception of the unnecessary `join('|')`. If you have other ideas or preferences how to do this, I'm happy to make changes.

### DIFF
--- a/packages/imagemin/index.js
+++ b/packages/imagemin/index.js
@@ -27,22 +27,19 @@ module.exports = (neutrino, opts = {}) => {
     rules: ['svg', 'img']
   }, opts);
 
-  const tests = options.rules.map((ruleId) => {
-    const test = neutrino.regexFromExtensions([ruleId]);
-
+  const test = options.rules.map((ruleId) => {
     neutrino.config.module
       .rule(ruleId)
-      .test(test)
       .use(options.useId)
         .loader(imageminLoader)
         .options(options.imagemin);
 
-    return test;
+    return neutrino.config.module.rule(ruleId).get('test');
   });
 
   options.plugin = merge({
     imageminOptions: options.imagemin,
-    test: tests
+    test
   }, options.plugin);
 
   neutrino.config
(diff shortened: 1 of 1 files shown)


packages/imagemin/index.js
```diff
@@ -27,22 +27,20 @@ module.exports = (neutrino, opts = {}) => {
...
-      .test(test)
       .use(options.useId)
         .loader(imageminLoader)
         .options(options.imagemin);
 
-    return test;
-  });
+    return neutrino.config.module.rule(ruleId).get('test');
+  })
+  .filter(test => !!test);
```
eliperelman: This can be `.filter(Boolean)`.

timkelty: Of course. Old habit 🙈

timkelty:
> I changed this during one of my refactors, I think for create-project tests, because I couldn't get some tests to pass. Maybe this is gone now though and we can move forward. Have you tested that this functions properly?

Yeah - actually I was testing other stuff w/ create-project and thats how I discovered it _wasn't_ working. I will do some more testing to be certain.

timkelty:
Working in my tests

PR #537, opened 2017-12-05
Merge test babel config in root config to deal with JSX pragma plugin precedence

eliperelman:
The command `neutrino test --coverage` command was blowing up when used with the jest preset. Doing a bisect revealed the culprit as #455. After digging it, it seems that the Babel pragma plugins add some import code, which is expected, but jest (and the other test middleware) adds its plugin for imports in the `test` env, which is lower precedence than the root configuration for Babel.

To be honest, since we are controlling adding to the babel configuration to occur only during the test event, using the Babel `test` env is really just a novelty. I have removed the environment here from the test runners.

We may want to consider removing all Babel environments and replacing them with `.when` merges, just to avoid things like this in the future, and to give us a single piece of config to update. Same for our users.

### DIFF
--- a/packages/jest/src/index.js
+++ b/packages/jest/src/index.js
@@ -83,13 +83,11 @@ module.exports = (neutrino, opts = {}) => {
 
     neutrino.config.when(usingBabel, () => {
       neutrino.use(loaderMerge('compile', 'babel'), {
-        env: {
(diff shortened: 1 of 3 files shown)


eliperelman:
Filed #542 to address future babel env removals.

PR #541, opened 2017-12-05
Add neutrino.options.source to modules.resolve

timkelty:
Anyone else find it odd that `neutrino.options.source` isn't in our resolve paths?

Could be a good reason for it, but it is always one of the first things I add to a neutrino setup. I generally perfer to import that way vs a relative path with `./`.

Fine if we don't want it in core, just thought I'd bring it up if others feel the same.

### DIFF
--- a/packages/node/index.js
+++ b/packages/node/index.js
@@ -101,6 +101,7 @@ module.exports = (neutrino, opts = {}) => {
         .add(neutrino.options.node_modules)
         .add(MODULES)
         .add(NEUTRINO_MODULES)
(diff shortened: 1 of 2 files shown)


eliperelman:
I don't typically do this in my projects, and it also has the potential to be a breaking change if a local file name collides with one in `node_modules`. I don't mind exploring this as an option to enable, but I think making this the default could let to situations of unintended expectations.

timkelty:
Nah thats cool, we can close for now.

PR #549, opened 2017-12-06
Use html-template in all web commands, not just start

eliperelman:
Fixes https://github.com/mozilla-neutrino/neutrino-dev/pull/509#issuecomment-349524253.

Looks like I put the template code in the wrong `forEach`. 😞


eliperelman:
@edmorley no worries!

PR #559, opened 2017-12-07
Simplify react-components to building instead of dictating preview app

eliperelman:
Maintaining the react previewer components just seemed to muddy up the preset, and prevented users from using their own custom previewer preferences. This PR removes our opinion, leaving the choice to the user.


eliperelman:
Not really a breaking change since this isn't released. If this had been released, it would be breaking. 😺

Issue #20, opened 2017-02-14
Create presets for SASS and LESS


eliperelman:

sotayamashita:
@eliperelman I'm creating SASS one. After finishing it and it is ok, could I transfer ownership to this organisation?

eliperelman:
@sotayamashita if you would like the SASS preset to live with Neutrino, it would need to be placed in `packages` with appropriate docs. Could you do that, or would you like to keep this separate?

sotayamashita:
@eliperelman I could do that. After creating it, I'll share the repo and please review this.

eliperelman:
Absolutely!

kopacki:
Hey @sotayamashita! I couldn't find the preset you were about to created, so I published it myself.
I read that additional presets should be maintained on external repositories and I am not sure now if I should create a PR for my work.
Anyway, I'd like to ask you and @eliperelman to review that small [SASS preset ](https://github.com/kopacki/neutrino-dev/tree/master/packages/neutrino-preset-sass). I am going to extend it with optional PostCSS as well.

sotayamashita:
@kopacki Sorry, I have not published my code yet because I am busy. I appreciate your work. I will create PR for your commit. :+1: I will implement for less.

kopacki:
I must say I have huge problem to add `extract-text-webpack-plugin` to the preset. Help with that one would be much appreciated.

So far I've done following:
```javascript
const ExtractTextPlugin = require('extract-text-webpack-plugin');

module.exports = function({ config }) {
  config
    .plugin('extractCss')
      .use(ExtractTextPlugin, {
        filename: 'style.css',
        disable: false,
        allChunks: true
      });

  const styleRule = config.module
    .rule('scss');

  ExtractTextPlugin
    .extract({
      fallback: 'style-loader',
      use: ['css-loader', 'sass-loader']
    })
    .forEach(({ loader, options }, index) => styleRule.loader(index, loader, options));
}
```
But it causes compilation errors I haven't resolved yet.

hugufc:
Depending on v5 release date I think it could be a neutrino-middleware-sass.

barraponto:
FWIW, I created a PostCSS base preset. PostCSS supports parsing LESS/SASS/Stylus/SugarSS so I guess you can build on top of it. https://github.com/barraponto/neutrino-preset-postcss

barraponto:
[...]

PR #566, opened 2017-12-07
Allow style-loader to accept loaders

timkelty:
This adds a `loaders` option to `style-loader`, in which users may pass their own custom css loaders (postcss, sass, etc)

## Why?

- CSS loaders can be tricky to configure correctly, especially when factoring in `extract-text`, css modules, hot-reloading, and css-loader's `importLoaders` option.
- Prevents the need for community-sourced middlewares that do nothing more than add a loader to the `style` rule (e.g. https://github.com/barraponto/neutrino-middleware-postcss, https://github.com/kopacki/neutrino-dev/blob/master/packages/neutrino-preset-sass/src/index.js)

## Example usage

```
module.exports = {
  use: [
    [
      '@neutrinojs/web',
      {
        style: {
          loaders: [

            // No options, auto-generated useId (css-*)
            require.resolve('postcss-loader'),

            // With options, specified useId
            {
              useId: 'postcss',
              loader: require.resolve('postcss-loader'),
              options: {}
            }
          ]
        }
      }
    ],
  ]
};
```
- The result is the loaders added in order: style, css, sass-loader, postcss-loader.
- css-loaders's `importLoaders` has been set properly, taking into account the # loaders your have, and `options.modules`.
- Everything still works with extract-text, and HMR :)

Related: https://github.com/mozilla-neutrino/neutrino-dev/issues/20

### DIFF
--- a/packages/style-loader/index.js
+++ b/packages/style-loader/index.js
@@ -67,7 +67,14 @@ module.exports = (neutrino, opts = {}) => {
         importLoaders: options.css.importLoaders + options.loaders.length
       }),
       useId: options.cssUseId
-    }, ...options.loaders];
+    }, ...options.loaders]
+    .map((loader, index) => {
+      const obj = typeof loader === 'object' ? loader : { loader };
+
+      return Object.assign(obj, {
+        useId: obj.useId || `${options.cssUseId}-${index}`
+      });
+    });
 
     loaders.forEach(loader => {
       styleRule
--- a/packages/style-loader/index.js
+++ b/packages/style-loader/index.js
@@ -56,18 +56,21 @@ module.exports = (neutrino, opts = {}) => {
 
   rules.forEach(options => {
     const styleRule = neutrino.config.module.rule(options.ruleId);
-    const loaders = [{
-      loader: require.resolve('style-loader'),
-      options: options.style,
-      useId: options.styleUseId
-    },
-    {
-      loader: require.resolve('css-loader'),
-      options: Object.assign(options.css, {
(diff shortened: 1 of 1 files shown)


packages/style-loader/index.js
```diff
@@ -52,23 +55,40 @@ module.exports = (neutrino, opts = {}) => {
   }
 
   rules.forEach(options => {
-    neutrino.config.module
-      .rule(options.ruleId)
+    const styleRule = neutrino.config.module.rule(options.ruleId);
+    const loaders = [{
```
eliperelman: Let's mess with the formatting a little here:

```js
const loaders = [
  {
    ...
  },
  {
    ...
  },
  ...options.loaders
]
.map(() => {
  ...
});
```

packages/style-loader/index.js
```diff
@@ -52,23 +55,40 @@ module.exports = (neutrino, opts = {}) => {
...
+      options: Object.assign(options.css, {
+        importLoaders: options.css.importLoaders + options.loaders.length
+      }),
+      useId: options.cssUseId
+    }, ...options.loaders]
+    .map((loader, index) => {
+      const obj = typeof loader === 'object' ? loader : { loader };
+
+      return Object.assign(obj, {
+        useId: obj.useId || `${options.cssUseId}-${index}`
```
eliperelman: I like this, but this will need to be clearly noted in the documentation, that without the `useId`, it will generate an id in this format.

eliperelman: I already have the docs writte in #574.

PR #567, opened 2017-12-07
Add a webextension preset

helfi92:


helfi92:
Feel free to use https://github.com/helfi92/webextension-neutrino-test to test this preset.

`web-ext` provides the following commands which the current commit doesn’t handle:

* lint (uses [mozilla/addons-linter](https://github.com/mozilla/addons-linter))
* [sign](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#web-ext_sign)

For linting, should we simply create a preset for it. What about the sign command?

helfi92:
To develop with this preset, you first need to `yarn build` to generate the build directory. `yarn start` will open a firefox instance reading from the build directory. the build directory is not meant to be touched. Change the source through the `src` directory. When that happens, the extension on that firefox instance will reload instantly.

helfi92:
I added react specific babel presets to support building extensions using React. Should this be included as a default? I'm not sure we should include React specifics in the preset...

eliperelman:
Most of this seems like a copy of the Web preset. Can we just use web and disable the pieces we don't need?

eliperelman:
For react, I think we probably don't need it. At least right now.

helfi92:
> Most of this seems like a copy of the Web preset. Can we just use web and disable the pieces we don't need?

My only concern is that adding something to the web preset in the future may break the web-extension preset. If you still feel like this shouldn't be a concern, I am okay with it.

> For react, I think we probably don't need it. At least right now.

Sure thing. Maybe we could add the `react` bit to the scaffolding part.

eliperelman:
> My only concern is that adding something to the web preset in the future may break the web-extension preset. If you still feel like this shouldn't be a concern, I am okay with it.

Understood, and I'm not concerned about it. If there is something that causes breakage in another preset, we should consider that to be a breaking change anyway, and something worthy of a major version bump. We shouldn't be purposefully breaking middleware.

> Sure thing. Maybe we could add the react bit to the scaffolding part.

Yeah we can consider it. Maybe it would be cool if it could be composed like so, but that does assume ordering, which I am not the biggest fan of:

```js
module.exports = {
  use: [
    '@neutrinojs/web',
    '@neutrinojs/webextension'
  ]
};

module.exports = {
  use: [
    '@neutrinojs/react',
    '@neutrinojs/webextension'
  ]
};
```

I guess we could say something about composition, but I still haven't figured out what a nice API for this would look like:

```js
module.exports = {
  use: [
    ['@neutrinojs/webextension', {
      extends: '@neutrinojs/react'
    }]
  ]
};
```

eliperelman:
As an aside, I think this will be named `webextension` instead of `web-extension`, given that they are specifically named `WebExtensions`.

https://developer.mozilla.org/en-US/Add-ons/WebExtensions/What_are_WebExtensions

helfi92:
I updated https://github.com/helfi92/webextension-neutrino-test to help test and experiment with the preset.

helfi92:
@kumar303 I am creating a preset for creating web-extensions using Neutrino. I was wondering if you could give me some feedback on the current pull-request. Thanks in advance!

packages/webextension/index.js
```diff
@@ -0,0 +1,58 @@
...
+        .add(MODULES)
+        .end()
+      .end()
+    .resolveLoader
+      .modules
+        .add(MODULES)
+        .end()
+      .end()
+    .when(neutrino.options.command === 'start', () => {
+      webExt.cmd.run({
```
kumar303: This returns a promise which could be rejected in some situations. I don't see `.catch()` handled. At the very least it should log the error because I don't think web-ext will do that in the `run()` function. You could test it by omitting `sourceDir` because that should reject the promise with some kind of error.

packages/webextension/index.js
```diff
@@ -0,0 +1,58 @@
...
+        noInput: true,
+        sourceDir: neutrino.options.output
```
kumar303: If this directory contains any unimportant build artifacts then you should consider passing an [`ignoreFiles`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#--ignore-files_-i) array to `run()`.

`run()` will start a file watcher within `sourceDir`. When a file gets touched, it will reload the extension unless the file matches any pattern in `ignoreFiles`.

It's important that the `web-ext` file watcher does *not* look for changes in the actual source files. It should only look for changes in the final executable files that are built from the source (I think that's how you have it configured). This is because if `web-ext` starts to reload the extension when a source file changes, it might reload too soon, before the final files are built by babel/webpack/whatever.

packages/webextension/index.js
kumar303: I don't know much about neutrino but is there a way to make some `run()` parameters configurable by the user? I can see a couple parameters [here](https://github.com/mozilla/web-ext/blob/1b22d60aefa31e2869e15d18ec52b39afa0de405/src/cmd/run.js#L31) that the user may want to configure such as `firefox` (path to executable), `firefoxProfile`, `browserConsole`, `pref`...

helfi92: Yep. I made all the options configurable in my latest commit. That was a good point!

helfi92: Agreed. I added a `.catch()` in my last commit.

packages/webextension/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "neutrino",
+    "neutrino-preset",
+    "web-extension"
+  ],
+  "author": "Hassan Ali <helfi92@gmail.com>",
+  "license": "MPL-2.0",
+  "repository": "https://github.com/mozilla-neutrino/neutrino-dev/tree/master/packages/webextension",
+  "homepage": "https://neutrino.js.org",
+  "bugs": "https://github.com/mozilla-neutrino/neutrino-dev/issues",
+  "dependencies": {
```
eliperelman: We will need the `publishConfig` and `engines` copied here.

packages/webextension/index.js
```diff
@@ -0,0 +1,71 @@
...
+        .add(MODULES)
+        .end()
+      .end()
+    .resolveLoader
+      .modules
+        .add(MODULES)
+        .end()
+      .end()
+    .when(neutrino.options.command === 'start', () => {
+      webExt.cmd
```
eliperelman: I see where we are going with this, but exiting the process from here is going to skip all the error handling we have in the API.

What if we took this part out of this conditional and use an event?

```js
neutrino.on('start', () => webExt.cmd.run(options.webExtRun, { shouldExitProgram: false }));
```

This way would mean we could skip handling the promise altogether and let the API handle it. Give it a shot and let me know if that works.

Morikko:
I just checked, it looks promising.

However a few questions: 
1. Is it mandatory to do `neutrino build` before `neutrino start`. If I did only the second command, the copying files were missing (e.g. manifest.json...)

2. The problem with question 1 raised a new problem. If you modify a copying file (eg: popup html), the file is not copy again to the build folder. You have to manually do again `neutrino build`.

3. For this wouldn't be better to split the folder used by `neutrino start` and `neutrino build`. The latter could use another folder where the build is optimized for production. 

> Do you think we should perform some extra magic by reading and doing some things if a user provides a manifest.json?

What do you mean ?
Are you thinking about finding the different entries from the manifest ?

Also allowing web-ext to read a config.js file would be nice to let the user choose the profile, firefox executable and so on.

Finally, a command like `neutrino release` would be useful for generating the .xpi (for FF) and .zip (for chrome) from the optimized build.

eliperelman:
> For this wouldn't be better to split the folder used by `neutrino start`

My preference would be for the `start` command to build files to a temp directory, or in memory.

> What do you mean ?

For example, if the user sets the popup file in the manifest.json, should we automatically add it to Neutrino's mains?

> Are you thinking about finding the different entries from the manifest ?

Yes, that's the gist of it. :)

> Also allowing web-ext to read a config.js file would be nice to let the user choose the profile, firefox executable and so on.

We would like this to be driven by settings in `.neutrinorc.js`, just so it's consistent with every other way of configuration across presets/middleware.

kumar303:
> Also allowing web-ext to read a config.js file would be nice to let the user choose the profile, firefox executable and so on.

Aside from hooking into the standard ways of configuring neutrino presets, I should mention that web-ext now supports [standalone config files](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Getting_started_with_web-ext#Setting_option_defaults_in_a_configuration_file). The config feature was added after this PR began. A `webExt` key in `package.json` is also supported but this isn't documented yet (FYI).

Morikko:
> My preference would be for the start command to build files to a temp directory, or in memory.

Make sense, it respects the standard behavior.

> We would like this to be driven by settings in .neutrinorc.js, just so it's consistent with every other way of configuration across presets/middleware.

@kumar303 said what I was thinking for. Everything is already provided by web-ext and would avoid to bind too much `.neutrinorc.js` to web-ext.

>   Could this happen in the build command?

Yes, it would avoid new commands.

@helfi92 do you have some plan to go ahead ?
I want to start bundling my extension, so I am really interested :)

@kumar303  What about chome/chromium support ?
Because using web-ext in this preset breaks the flow for chrome (only) extension...

Morikko:
@eliperelman One question about the PR workflow.
I would like to use the last version of neutrino to continue the work. I have cloned the PR locally.
Can I rebase it (Commit hash is changed) or should I do a merge to master?

So, after my commits will be added on top of helfi92's ones in this PR.

eliperelman:
I would recommend rebasing against master. If you need a guide of how to do a rebase, I would recommend giving this a read:

http://blessing.io/git/git-rebase/open-source/2015/08/23/git-rebasing-to-resolve-conflicts.html

Morikko:
In neutrino by adding an entry in mains, can you compile only a JS without an HTML ?

Morikko:
```
module.exports = {
  options: {
    mains: {
      index: 'index', // outputs index.html from src/index.*
      admin: 'admin', // outputs admin.html from src/admin.*
      account: 'user' // outputs account.html from src/user.*
    }
  },
  use: ['@neutrinojs/web']
}
```
This creates HTML files (after building). 
Can I create just a loadable JS file ?

Also should I install this locally https://github.com/mozilla-neutrino/babel-minify-webpack-plugin ?
I got 
```
{ Error: Cannot find module '@neutrinojs/babel-minify-webpack-plugin'
    at Function.Module._resolveFilename (module.js:538:15)
    at Function.Module._load (module.js:468:25)
    at Module.require (module.js:587:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/home/eric/GitGalaxy/Web/neutrino/neutrino-dev/packages/babel-minify/index.js:1:89)
    at Module._compile (module.js:643:30)
    at Object.Module._extensions..js (module.js:654:10)
    at Module.load (module.js:556:32)
    at tryModuleLoad (module.js:499:12)
    at Function.Module._load (module.js:491:3)
    at Module.require (module.js:587:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/home/eric/GitGalaxy/Web/neutrino/neutrino-dev/packages/minify/index.js:2:21)
    at Module._compile (module.js:643:30)
    at Object.Module._extensions..js (module.js:654:10)
    at Module.load (module.js:556:32) code: 'MODULE_NOT_FOUND' }
```

Morikko:
I have [commited](https://github.com/Morikko/neutrino-dev/commit/a383d5376821eef3bfa8cd184e16364f1c87fdf6) new things but it is visible only on my local repository. 
How do I continue on this PR ? (link the commit here)
Isn't it rebase the problem ? I still have the helfi92's commits but they have new hashes.

In brief: 
 - Update to neutrino v8
 - Split folder used by start/build commands
 - Files to copy are also watched
 - Some work to use custom options for web-ext command from `.neutrinorc.js`

Also I am using [this extension](https://github.com/Morikko/webextension-neutrino-test) to test it.

Morikko:
@eliperelman Do neutrino has an event for when a command ends  ? 

Like 
```
api.on('end-start', () => {
  // ...
});
```

Morikko:
Few updates: 
 - Use web-ext lint on start/build/lint
 - Read Manifest for entries

@eliperelman  On `build`, a chunk optimization create a `runtime.js`. For HTML files, it is added with `html-template`. I don't know how to add chunks directly to JavaScript file. As a consequence, all JavaScript files crash on start because they don't find the resources in `runtime.js`.

Note: JavaScript files are background and content scripts.

eliperelman:
> Do neutrino has an event for when a command ends ?

Yes, there is `prestart`, `start`, `prebuild`, `build`, and `prerun`, `run`.

https://neutrino.js.org/api/#events

eliperelman:
See #831 for next steps here, or if you think we should just close this at the moment.

edmorley:
Closing this in favour of #831.

PR #569, opened 2017-12-08
Remove buggy devserver compiler message debouncing

eliperelman:
Debouncing these messages just proved to be buggier than just logging to the console without limit, so for now I'm just removing this until we can put something a little smarter in.


edmorley:
Reading up more, it sounds like there is already a devserver feature for doing what was intended?
https://webpack.js.org/configuration/watch/#watchoptions-aggregatetimeout

PR #577, opened 2017-12-08
Fix ability to disable HMR from react preset

### LINKED ISSUE #576: Regression: Can't set hot option to false in react preset
It runs [this path](https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/react/index.js#L35):

```
[...(options.hot ? [require.resolve('react-hot-loader/babel')] : null)]
```

Which triggers a `TypeError: Cannot read property 'Symbol(Symbol.iterator)' of null`. That null should be `[]`.

eliperelman:
Fixes #576.

### DIFF
--- a/packages/react/index.js
+++ b/packages/react/index.js
@@ -32,7 +32,7 @@ module.exports = (neutrino, opts = {}) => {
         ...(
           process.env.NODE_ENV === 'development'
             ? [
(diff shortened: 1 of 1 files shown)


helfi92 (approved):

eliperelman:
Released in v8.0.1.

PR #584, opened 2017-12-09
Add output path for the library preset

helfi92:


helfi92:
If the path is not defined, the `build` command will output the files at the root level.

PR #585, opened 2017-12-09
Add output path to react-components preset

helfi92:


packages/react-components/index.js
```diff
@@ -68,6 +68,7 @@ module.exports = (neutrino, opts = {}) => {
           .hints('error')
           .end()
         .output
+          .path(neutrino.options.output)
```
eliperelman: Isn't this caught by the web preset?

eliperelman: https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/web/index.js#L134

helfi92: Yes, that was a silly PR. Closing this now :)

eliperelman: lol no worries!

PR #593, opened 2017-12-11
Disable html via web with false

timkelty:


packages/web/index.js
```diff
@@ -114,14 +114,17 @@ module.exports = (neutrino, opts = {}) => {
...
-      neutrino.config.entry(key).add(neutrino.options.mains[key]);
-      neutrino.use(htmlTemplate, merge({
-        pluginId: `html-${key}`,
-        filename: `${key}.html`,
-        // When using the chunk middleware, the names in use by default there
-        // need to be kept in sync with the additional values used here
-        chunks: [key, 'vendor', 'runtime']
-      }, options.html));
+      neutrino.config
+        .entry(key).add(neutrino.options.mains[key])
```
eliperelman: Move `add` to the next line and indent.

timkelty: Done!

PR #595, opened 2017-12-11
Extract text on build, not production

timkelty:
This seems like a better default to me.

In my case I ran `neutrino build --options.env.NODE_ENV = development` and expected to see my css extracted, but not minified/optimized.

### DIFF
--- a/packages/web/index.js
+++ b/packages/web/index.js
@@ -38,7 +38,7 @@ module.exports = (neutrino, opts = {}) => {
     style: {
       hot: opts.hot !== false,
       extract: (opts.style && opts.style.extract) ||
-        (process.env.NODE_ENV === 'production' && {})
+        (neutrino.options.command === 'build' && {})
     },
     manifest: opts.html === false ? {} : false,
     clean: opts.clean !== false && {


eliperelman:
I think I disagree with this. So far we've been running with the idea that `NODE_ENV=production` is what sets off a series of optimizations. Going this route would change that, I think. Any thoughts @mozilla-neutrino/core-contributors @SeeThruHead?

Also, this would be a breaking change, so we couldn't ship this right now. :)

SeeThruHead:
i think changing the file output is usually an optimization of some kind. If the config change has no discernible outcome other than increasing performance it fits nicely into NODE_ENV === production.

At least that's what the current mental model supports.
If this pr were to be accepted it would require another concept to be added. File output formats separate from optimization. You would probably want to add the chunking preset on build as well if you expect build to always output the same files.

eliperelman:
> You would probably want to add the chunking preset on build as well if you expect build to always output the same files.

I think this is an important thing to note because it has established precedent in Neutrino. Text extraction can be seen as a specialized form of chunking, which occurs during optimization. I think for v8 we have to stay the course, and for v9 we can discuss whether this still makes sense. In my opinion, I still see this as an optimization.

timkelty:
Yep, I knew this would be controversial - just threw it up here for discussion :)

My only hang-up is this: when using this is tandem with a server-side app (`options.html: false`), it feels like a bit more than mere optimization, as it would actually change how you load those assets in your view/template.

However - since I can just pass `options.style.extract: {}` to enable extracting, I think this is a totally acceptable solution for my use-case.

Perhaps we should coerce `true` to `{}` for this purpose, like we are in other instances in `web`?

PR #598, opened 2017-12-11
Set default X-Dev-Server-Proxy for easier server-side integration

timkelty:
When using `@neutrinojs/web` with a separate server-side backend, it your backend app will usually need some way of knowing if a request is a dev-server request, or a normal one.

This is so it knows when to look in the `manifest.json` and request the revved file, or to request the un-revved path and load via devServer.

This adds an http header `X-Dev-Server-Proxy`, so you can easily detect from your server-side app.

Example w/ a dumb 'ol php app:

```php
$isDevServer = (bool) $_SERVER['HTTP_X_DEV_SERVER_PROXY'];
```

```html
<script src="<?= $isDevServer ? '/build/index.js' : getManifestAsset('index.js') ?>"></script>
```

### DIFF
--- a/docs/packages/web/README.md
+++ b/docs/packages/web/README.md
@@ -190,8 +190,6 @@ module.exports = {
       html: {},
 
       // Change options related to starting a webpack-dev-server
-      // See @neutrinojs/html-template for the defaults
-      // used by the Web preset
       devServer: {
         // Disabling options.hot will also disable devServer.hot
         hot: options.hot
@@ -254,11 +252,51 @@ module.exports = {
       // Example: change the page title
       html: {
         title: 'Epic Web App'
+      },
+
+      // Example: Proxy webpack-dev-server requests to http://localhost:3000
+      devServer: {
+        proxy: 'http://localhost:3000'
       }
     }]
   ]
 };
 ```
+
+### Dev Server Proxy
+
+If your are integrating Neutrino into an existing app, you may want to set up a proxy for development. See Webpack's [`devServer.proxy`](https://webpack.js.org/configuration/dev-server/#devserver-proxy) for all available options.
+
+Optionally, you may pass a url string (instead of an object) to `devServer.proxy`.
+This will proxy **all requests** through the given url, and set some sensible defaults.
+
+For example:
+```js
+['@neutrinojs/web', {
+  devServer: {
+    proxy: 'http://localhost:3000'
+  }
+}
+```
+
+Is equivalent to:
+```js
+['@neutrinojs/web', {
+  devServer: {
+    proxy: {
+      '**': {
+        target: 'http://localhost:3000',
+        changeOrigin: true,
+        headers: {
+          'X-Dev-Server-Proxy': 'http://localhost:3000'
+        }
+      }
+    }
+  }
+}
+```
+
+The `X-Dev-Server-Proxy` header can be useful for detecting if your existing app is being request through the proxy.
 
 ## Hot Module Replacement
 
--- a/docs/packages/web/README.md
+++ b/docs/packages/web/README.md
@@ -265,12 +265,13 @@ module.exports = {
 
 ### Dev Server Proxy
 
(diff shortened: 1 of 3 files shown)


packages/web/index.js
```diff
@@ -59,7 +59,10 @@ module.exports = (neutrino, opts = {}) => {
     options.devServer.proxy = {
       '**': {
         target: options.devServer.proxy,
-        changeOrigin: true
+        changeOrigin: true,
+        headers: {
```
eliperelman: This is only set when `options.devServer.proxy === 'string'`. Should this go in the main `devServer` defaults above?

timkelty: I don't think so…this, along with the `**` property and `changeOrigin: true` are meant to be be the "easy" way of enabling a proxy.

If you wanted to specify an obj, then I think we assume you know what you're doing and you'll set your own http-proxy-middleware options.

Make sense?

eliperelman: Makes sense to me, we should probably document this though.

timkelty: Indeed…thought we had. I'll add to this PR.

timkelty: Ok I added some docs for devServer.proxy

timkelty:
Would appreciate if anyone could take a peek at this to move it along to :shipit:
/cc @mozilla-neutrino/core-contributors

docs/packages/web/README.md
```diff
@@ -254,12 +252,52 @@ module.exports = {
...
+        proxy: 'http://localhost:3000'
       }
     }]
   ]
 };
 ```
 
+### Dev Server Proxy
+
+If your are integrating Neutrino into an existing app, you may want to set up a proxy for development. See Webpack's [`devServer.proxy`](https://webpack.js.org/configuration/dev-server/#devserver-proxy) for all available options.
```
eliperelman: Webpack should be lower-case.

eliperelman: Also, let's change "existing app" to something like "developing against a server". I'm not for sure.

docs/packages/web/README.md
```diff
@@ -254,12 +252,52 @@ module.exports = {
...
+For example:
+```js
+['@neutrinojs/web', {
+  devServer: {
+    proxy: 'http://localhost:3000'
+  }
+}
+```
+
+Is equivalent to:
```
eliperelman: Insert blank line between sentence and code block.

docs/packages/web/README.md
```diff
@@ -254,12 +252,52 @@ module.exports = {
...
+        headers: {
+          'X-Dev-Server-Proxy': 'http://localhost:3000'
+        }
+      }
+    }
+  }
+}
+```
+
+The `X-Dev-Server-Proxy` header can be useful for detecting if your existing app is being request through the proxy.
```
eliperelman: s/request/requested.

packages/web/README.md
```diff
@@ -254,12 +252,52 @@ module.exports = {
...
+        proxy: 'http://localhost:3000'
       }
     }]
   ]
 };
 ```
 
+### Dev Server Proxy
+
+If your are integrating Neutrino into an existing app, you may want to set up a proxy for development. See Webpack's [`devServer.proxy`](https://webpack.js.org/configuration/dev-server/#devserver-proxy) for all available options.
```
eliperelman: Make sure to apply the requests from one README into the other. :smile:

eliperelman (changes requested):
Just some docs changes.

docs/packages/web/README.md
```diff
@@ -265,12 +265,13 @@ module.exports = {
 
 ### Dev Server Proxy
 
-If your are integrating Neutrino into an existing app, you may want to set up a proxy for development. See Webpack's [`devServer.proxy`](https://webpack.js.org/configuration/dev-server/#devserver-proxy) for all available options.
+If you are handling requests with a server, you may want to set up a proxy for development. See webpack's [`devServer.proxy`](https://webpack.js.org/configuration/dev-server/#devserver-proxy) for all available options.
```
timkelty: @eliperelman how about this?

eliperelman: I like it.

eliperelman (approved):

PR #604, opened 2017-12-12
Fix broken regex

timkelty:
Almost lost my mind trying to figure out why `.scss` were importing without adding a test…

Easy to see why this got missed, as eslint warns of a "useless" escape inside a template literal, but being a regex string, it has meaning.


eliperelman:
:tada:

PR #609, opened 2017-12-13
style-loader: allow overriding options.test/modulesTest overrides with css modules enabled

### LINKED ISSUE #603: style-loader: make it easier to override test without clobbering modules
Someone might want to override the `test` option with something like test: `/\.s?css$/`.

Unfortunately, this isn't going to work well with `modules: true`, as we'd have to replicate what we're doing [here](https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/style-loader/index.js#L9-L18) to exclude `modulesTest` from `test`, etc.

@eliperelman any idea the best way to better support this?

timkelty:
Fixes #603

This PR uses `rule.exclude`, instead of passing `rule.test` a function.
This will allows users to change the tests, while still enabling css modules support:

```
{
        test: /\.s?css$/,
        modulesTest: /\.module\.s?css$/,
}
```

### DIFF
--- a/packages/style-loader/index.js
+++ b/packages/style-loader/index.js
@@ -41,8 +41,7 @@ module.exports = (neutrino, opts = {}) => {
         hotUseId: `${options.hotUseId}${options.modulesSuffix}`,
         extractId: `${options.extractId}${options.modulesSuffix}`,
         css: {
-          modules: options.modules,
-          importLoaders: 1
+          modules: options.modules
         }
       })
     );
--- a/packages/style-loader/index.js
+++ b/packages/style-loader/index.js
@@ -59,6 +59,7 @@ module.exports = (neutrino, opts = {}) => {
       },
       {
         loader: require.resolve('css-loader'),
+        options: options.css,
         useId: options.cssUseId
       },
       ...options.loaders
--- a/packages/style-loader/index.js
+++ b/packages/style-loader/index.js
@@ -24,35 +24,28 @@ module.exports = (neutrino, opts = {}) => {
       }
     }
   }, opts);
-
-  Object.assign(options, {
-    test: (input) => {
-      const isCssModule = options.modulesTest.test(input);
-      const isRegularCss = options.test.test(input);
-
-      if (options.modules !== false && isCssModule) {
-        return false;
-      }
-
-      return isRegularCss;
-    }
-  });
-
-  const rules = [options];
+  const rules = [];
 
   if (options.modules) {
-    rules.push(merge(options, {
-      test: options.modulesTest,
-      ruleId: `${options.ruleId}${options.modulesSuffix}`,
-      styleUseId: `${options.styleUseId}${options.modulesSuffix}`,
-      cssUseId: `${options.cssUseId}${options.modulesSuffix}`,
-      hotUseId: `${options.hotUseId}${options.modulesSuffix}`,
-      extractId: `${options.extractId}${options.modulesSuffix}`,
-      css: {
-        modules: options.modules,
-        importLoaders: 1
-      }
-    }));
+    rules.push(
+      merge(options, {
+        exclude: options.modulesTest
+      }),
+      merge(options, {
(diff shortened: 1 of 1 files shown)


timkelty:
Thinking/hoping this should be non-breaking, as previously weren't doing anything with `.exclude`. This just makes it possible to override without recreating the test function.

packages/style-loader/index.js
```diff
@@ -36,22 +24,28 @@ module.exports = (neutrino, opts = {}) => {
...
-      cssUseId: `${options.cssUseId}${options.modulesSuffix}`,
-      hotUseId: `${options.hotUseId}${options.modulesSuffix}`,
-      extractId: `${options.extractId}${options.modulesSuffix}`,
-      css: {
-        modules: options.modules,
-        importLoaders: 1
-      }
-    }));
+    rules.push(
+      merge(options, {
```
eliperelman: This will add a whole new rule, making it 3 rules long instead of 2. Is that right?

eliperelman: Oh nevermind, I see we aren't adding `options` to the original rules array now.

eliperelman: Could this be part of the original object for options?

```js
extract: opts.modules && opts.modulesTest || neutrino.regexFromExtensions(['module.css'])
```

May not read the best though. Thoughts?

timkelty: You mean `modulesTest`, not `extract`, right?

timkelty: oh no - you prob mean `exclude`

timkelty: I don't think that would work, as we _only_ want to exclude the modulesTest for the non-modules rule.

timkelty:
@eliperelman ok - I moved things around a bit based on your suggestion.
I think its a bit more clear now.

packages/style-loader/index.js
```diff
@@ -40,19 +31,22 @@ module.exports = (neutrino, opts = {}) => {
...
+        test: options.modulesTest,
+        exclude: options.modulesExclude,
+        ruleId: `${options.ruleId}${options.modulesSuffix}`,
+        styleUseId: `${options.styleUseId}${options.modulesSuffix}`,
+        cssUseId: `${options.cssUseId}${options.modulesSuffix}`,
+        hotUseId: `${options.hotUseId}${options.modulesSuffix}`,
+        extractId: `${options.extractId}${options.modulesSuffix}`,
+        css: {
+          modules: options.modules,
+          importLoaders: 1
```
eirikurn: I know you did not change this line, but is it correct? See `importLoaders` for the css rule above.

Why do the rules have different importLoaders values @eliperelman?

timkelty: @eirikurn when using css-loader, `importLoaders` should equal the number of additional loaders (eg postcss, sass, etc) to be added before css-loader.

However, I'm _not_ sure why we're also incrementing importLoaders for `modules: true`. @eliperelman did that bit.

timkelty: It does appear they are doing this here: https://github.com/css-modules/webpack-demo/blob/master/webpack.config.js#L21, though I can't find any docs on whether importLoaders should be incremented when using css modules.

timkelty: I just did some tests and didn't seem to see a difference if the modules rule had importLoaders incremented or not…curious to see if @eliperelman knows where this came from.

timkelty:
@eirikurn @eliperelman ok, I removed the incrementing of `importLoaders` for the css modules rules, as don't believe it should have been there to begin with.

importLoaders will now just default to the length of the passed `loaders`, and will also be able to be manually set by the user via `options.css.importLoaders`

timkelty:
Another thing fixed in this PR - css modules weren't actually working!

I thought they were - because they were in the extracted modules, but importing them from js just gave you an empty obj :(

Hopefully we can get this merged and released asap!

timkelty:
👏

PR #614, opened 2017-12-13
Warn about the evils of mutating `neutrino.options`.

timkelty:


docs/customization/README.md
```diff
@@ -159,6 +159,47 @@ module.exports = {
...
+
+```js
+// Bad: Using function format, overriding `neutrino.options` properites.
+// Paths will not be relative to `neutrino.options.root` as expected.
+module.exports = neutrino => {
+  Object.assign(neutrino.options, {
+    source: 'lib'
+  });
+
+  neutrino.options.output = 'dist';
```
eliperelman: This line here is actually a supported way of changing some of the options. Each of these is a getter/setter, so setting the value to a relative path let's you use a getter for the absolute path. Same for `mains`, you can get/set them just fine.

The problem is using `Object.assign` or merge on them, as this *overwrites* the property definition.

I'd change this to highlight this technique as "good".

timkelty: Ah, wondered about that - I'll change it.

Issue #612, opened 2017-12-13
`lint` command, eslint and additional linters


timkelty:
Looking for advice on the best way to add a linters for additional resource types, (e.g. `stylelint`, `htmllint`).

`@neutrinojs/eslint` registers tasks: `lint` and `eslintrc`.

When factoring in linters for additional resource types, this can get confusing.

Should the `lint` command be renamed to `eslint`?
Or - is there some way to register additional calls to the `lint` command?

Ideally `lint` would call `eslint`, `stylelint`, or any other middlwares that have somehow registered lint commands.

eliperelman:
You could re-declare the `lint` command, and call its previous output. We are considering doing this for the WebExtensions linting.

```js
const { lint } = neutrino.commands;

neutrino.register('lint', (...args) => {
  lint(...args);
  // do your thing
});
```

timkelty:
👍  I'll try it!

timkelty:
So I tried your suggestion and it seemed things were getting called recursively (getting Maximum call stack size exceeded).

eliperelman:
Could you share the code for it so I can try it out?

timkelty:
@eliperelman here's an example: https://github.com/timkelty/neutrino-dev/blob/stylelint-lint-command/packages/stylelint/index.js#L38-L49

timkelty:
I think it would be nice to figure this out for 9.x.

It seems to me that the `lint` command could be renamed to `eslint`, and then provide a way for multiple middlewares to hook into a shared `lint` command?

Thoughts @mozilla-neutrino/core-contributors?

edmorley:
@timkelty Do you mean having a command category (`lint`) and type (eg `eslint`) whereby either just the individual type can be run, or the whole category?

timkelty:
@edmorley that's what I was getting at, and it might be nice, but I'm not sure how helpful it would be.

I think the most immediate clarification would be to rename the `lint` command to `eslint`.

edmorley:
After #852 this issue no longer exists, since there is no `lint` command. Instead individual `eslintrc` and `stylelintrc` output handlers are registered by their respective presets.

PR #615, opened 2017-12-13
stylelint middleware

timkelty:
Most of what's here is adapted from https://github.com/barraponto/neutrino-middleware-stylelint 👏 

Curious if @mozilla-neutrino/core-contributors think this should be in core, or just maintained by community. If the consensus is the latter, I'm happy to PR @barraponto's package.

`stylelint` felt like a commonly used enough tool that it could be a good case for a core middleware.

Related: https://github.com/mozilla-neutrino/neutrino-dev/issues/612

### DIFF
--- a/packages/stylelint/README.md
+++ b/packages/stylelint/README.md
@@ -1,6 +1,6 @@
-# Neutrino ESLint Middleware
+# Neutrino stylelint Middleware
 
-`@neutrinojs/eslint` is Neutrino middleware for linting source code using ESLint and eslint-loader.
+`@neutrinojs/stylelint` is Neutrino middleware for linting styles using stylelint.
 
 [![NPM version][npm-image]][npm-url]
 [![NPM downloads][npm-downloads]][npm-url]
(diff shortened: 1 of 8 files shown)


packages/stylelint/test/middleware_test.js
```diff
@@ -0,0 +1,37 @@
...
+  const api = Neutrino();
+
+  api.use(mw(), options);
+
+  t.notThrows(() => api.config.toConfig());
+});
+
+test('exposes stylelintrc config', t => {
+  t.is(typeof Neutrino().use(mw()).call('stylelintrc'), 'object');
+});
```
timkelty: Hmmm - This last test causes tests to fail: https://travis-ci.org/mozilla-neutrino/neutrino-dev/jobs/316091388#L920-L932 not sure why…seems identical to the eslintrc test…

eliperelman: Run the tests locally and see the output of `Neutrino().use(mw()).call('stylelintrc')`.

timkelty: Oh derp - it's b/c stylelint throws err if you don't pass it anything.
Fixed.

barraponto:
YAY let's move this into core!

timkelty:
Is anyone else using Atom/AtomLinter with [linter-stylelint](https://atom.io/packages/linter-stylelint)?

If I try to use neutrinorc.js for my stylint config, I get this error: https://github.com/atom/atom/issues/15093

.stylelintrc.js
```js
const {Neutrino} = require('neutrino');
const {resolve} = require('path');

module.exports = Neutrino()
  .use(resolve(__dirname, '.neutrinorc.js'))
  .call('stylelintrc');
```

- `neutrino stylelintrc` outputs the correct thing from the cli.
- Sublime Text 3/SublimeLinter works fine.

packages/stylelint/index.js
```diff
@@ -0,0 +1,39 @@
...
+const StylelintPlugin = require('stylelint-webpack-plugin');
+const { lint } = require('stylelint');
+
+module.exports = (neutrino, opts = {}) => {
+  const options = merge({
+    pluginId: 'stylelint',
+    files: '**/*.+(css|scss|sass|less)',
+    context: neutrino.options.source,
+    failOnError: neutrino.options.command !== 'start',
+    quiet: neutrino.options.command === 'start'
```
eliperelman: We also have `neutrino.options.quiet`, would that be useful here?

eliperelman:
So far I think this is pretty cool! Could you post some screenshots showing this in action?

timkelty:
Would also like to tie this into the `lint` command. Haven't had success so far: https://github.com/mozilla-neutrino/neutrino-dev/issues/612

packages/stylelint/README.md
```diff
@@ -0,0 +1,65 @@
+# Neutrino stylelint Middleware
```
edmorley: Need a `docs/` copy of this too.. :-)

edmorley: Also a reference from `docs/SUMMARY.md`

timkelty: Yup - I'll wait till things gel a bit, then add it.

timkelty:
Here's some screenshots: 
`neutrino stylelint` command:
screen shot 2017-12-19 at 8 49 11 pm

`neutrino start`:
screen shot 2017-12-19 at 9 20 57 pm

packages/stylelint/index.js
```diff
@@ -0,0 +1,41 @@
...
+    .get('args')[0].config || {};
+
+  neutrino.register(
+    'stylelint',
+    () => {
+      const { fix = false } = neutrino.options.args;
+      const files = join(options.context, options.files);
+
+      return lint(merge(options, { fix, files }))
+        .then(result => Promise[result.errored ? 'reject' : 'resolve'](result.output));
```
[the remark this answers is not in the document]
timkelty: nitpicky is good :)

packages/stylelint/README.md
timkelty: Added!

packages/stylelint/index.js
```diff
@@ -0,0 +1,41 @@
...
+const StylelintPlugin = require('stylelint-webpack-plugin');
+const { lint, formatters } = require('stylelint');
+
+module.exports = (neutrino, opts = {}) => {
+  const options = merge({
+    pluginId: 'stylelint',
+    files: '**/*.+(css|scss|sass|less)',
+    context: neutrino.options.source,
+    failOnError: neutrino.options.command !== 'start',
+    quiet: neutrino.options.command === 'start',
```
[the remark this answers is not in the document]
timkelty: This ensures that the actual error is output, when `failOnError` is enabled.

E.g. with settings the way they currently are, you will get:
```
⠼ Building project
src/css/index.scss
 8:1  ✖  Unexpected empty line before at-rule   at-rule-empty-line-before


✖ Building project failed
Error: Failed because of a stylelint error.

    at linterSuccess (/Users/timkelty/Code/neutrino-dev/node_modules/stylelint-webpack-plugin/lib/run-compilation.js:34:14)
    at <anonymous>
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

But if we don't set `quiet` (defaulting to true), you won't get the actual error:
```
✖ Building project failed
Error: Failed because of a stylelint error.

    at linterSuccess (/Users/timkelty/Code/neutrino-dev/node_modules/stylelint-webpack-plugin/lib/run-compilation.js:34:14)
    at <anonymous>
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

The other option is to not set either `failOnError` _or_ `quiet`, since by default it emits errors to webpack, so the build will fail anyway.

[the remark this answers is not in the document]
timkelty: > The other option is to not set either `failOnError` or `quiet`, since by default it emits errors to webpack, so the build will fail anyway.

I opted to go this route.

timkelty:
Ok, I think all review notes should be addressed, and I added some documentation for the cli commands.

timkelty:
🍾

Issue #608, opened 2017-12-12
A created project comes with linting issues


armenzg:
I chose to create a React project with no testing and with the airbnb configuration (created with `npx @neutrinojs/create-project ci-metrics`.
After that I would run `yarn start` and I get linting errors.

```
armenzg@armenzg-mbp ci-metrics$ yarn start
yarn run v1.3.2
$ neutrino start
✔ Development server running on: http://localhost:5000
✔ Build completed

ERROR in ./src/index.jsx
error: Expected a newline after '(' (function-paren-newline) at src/index.jsx:6:26:
  4 |
  5 | const root = document.getElementById('root');
> 6 | const load = () => render((
    |                          ^
  7 |   <AppContainer>
  8 |     <App />
  9 |   </AppContainer>


error: Expected a newline before ')' (function-paren-newline) at src/index.jsx:10:8:
   8 |     <App />
   9 |   </AppContainer>
> 10 | ), root);
     |        ^
  11 |
  12 | // This is needed for Hot Module Replacement
  13 | if (module.hot) {


2 errors found.
2 errors potentially fixable with the `--fix` option.
 @ multi (webpack)-dev-server/client?http://localhost:5000 (webpack)/hot/dev-server.js ./node_modules/react-hot-loader/patch.js ./src/index

ERROR in ./src/App.jsx
error: Missing trailing comma (comma-dangle) at src/App.jsx:6:23:
  4 | export default class App extends Component {
  5 |   state = {
> 6 |     name: 'ci-metrics'
    |                       ^
  7 |   };
  8 |
  9 |   render() {


1 error found.
1 error potentially fixable with the `--fix` option.
 @ ./src/index.jsx 4:0-24 15:2-34 15:29-33
 @ multi (webpack)-dev-server/client?http://localhost:5000 (webpack)/hot/dev-server.js ./node_modules/react-hot-loader/patch.js ./src/index
```

eliperelman:
Hmm, this is another thing I don't replicate. I wonder if you have some global install of eslint or something that is conflicting with Neutrino. Here is the output of me running the same command:

```
❯ npx @neutrinojs/create-project ci-metrics
                          _          _
      _ __    ___  _   _ | |_  _ __ (_) _ __    ___
     | '_ \  / _ \| | | || __|| '__|| || '_ \  / _ \
     | | | ||  __/| |_| || |_ | |   | || | | || (_) |
[...]

PR #616, opened 2017-12-13
Add debugging to CNP using the debug flag

helfi92:
There is no way to debug the CNP when something goes wrong. This PR will push everything to `stdout` when the debug flag is set. We saw it happen in #608.


helfi92:
Will add docs in a different PR.

PR #610, opened 2017-12-13
Add create-project example to the quickstart

helfi92:
Reference: https://github.com/mozilla-neutrino/neutrino-dev/issues/606

PR #621, opened 2017-12-15
Nest create-new-project under installation and link to it from index

timkelty:
Thought it made sense for newcomers to see create-project linked to up-front.

Related: https://github.com/mozilla-neutrino/neutrino-dev/pull/610


timkelty:
Just checked - that was the only place I could find.

eliperelman:
Ah I see, I was getting confused with #610. With that now merged, I think the links in there now need to be updated. Sorry!

timkelty:
Ok - rebased off master and applied these changes. Also stole some of the language used in #610 so they match up.

timkelty:
On a side note - I noticed there is no `docs/packages/create-project`… is that intentional?

eliperelman:
@timkelty yes intentional. `packages` are for middleware, while `neutrino` and `create-project` are split out.

PR #624, opened 2017-12-15
Optionally pass resolve.modules to web, node presets

timkelty:
This was the last bit of my standard neutrino setup that I couldn't do w/o tapping into the neutrino api, so I thought it would be nice to include.

If we wanted the option name more explicit, perhaps `resolveModules`?

Related: https://github.com/mozilla-neutrino/neutrino-dev/pull/541

### DIFF
--- a/packages/node/README.md
+++ b/packages/node/README.md
@@ -277,7 +277,12 @@ module.exports = {
             }
           }]
         ]
-      }
+      },
+
+      // Example: add paths to `resolve.modules`
+      modules: [
+        neutrino.options.source,
+      ]
     }]
   ]
 };
--- a/packages/node/README.md
+++ b/packages/node/README.md
@@ -181,7 +181,7 @@ if (module.hot) {
 }
 
 createServer((req, res) => {
-  res.end(app('example'));  
+  res.end(app('example'));
 }).listen(/* */);
 ```
 
@@ -196,7 +196,7 @@ if (module.hot) {
 }
 
 createServer((req, res) => {
-  res.end(app('example'));  
+  res.end(app('example'));
 }).listen(/* */);
 ```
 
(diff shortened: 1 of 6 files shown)


packages/node/README.md
```diff
@@ -277,7 +277,12 @@ module.exports = {
             }
           }]
         ]
-      }
+      },
+
+      // Example: add paths to `resolve.modules`
+      modules: [
+        neutrino.options.source,
```
edmorley: Could you update the `docs/` copy of the docs too? :-)

timkelty: Done!

packages/node/README.md
```diff
@@ -181,7 +181,7 @@ if (module.hot) {
 }
 
 createServer((req, res) => {
-  res.end(app('example'));
```
edmorley: Could you make this whitespace change to the `docs/` version too? :-)

timkelty: Done.

timkelty:
Holler if theres anything more I can do to get this merged! 😘

eliperelman:
So, I'm not sure there's much value to to this other than trying to avoid having to do custom configuration to achieve this. These aren't really settings that control the middleware at all, we are just creating random options for configuration here.

I think I would prefer to not take this and just allow overriding the configuration normally:

```js
module.exports = {
  use: [
    '@neutrinojs/react',
    (neutrino) => neutrino.config.resolve.modules.add(neutrino.options.source)
  ]
};
```

eliperelman:
I may be persuaded if this was a boolean option, like `resolveSourceModules` or something:

```js
module.exports = {
  use: [
    ['@neutrinojs/react', {
      resolveSourceModules: true
    }]
  ]
};
```

But we can't turn this on by default.

timkelty:
@eliperelman nah, I think you're right.

PR #628, opened 2017-12-17
Add supported versions of npm, Yarn and Node.js

### LINKED ISSUE #625: Add engines check to all packages
Right now the documentation lists out the requirements of using Neutrino or middleware:

> - Node.js v6.10+
> - Yarn or npm client
> - Neutrino v8

Now, in the root workspace package.json:

```json
"engines": {
  "node": ">=6.10.0",
  "yarn": "^1.2.1"
}
```

Good for development, not very helpful for our users. Let's update the docs to reflect our current support, and codify this in every monorepo package.json's `engines`:

Node.js v6 LTS, v8, v9
Yarn v1.2.1+, or npm v5.5 (I think? @helfi92?)

helfi92:
Fixes #625.


package.json
```diff
@@ -7,8 +7,9 @@
   "bugs": "https://github.com/mozilla-neutrino/neutrino-dev/issues",
   "repository": "mozilla-neutrino/neutrino-dev",
   "engines": {
-    "node": ">=6.10.0",
-    "yarn": "^1.2.1"
+    "node": ">=6.9.0",
```
eliperelman: The `node` engine needs to be more complex to handle the actual support. I think you can use `||` to delineate these. Also, since `7` is excluded, we probably don't want `>=` except in the `9` case.

```json
"node": "^6.9.0 || ^8 || >=9"
```

helfi92: ++

edmorley: I'm pretty sure the `^8` can be written as `8` (the latter being what I've seen used elsewhere).

eliperelman: I haven't seen that, but I think sticking with the `^` makes the semver relationship clearer.

Issue #635, opened 2017-12-19
Consider using Jest instead of Ava


okonet:
Ava is awesome but I think Jest offers so much more out of the box. Particularly I wanted to write a test for a fix of #632 but this seems not possible without pulling in at least 2 additional dependencies for 1. mocking and stubbing (sinon?) and 2. for hijacking the node `require` in order to mock the Jest APIs. With Jest that would not require any additional dependencies.

Thoughts?

eliperelman:
I am totally open to using whatever improves our testing situation. I think at the time I used AVA because it was fast, and Jest wasn't really a thing I knew could be used outside React. But yeah, I'm totally open to it.

eliperelman:
As an aside, AVA now supports snapshot testing.

PR #636, opened 2017-12-19
fix(Jest): Fix Jest middleware always exit with zero code

### LINKED ISSUE #632: neutrino test always exits with 0
When using Jest preset, even if my test suit fails the exit code is always 0 which causes CI to always pass. 

neutrino version 7.2.1+ (also tested on 7.4.0)

It works as expected on v6.2.0

okonet:
Closes #632

Since this regression was introduced in v7, I think it should be released as a patch release in v7 and v8 accordingly.

After spending ~30 mins trying to write tests with Ava I gave up and offering my help with #635

### DIFF
--- a/packages/jest/src/index.js
+++ b/packages/jest/src/index.js
@@ -118,10 +118,9 @@ module.exports = (neutrino, opts = {}) => {
 
       writeFileSync(configFile, `${JSON.stringify(options, null, 2)}\n`);
 
-      jest.runCLI(cliOptions, options.roots || [options.rootDir], result =>
-        (result.numFailedTests || result.numFailedTestSuites ?
-          reject() :
-          resolve()));
+      jest.runCLI(cliOptions, options.roots || [options.rootDir])
+        .then(results => results.success ? resolve() : reject())
+        .catch(error => reject(error));
     });
   });
 };


eliperelman:
Strange, the test for this failed in travis, even though jest says the test passed.

okonet:
Very strange, indeed. I've looked at Jest's results Object and it's says it's all good. Not sure where the 1 code is coming from. For some reason I also see the error code printed twice:

```

0
  ✔ cli_test › @neutrinojs/react (10.1s)
1
0
  ✖ cli_test › @neutrinojs/react + @neutrinojs/jest Failed to test project:

```

when adding a `console.log` here https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/create-project/test/cli_test.js#L28

okonet:
My assumption is that it was passing before since the Jest would never reject.

packages/jest/src/index.js
```diff
@@ -118,10 +118,9 @@ module.exports = (neutrino, opts = {}) => {
 
       writeFileSync(configFile, `${JSON.stringify(options, null, 2)}\n`);
 
-      jest.runCLI(cliOptions, options.roots || [options.rootDir], result =>
-        (result.numFailedTests || result.numFailedTestSuites ?
-          reject() :
-          resolve()));
+      jest.runCLI(cliOptions, options.roots || [options.rootDir])
```
eliperelman: I was just thinking, since we are no longer using a callback here, we can remove the `return new Promise` on L99 and instead just return the `runCLI` call, right?

```js
return jest.runCLI(cliOptions, options.roots || [options.rootDir]);
```

Do we even then need to capture `results.success`, or will jest reject if `success` is false?

okonet: It won't reject as far as I can say by looking at their code. I think their promise is for some unexpected inside a `try ... catch`. That's why they have the `success` property.

eliperelman: OK, so we will need to manually throw or reject or whatever. Gotcha.

okonet: Yeah. The implementation I've submitted in this PR works in my repository. I dunno why are tests failing, though.

eliperelman: OK, I'll see if I can clone locally and replicate.

eliperelman: If you make this change, I will get it out in a patch release today. :)

okonet: 🤔 why did it work for me? Did they break the API again or I’m just stupid? I’ll test the code again with different Jest versions to check that.

eliperelman: So, the just tests pass, the only thing was we were using the results properties from jest to determine pass or fail incorrectly, hence why it was failing our tests.

okonet: No, I mean I tried this fix on my project and it fixed the issue. That's why I submitted the fix in the first place.

eliperelman: Hmm not sure. You did fix the fact that it should be a promise now, so maybe that was part of it.

okonet:
You decided to go with it? I’m sorry I still didn’t have time to investigate it

PR #640, opened 2017-12-19
Fix issue where ES6 won't compile with Jest and Neutrino v8

### LINKED ISSUE #638: ES6 won't compile when using Jest preset with v8
I'm trying to upgrade to v8 from v7 and following the guide. But after I switched to new presets I'm getting lots of these errors:

```
 FAIL  src/content/app/components/ui/DataGrid/SortedDataGrid.test.jsx
  ● Test suite failed to run

    /Users/okonet/Projects/feedly/client/src/content/app/components/ui/DataGrid/SortedDataGrid.test.jsx:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import React from 'react';
                                                                                             ^^^^^^
    
    SyntaxError: Unexpected token import
      
      at ScriptTransformer._transformAndBuildScript (node_modules/jest-runtime/build/script_transformer.js:305:17)
          at Generator.next (<anonymous>)
          at new Promise (<anonymous>)
```

No matter what I try I can't get it working again. Looks like the preset doesn't pick the `compile` part from `react` preset.

helfi92:
Fixes #638.

### DIFF
--- a/packages/jest/src/index.js
+++ b/packages/jest/src/index.js
@@ -21,9 +21,9 @@ function getFinalPath(path) {
 }
 
 function normalizeJestOptions(opts, neutrino, usingBabel) {
-  const mediaNames = `\\.(${mediaExtensions.join('|')})`;
-  const styleNames = `\\.(${['css', 'less', 'sass', 'scss'].join('|')})`;
-  const jsNames = neutrino.regexFromExtensions(['js', 'jsx']);
+  const mediaNames = `\\.(${mediaExtensions.join('|')})$`;
+  const styleNames = `\\.(${['css', 'less', 'sass', 'scss'].join('|')})$`;
+  const jsNames = `\\.(${['js', 'jsx'].join('|')})$`;
   const aliases = neutrino.config.resolve.alias.entries() || {};
   const moduleNames = Object
     .keys(aliases)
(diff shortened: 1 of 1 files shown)


packages/jest/src/index.js
```diff
@@ -21,9 +21,9 @@ function getFinalPath(path) {
 }
 
 function normalizeJestOptions(opts, neutrino, usingBabel) {
-  const mediaNames = `\\.(${mediaExtensions.join('|')})`;
-  const styleNames = `\\.(${['css', 'less', 'sass', 'scss'].join('|')})`;
-  const jsNames = neutrino.regexFromExtensions(['js', 'jsx']);
+  const mediaNames = `\\.(${mediaExtensions.join('|')})$`;
+  const styleNames = `\\.(${['css', 'less', 'sass', 'scss'].join('|')})$`;
+  const jsNames = `\\.(${['js', 'jsx'].join('|')})$`;
```
eliperelman: We still want to use the stored Neutrino extensions, so probably something like:

```js
`\\.(${[...neutrino.options.extensions].join('|')})$`;
```

edmorley: I think we should also add a comment explaining why we're not using `regexFromExtensions()` :-)

hshoff:
Applied this patch to my local install and it worked great. Thanks for the quick fix. Excited to start using neutrino for [vx](http://github.com/hshoff/vx)

eliperelman:
Looks great!

PR #673, opened 2018-01-15
Refactor imagemin plugin to fix minification when used with libraries

### LINKED ISSUE #670: `image-minify` plugin breaks json-loader
Here is a [project sample](https://mega.nz/#!KUMH1IpY!6sfZUI9aNncJ5ScZOrR8yec7G9rd2Cw30BUP6VP2x7c). Steps to reproduce a bug.

1. Install dependencies
2. Run `npm run build`
3. CLI throws an error 

```
✖ Building project failed
./src/config.json
Module parse failed: Unexpected token (2:13)
| {
|     "baseUrl": "default"
| }
```

4. Comment `minify` plugin in `.neutrinorc.js`

```
 // use(minify)
```

5. Run `npm run build`
6. Build is successful. You can check this opening `index.html`

**Expected result:** Build should not fail when minification is enabled
**Actual result:** `@neutrinojs/minify` breaks Webpack json-loader

eliperelman:
This fixes #670. Turns out the problem is not babel-minify, but rather some kind of interference between the imagemin plugin we were using and creating libraries.

I switched to a different imagemin plugin with a nice API, which let us cut out some code and direct dependencies. @timkelty what do you think?

### DIFF
--- a/packages/image-minify/index.js
+++ b/packages/image-minify/index.js
@@ -7,9 +7,9 @@ module.exports = (neutrino, opts = {}) => {
   const options = merge({
     imagemin: {
       plugins: [],
-      optipng: {},
-      gifsicle: {},
-      jpegtran: {},
+      optipng: null,
+      gifsicle: null,
+      jpegtran: null,
       svgo: {},
       pngquant: {},
       webp: {}
--- a/packages/image-minify/index.js
+++ b/packages/image-minify/index.js
@@ -8,7 +8,7 @@ module.exports = (neutrino, opts = {}) => {
     imagemin: {
       plugins: [],
       optipng: null,
-      gifsicle: null,
+      gifsicle: {},
       jpegtran: null,
       svgo: {},
       pngquant: {},
--- a/packages/image-minify/index.js
+++ b/packages/image-minify/index.js
@@ -1,6 +1,7 @@
 const merge = require('deepmerge');
 const ImageminWebpackPlugin = require('imagemin-webpack-plugin').default;
 const webp = require('imagemin-webp');
+const mozjpeg = require('imagemin-mozjpeg');
 
 module.exports = (neutrino, opts = {}) => {
   const options = merge({
@@ -10,7 +11,7 @@ module.exports = (neutrino, opts = {}) => {
       gifsicle: {},
       jpegtran: {},
       svgo: {},
-      pngquant: null,
+      pngquant: {},
       webp: {}
     },
     pluginId: 'imagemin'
@@ -18,6 +19,10 @@ module.exports = (neutrino, opts = {}) => {
 
   if (options.webp) {
     options.imagemin.plugins.push(webp(options.webp));
+  }
+
+  if (options.mozjpeg) {
+    options.imagemin.plugins.push(mozjpeg(options.mozjpeg));
   }
 
   neutrino.config
--- a/packages/image-minify/index.js
+++ b/packages/image-minify/index.js
@@ -1,49 +1,26 @@
 const merge = require('deepmerge');
-const ImageminWebpackPlugin = require('imagemin-webpack/ImageminWebpackPlugin');
-const gifsicle = require('imagemin-gifsicle');
-const svgo = require('imagemin-svgo');
-const pngquant = require('imagemin-pngquant');
-const mozjpeg = require('imagemin-mozjpeg');
+const ImageminWebpackPlugin = require('imagemin-webpack-plugin').default;
 const webp = require('imagemin-webp');
-
-const imageminLoader = require.resolve('imagemin-webpack/imagemin-loader');
 
 module.exports = (neutrino, opts = {}) => {
   const options = merge({
     imagemin: {
-      plugins: [
-        gifsicle(),
-        svgo(),
-        pngquant(),
-        mozjpeg(),
-        webp()
-      ]
+      plugins: [],
+      optipng: {},
+      gifsicle: {},
+      jpegtran: {},
+      svgo: {},
+      pngquant: null,
+      webp: {}
     },
-    plugin: {
-      name: '[path][name].[ext]'
-    },
-    pluginId: 'imagemin',
-    useId: 'imagemin',
-    rules: ['svg', 'img']
+    pluginId: 'imagemin'
   }, opts);
 
-  const test = options.rules.map((ruleId) => {
-    neutrino.config.module
-      .rule(ruleId)
-      .use(options.useId)
-        .loader(imageminLoader)
-        .options(options.imagemin);
-
-    return neutrino.config.module.rule(ruleId).get('test');
-  })
-  .filter(Boolean);
-
-  options.plugin = merge({
-    imageminOptions: options.imagemin,
-    test
-  }, options.plugin);
+  if (options.webp) {
+    options.imagemin.plugins.push(webp(options.webp));
+  }
 
   neutrino.config
     .plugin(options.pluginId)
-    .use(ImageminWebpackPlugin, [options.plugin]);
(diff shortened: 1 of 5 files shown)


timkelty:
@eliperelman Will review asap
– from what I remember I chose `itgalaxy/imagemin-webpack` vs `Klathmon/imagemin-webpack-plugin` because it didn't load any of the imagmin-* deps by default, which can be slow to build. Other than that it seems the former is a plugin + loader, while the latter is just a plugin.

Not sure why you'd need both though…

packages/image-minify/index.js
```diff
@@ -1,49 +1,26 @@
...
-    plugin: {
-      name: '[path][name].[ext]'
+      plugins: [],
+      optipng: {},
+      gifsicle: {},
+      jpegtran: {},
+      svgo: {},
+      pngquant: null,
+      webp: {}
     },
```
timkelty: Is there a reason for setting these all to the same defaults they already have, or is this just for clarity? https://github.com/Klathmon/imagemin-webpack-plugin#optionsoptipng

packages/image-minify/index.js
```diff
@@ -1,49 +1,26 @@
...
-  })
-  .filter(Boolean);
-
-  options.plugin = merge({
-    imageminOptions: options.imagemin,
-    test
-  }, options.plugin);
+  if (options.webp) {
+    options.imagemin.plugins.push(webp(options.webp));
+  }
```
timkelty: I wonder how we should decide what to include by default…it would seem that `imagemin-webpack-plugin` enables "optimizers" by default, but not "compressors". Including `imagemin-webp` by default may be going against that.

However if we do, perhaps we should do the same with `imagemin-mozjpeg`. That is included in the current version of of this MW, but not present in this PR (nor is any jpeg compression).

timkelty (commented):
In general, this seems like a good idea, and `imagemin-webpack-plugin` seems like it has a nicer api.

I'm not sure why I thought we needed a loader _and_ a plugin to do this…seems like a plugin covers all cases.

My only concern is what imagemin plugins we enable by default, and why.  It seems like we have 2 options:

1. By default, enable only "optimizers" (as imagemin-webpack-plugin) seems to do, and leave it to the user to configure any actual compression.
2. By default, configure some sensible defaults, which in this case, might include adding `imagemin-mozjpeg`.

If we go with 1., we might justify enabling it by default in `@neutrinojs/web`, as we are doing with other minify MWs.

eliperelman:
> enable only "optimizers" (as imagemin-webpack-plugin) seems to do, and leave it to the user to configure any actual compression

What is the difference between optimizing and compression here? What were we doing before this patch?

> configure some sensible defaults, which in this case, might include adding `imagemin-mozjpeg`

I think `jpegtran` takes the place of `imagemine-mozjpeg`, is that correct?

I tried to enable a plugin for every type of file we had before to keep it consistent. Please let me know if it doesn't appear to be the case.

timkelty:
> What is the difference between optimizing and compression here? What were we doing before this patch?

I'd consider optimizing reducing filesize without actually degrading image quality. I'm assuming that is the aim of `imagemin-webpack-plugin`'s defaults.

> I think jpegtran takes the place of imagemine-mozjpeg, is that correct?

No. Jpegtran, only performs lossless optimizations. To actually compress a jpg, you'd want `imagemin-mozjpeg` as well.

Before this, we also were including pngquant and mozjpeg, which are lossy compressors.

### To summarize

I think we should either:
1. stick with lossless optimizers, i.e. `imagemin-webpack-plugin`'s defaults
2. or add `pngquant`, `webp`, and `mozjpeg` as sensible defaults

1 means less assumptions and less to manage, 2 would mean potentially less boilerplate for users.

eliperelman:
@timkelty How about option 2, with an option to disable lossy compression?

timkelty:
> @timkelty How about option 2, with an option to disable lossy compression?

That could work, though I'd be a little concerned over confusion as to what is classified as lossy/lossless or not. Personally, I might prefer to remove that layer of indirection and just include the lossy ones by default and instruct users to pass `null` to mozjpeg, webp, pngquant or any other they want to disable.

Another option would be to require imagemin-mozjpeg and imagemin-webp and expose them as top level options like the others, but disable them by default (as imagemin-webpack-plugin [does with pngquant](https://github.com/Klathmon/imagemin-webpack-plugin#optionspngquant))

edmorley:
One thing I noticed with my Neutrino 8 upgrade WIP, is that even though image minification defaults to off - the addition of the image related packages to the web preset means I quite often get costly (in terms of time) yarn installs (made worse by yarnpkg/yarn#932, which turns what should be <2s close-to-no-op reinstalls into 20-30s installs), as well as having to pay the download cost for travis cache/docker image (which will still be the case with the yarn bug fixed). It's quite a noticeable difference over Neutrino 7.

The options seem to be:
1. make the web preset not depend on `@neutrinojs/image-minify` and have those who wish to use it specify it in their package.json -- though this reduces convenience for them
2. for me to re-implement the web preset in our app, using the individual sub-presets directly -- which means a lot of duplication / boilerplate
3. just accept the longer yarn install / broken no-op installs and do nothing

Whilst it's not as convenient for those using the feature, what are both of your thoughts on (1)?

It seems like as Neutrino gets more mature (plus webpack gains more functionality turned on by default) - that more and more of the value-add Neutrino offers might end up being features that not everyone will need/use. As such, having a policy for what should be enabled/disabled by default, and whether disabled by default features should appear in the dependency tree seems useful.

timkelty:
> As such, having a policy for what should be enabled/disabled by default, and whether disabled by default features should appear in the dependency tree seems useful.

Agreed all around!

eliperelman:
I'm fine with blowing away the aggregation, but right now it's a breaking change. We need to get something in here still usable for v8.

timkelty:
@eliperelman to remain non-breaking for this version, we'd minimally need to enable pngquant and include mozjpeg, as they are used in the current version.

Current version has these defaults:
```        gifsicle(),
        svgo(),
        pngquant(),
        mozjpeg(),
        webp()```

edmorley:
What's the verdict here? It seems like the changes are currently breaking, or am I misunderstanding? :-)

timkelty:
@edmorley sorry for delay!

I think the idea is this _should_ be non-breaking,  as we're swapping out the webpack plugin, but the actual imagemin deps are the same. However, theres lots of moving pieces…might be safest just to wait for 9.x

edmorley:
I've opened PR #815 to detach the `@neutrinojs/image-minify` preset from the web and library presets.

There's also Renovate PR #801 to update the existing [imagemin-webpack](https://github.com/itgalaxy/imagemin-webpack) to v3. The configuration steps appear to be simpler now (the loader is added by the plugin, no need to do so ourselves):
https://github.com/itgalaxy/imagemin-webpack/blob/master/CHANGELOG.md

As for which plugin is better, it seems a close call:
* [imagemin-webpack-plugin](https://github.com/Klathmon/imagemin-webpack-plugin) (used by this PR) has more stars on github
* but [imagemin-webpack](https://github.com/itgalaxy/imagemin-webpack) says it supports assets imported via other types of loaders (eg url-loader) - [see here](https://github.com/itgalaxy/imagemin-webpack#why)

edmorley:
No preference :-)

edmorley:
This would be a breaking change, so needs to be finished and landed before Neutrino 9 is released, or else will have to wait until Neutrino 10.

timkelty:
@edmorley @eliperelman I've been swamped and out of the loop, but would love to this land for v9. I'll make sure I make time for a review today.

eliperelman:
Superseded by #1104.

PR #679, opened 2018-01-22
Fix Jest entry files search on Windows

constgen:
Jest can't find test (spec) files on Windows. The reason is backslashes in a test path. This PR changes slashes to Unix style in a file 'test' RegExp only. I hope there is no issue with other paths in Jest.

### DIFF
--- a/packages/jest/src/index.js
+++ b/packages/jest/src/index.js
@@ -51,10 +51,7 @@ function normalizeJestOptions(opts, neutrino, usingBabel) {
   const { extensions, source, tests, root } = neutrino.options
   const collectCoverageFrom = [join(relative(root, source), `**/*.{${extensions.join(',')}}`)]
 
-  const testRegex = join(
-    basename(tests),
-    `.*(_test|_spec|\\.test|\\.spec)\\.(${extensions.join('|')})$`
-  );
+  const testRegex = `${basename(tests)}/.*(_test|_spec|\\.test|\\.spec)\\.(${extensions.join('|')})$`;
 
   return merge.all([
     {


packages/jest/src/index.js
```diff
@@ -51,10 +51,7 @@ function normalizeJestOptions(opts, neutrino, usingBabel) {
   const { extensions, source, tests, root } = neutrino.options
   const collectCoverageFrom = [join(relative(root, source), `**/*.{${extensions.join(',')}}`)]
 
-  const testRegex = join(
-    basename(tests),
-    `.*(_test|_spec|\\.test|\\.spec)\\.(${extensions.join('|')})$`
-  );
+  const testRegex = `${basename(tests)}/.*(_test|_spec|\\.test|\\.spec)\\.(${extensions.join('|')})$`;
```
eliperelman: I don't understand the problem here. When we use `join`, Node.js should use the path separator for the OS you are running, and for Windows that would be a backslash. Can you explain why that is problematic?

constgen: In fact it doesn't work. I think it works the same as "test" in Webpack where we don't use OS specific slashes.

Here is a [sample](
https://mega.nz/#!jUkmgC5L!B7jWZ2kjfVq-SA-NTTFflTD9oaiLSKlJG9aALU7uvG4
) for testing

constgen: I want to admit that all Windows users can't run tests on the current version of Neutrino.

constgen: @eliperelman Please look at documentation https://facebook.github.io/jest/docs/en/configuration.html#testregex-string . All defaults use just a forward slash without any OS considerations. The same is valid for rest of config options. So probably we have multiple places where we shouldn't use `path.join()`

sgillespie: @eliperelman: The problem is that the windows path separator is also an escape sequence in regular expressions. For example, if my test directory is `src`, the generated regex would start with `src\\.*`, which now matches a literal `.` instead of the wildcard.

eliperelman:
Released in v8.1.2.

PR #691, opened 2018-01-29
Increase the displayed count of community packages on docs

### LINKED ISSUE #689: Custom presets not showing up in list
I have created two presets and as far as I can tell they satisfy all the [conditions listed](https://neutrino.js.org/presets/community-presets.html) but they are still not showing up in the list.

The presets I have made are:

- [neutrino-preset-flow](https://www.npmjs.com/package/neutrino-preset-flow) and
- [neutrino-preset-emotion](https://www.npmjs.com/package/neutrino-preset-emotion)

Is there an issue with the list not updating or is there something wrong with my presets?

Also note that it has been 5 days since I published those packages so I doubt its a batch processing issue.

Thanks in advance

eliperelman:
Should fix #689.

### DIFF
--- a/docs/middleware/community-middleware.md
+++ b/docs/middleware/community-middleware.md
@@ -22,4 +22,4 @@ In addition, your middleware should also:
 It should also have a relevant description. If you prefer, you can also add your middleware to the
 [neutrino-dev wiki](https://github.com/mozilla-neutrino/neutrino-dev/wiki/Community-Middleware).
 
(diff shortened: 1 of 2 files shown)


edmorley (approved):
Related: If we end up switching to the Gitbook beta I guess we'll either need to make this plugin compatible with gitbook 4+ or else decide to stop the embedding and just link to an NPM search page instead.

eliperelman:
Yeah, I'd almost rather link at this point since I don't want to tackle the pagination.

PR #694, opened 2018-01-29
Correctly merge eslint rules, fix hardcoded formatter

### LINKED ISSUE #665: merge.all concatenates eslint arrays
What version of Neutrino are you using?
`"neutrino": "^8.0.16"`

Are you trying to use any presets? If so, which ones, and what versions?
`"@neutrinojs/standardjs": "^8.0.16"`

Are you using the Yarn client or the npm client? What version?
`yarn --version`
`1.3.2`

What version of Node.js are you using?
`9.3.0`

What operating system are you using?
`OSX 10.13.1 High Sierra`

What did you do?
I tried to override babel/object-curly-spacing value. The hardcoded one is ["error", "always"]. I wanted to override it to [2, "never"]. I use the 2 to make the issue clearer as it is easy to see if it replaces the string "error".

What did you expect to happen?
To be able to change the value.

What actually happened, contrary to your expectations?
the value was concatenated by `merge.all()`
(4) `["error", "always", 2, "never"]`

Related issue:

I am also unable so far to find a way to override that setting using an .eslintrc file, with the eslint.useEslintrc flag set to true.

eliperelman:
Fixes #665.

### DIFF
--- a/packages/airbnb-base/index.js
+++ b/packages/airbnb-base/index.js
@@ -1,25 +1,22 @@
 const lint = require('@neutrinojs/eslint');
-const merge = require('deepmerge');
 
 module.exports = (neutrino, opts = {}) => {
-  neutrino.use(lint, merge.all([
-    {
-      eslint: {
-        baseConfig: {
-          extends: ['airbnb-base']
-        },
-        rules: {
-          // handled by babel rules
-          'new-cap': 'off',
-          // handled by babel rules
-          'object-curly-spacing': 'off',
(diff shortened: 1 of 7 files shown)


helfi92 (approved):
Tried the following scenario: https://github.com/mozilla-neutrino/neutrino-dev/issues/665#issuecomment-358679622 locally and seems to be fixed. Awesome work.

eliperelman:
Woohoo!

PR #703, opened 2018-02-02
Update the suggested .eslintrc.js shim so it can find .neutrinorc.js from any location

mdboom:
When emacs flycheck package runs `eslint`, it runs it from the same directory as the source file.  If that's the case, the suggested `.eslintrc.js` shim in this file will fail to find the `.neutrinorc.js` file.  This simple fix just looks for it in the same directory as the `.eslintrc.js` file.  This shouldn't (hopefully) have any negative consequences for any IDEs that run `eslint` from the project root that are currently working.

### DIFF
--- a/docs/packages/airbnb/README.md
+++ b/docs/packages/airbnb/README.md
@@ -5,7 +5,7 @@ config, following the [Airbnb styleguide](https://github.com/airbnb/javascript).
 
 [![NPM version][npm-image]][npm-url]
 [![NPM downloads][npm-downloads]][npm-url]
-[![Join the Neutrino community on Spectrum][spectrum-image]][spectrum-url] 
+[![Join the Neutrino community on Spectrum][spectrum-image]][spectrum-url]
 
 ## Features
 
@@ -258,8 +258,8 @@ const { Neutrino } = require('neutrino');
 // Specify middleware to Neutrino prior to calling eslintrc.
 // Even if using .neutrinorc.js, you must specify it when using
 // the API
-module.exports = Neutrino()
-  .use(__dirname + '/.neutrinorc.js')
+module.exports = Neutrino({ root: __dirname })
+  .use('.neutrinorc.js')
   .call('eslintrc');
 ```
 
--- a/docs/packages/airbnb/README.md
+++ b/docs/packages/airbnb/README.md
@@ -272,7 +272,7 @@ const { Neutrino } = require('neutrino');
 // Specify middleware to Neutrino prior to calling eslintrc.
 // You can choose to not use .neutrinorc.js as the middleware to
 // use if you prefer, specifying any middleware you wish.
-module.exports = Neutrino()
+module.exports = Neutrino({ root: __dirname })
   .use('@neutrinojs/airbnb', {
     eslint: {
       rules: { semi: 'off' }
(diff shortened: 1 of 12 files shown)


constgen:
Have the same issue in VSCode. But I tried this fix in VSCode and it produces another issue 
```
[Info  - 15:04:37] ESLint library loaded from: d:\Work\Projects_AtomSpace\Misha\frontend-js-snake\node_modules\eslint\lib\api.js
[Error - 15:04:43] Cannot read config file: d:\Work\Projects_AtomSpace\Misha\frontend-js-snake\.eslintrc.js Error: Cannot find module 'C:\Program Files (x86)\Microsoft VS Code\package.json'
```

I can confirm that code editors ESLint highlighting doesn't work in the current version

docs/packages/airbnb/README.md
```diff
@@ -259,7 +259,7 @@ const { Neutrino } = require('neutrino');
 // Even if using .neutrinorc.js, you must specify it when using
 // the API
 module.exports = Neutrino()
-  .use('.neutrinorc.js')
+  .use(__dirname + '/.neutrinorc.js')
```
eliperelman: I think the more correct solution here would be to pass the correct working directory to Neutrino so it can resolve anything else from the `root` dir:

```js
module.exports = Neutrino({ root: __dirname })
  .use('.neutrinorc.js')
  .call('eslintrc');
```

mdboom: Cool, I didn't know update that.  I'll update this PR.

constgen: I tried this. In VSCode I get again
```
[Error - 12:07:54] Cannot read config file: d:\Work\Projects_Fabrica\frontend-js-snake\.eslintrc.js Error: Cannot find module 'C:\Program Files (x86)\Microsoft VS Code\package.json'
```

constgen:
@mdboom did it really worked for you. It didn't for me. Can you test it in VSCode (with ESLint extension installed) on your side?

mdboom:
@constgen: It works for me in emacs.  I'm not a VSCode user.

mdboom:
For clarity: @constgen: Does it work for you before this change and not after, or in neither case?

constgen:
@mdboom It didn't work in both: before and after. I'm also looking for a solution.

constgen:
Have spent some time on research and debugging and found the origin of the issue on my side. It was a project setting of ESLint VS Code Extension that look like this.
```
"eslint.workingDirectories": [
    "./src",
    "./test"
]
```
I don't remember already why it was necessary. But removing this fixes the issue. This setting changes `cwd` of ESLint process that is started by an editor. And `process.cwd()` call in presets was incorrect. Of course `Neutrino({ root: __dirname })` is required and stays as it fixes another issue. 

So I confirm that it works on Windows now. Please review.

mdboom:
I've updated this to add additional instances of this change as listed above.

eliperelman:
I know it's nitpicky, but we could also drop the `.js` from the `.use` call since Node.js will pick it up by default. Thoughts?

edmorley:
I think I'd rather leave it, since it makes it clearer to the user what file is being referenced (people are perhaps used to knowing about the implicit file extension support for imports, but it's not entirely clear that the argument being passed takes advantage of node module resolution).

eliperelman:
OK, I think that's a fair point; we can leave it.

mdboom:
I've added what are hopefully the last instances remaining.

PR #709, opened 2018-02-16
Neutrino start should pay attention to NODE_ENV env variable

armenzg:
neutrino start always sets NODE_ENV to development regardless of process.env.NODE_ENV.
This fixes running neutrino as development in production environments like Heroku.

### DIFF
--- a/packages/neutrino/bin/start.js
+++ b/packages/neutrino/bin/start.js
@@ -9,7 +9,7 @@ module.exports = (middleware, args, cli) => {
     cli,
     middleware,
     args,
-    NODE_ENV: 'development',
+    NODE_ENV: process.env.NODE_ENV || 'development',
     commandHandler(config, neutrino) {
       if (!args.start) {
         spinner.enabled = global.interactive;


packages/neutrino/bin/start.js
```diff
@@ -9,7 +9,7 @@ module.exports = (middleware, args, cli) => {
     cli,
     middleware,
     args,
-    NODE_ENV: 'development',
+    NODE_ENV: process.env.NODE_ENV || 'development',
```
edmorley: These are defaults, which get overwritten here:
https://github.com/armenzg/neutrino-dev/blob/7b9e6a9d63a17bab8e3aa7b35cfb7c5ea96ff324/packages/neutrino/bin/base.js#L21

I believe you want NODE_ENV to propagate via `args.options` and so the fix would need to be in `packages/neutrino/src/api.js` instead.

eliperelman:
The API specifically does not care about the `NODE_ENV` so it can be specified by consumers. The CLI is responsible for providing the right `NODE_ENV` to the API.

eliperelman:
From what I remember, it used to get pretty hairy trying to respect the different ways environment variables could be used, so we stopped giving them special treatment. There's:

```sh
neutrino start --options.env.NODE_ENV production
```

armenzg:
I switched my app to using build instead of start.
Perhaps a message in the output telling users that it is running on dev mode regardless on NODE_ENV=production could be good. Perhaps not even letting them unless they use `--options.env.NODE_ENV production`

PR #714, opened 2018-02-19
Prefix neutrino commands with npm bin path

timkelty:
…In case user doesn't have `node_modules/bin` in their `PATH`.
Will also ensure local neutrino is used, in lieu of anything globally installed.

### DIFF
--- a/packages/create-project/commands/init/index.js
+++ b/packages/create-project/commands/init/index.js
@@ -67,20 +67,26 @@ module.exports = class Project extends Generator {
     return {};
   }
 
+  _createCliCommand(command) {
+    const installer = isYarn ? 'yarn' : 'npm';
+
+    return `$(${installer} bin)/${command}`;
+  }
+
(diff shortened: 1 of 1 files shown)


edmorley:
`yarn run` / `npm run` automatically add `node_modules/.bin` to `PATH`, when running `scripts`. Are there other use-cases where it won't be set appropriately?

The part of the PR that changes the "one time lint" feature does seem like it might help fix #608 - though instead of prefixing the bin directory, it might instead be better to make the one time lint run `$installer run lint`?

edmorley:
> The part of the PR that changes the "one time lint" feature does seem like it might help fix #608 - though instead of prefixing the bin directory, it might instead be better to make the one time lint run $installer run lint?

This part has been fixed in #817.

I'm not sure the rest of the PR is needed?

eliperelman:
Yeah, my opinion is we don't need this. I recommend closing.

PR #732, opened 2018-03-03
Fix typo in command to install npx

sylvhama:
Based on https://github.com/zkat/npx#install

### DIFF
--- a/docs/packages/library/README.md
+++ b/docs/packages/library/README.md
@@ -81,7 +81,7 @@ _Note: The `create` command is a shorthand that helps you do two things at once.
 #### npm/npx
 
 [`npx`](https://github.com/zkat/npx) comes pre-installed with `npm`. If you’re running an older version of `npm`, then
-`npm install -g npm` to update to the latest version.
+`npm install -g npx` to update to the latest version.
 
 ```
 ❯ npx @neutrinojs/create-project <directory-name>
--- a/docs/packages/node/README.md
+++ b/docs/packages/node/README.md
@@ -81,7 +81,7 @@ _Note: The `create` command is a shorthand that helps you do two things at once.
 #### npm/npx
 
 [`npx`](https://github.com/zkat/npx) comes pre-installed with `npm`. If you’re running an older version of `npm`, then
-`npm install -g npm` to update to the latest version.
+`npm install -g npx` to update to the latest version.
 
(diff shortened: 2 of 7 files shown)


edmorley:
I think this is as-intended? Newer versions of npm come with npx, so the docs are suggesting updating npm rather than manually installing npx?

sylvhama:
@edmorley on Windows 10 it will raise an error when running `npm install -g npm`:
```npm ERR! path C:\Program Files\nodejs\npm.cmd
npm ERR! code EEXIST
npm ERR! Refusing to delete C:\Program Files\nodejs\npm.cmd: is outside C:\Program Files\nodejs\node_modules\npm and not a link
npm ERR! File exists: C:\Program Files\nodejs\npm.cmd
npm ERR! Move it away, and try again.```

I've just used nvm for Windows to install node 9.8.0 which comes with npm v5.6.0. When switching to this version it says that npx is not  recognized.

edmorley:
> on Windows 10 it will raise an error when running npm install -g npm

This sounds like npm/npm#5050

> I've just used nvm for Windows to install node 9.8.0 which comes with npm v5.6.0. When switching to this version it says that npx is not recognized.

I can reproduce on Windows when using nvm-windows (but not with a standard nodejs install) - it looks like this is a known issue with nvm-windows specifically:
coreybutler/nvm-windows/issues/293

I'm torn here. I think generally we should encourage people to be on the latest version of `npm`, so ideally we'd keep that part of the docs, but given it causes an error on Windows, and that the explicit npx install is required when using nvm-windows, perhaps switching the instructions over for now is fine. 

@eliperelman, thoughts?

users of nvm-windows will need to install npx explicitly - so perhaps the docs could say to install

eliperelman:
I think the docs should promote `npx` primarily. To cover this situation, maybe we can add a note underneath it for nvm-windows users.

edmorley:
(And link to that nvm-windows issue)

sylvhama:
Should I close this PR and open a new one based on @edmorley advice?

eliperelman:
@sylvhama sure, I think that sounds like a simple solution.

eliperelman:
Closing per https://github.com/mozilla-neutrino/neutrino-dev/pull/732#issuecomment-390821358.

PR #738, opened 2018-03-10
Update yarn lockfile

edmorley:
Generated by `rm yarn.lock && yarn` to pick up latest in-range updates that end-users will see when installing packages.

In another PR I'll enable Renovate's periodic updating of lockfile.


edmorley:
Only affects monorepo development/testing, and Travis passes, so merging without additional review.

PR #744, opened 2018-03-12
Remove stray package-specific yarn.lock

edmorley:
And update the repo root `yarn.lock` again.

Closes #743.


rarkins:
@edmorley if you can work out a simple repo/example to reproduce it, I can take a look. Otherwise it's frankly going to remain low in priorities if I need to be able to reverse engineer a solution from the logs. It might be as simple though as the stray lockfile wasn't found after workspaces was run, although adding "lock file not found" isn't that much more of a help over `undefined` :)

PR #758, opened 2018-03-22
Cleanup monorepo linting configuration

edmorley:
Fixes some of the issues found whilst looking at #608.

See individual commits/commit messages for more details.

### DIFF
--- a/.neutrinorc.js
+++ b/.neutrinorc.js
@@ -11,15 +11,12 @@ module.exports = {
       eslint: {
         baseConfig: {
           extends: [
-            'plugin:react/recommended',
-            'prettier',
-            'prettier/react'
+            'prettier'
           ]
         },
         envs: ['browser', 'commonjs', 'node'],
         plugins: [
-          'eslint-plugin-prettier',
-          'eslint-plugin-react'
+          'prettier'
         ],
         rules: {
           // Algebraic and functional types should allow capital constructors without new
--- a/.neutrinorc.js
+++ b/.neutrinorc.js
@@ -21,80 +21,18 @@ module.exports = {
         rules: {
           // Algebraic and functional types should allow capital constructors without new
           'babel/new-cap': 'off',
-          // Disable necessitating return after a callback
-          'callback-return': 'off',
-          // Allow using class methods with static/non-instance functionality
-          // React lifecycle methods commonly do not use an instance context for anything
-          'class-methods-use-this': 'off',
           // Disallow trailing commas on arrays, objects, functions, et al
           'comma-dangle': ['error', 'never'],
-          // Require all requires be top-level
-          // http://eslint.org/docs/rules/global-require
-          'global-require': 'error',
-          // Enforces error handling in callbacks (node environment)
-          'handle-callback-err': 'off',
           // Allow dynamic requires
           'import/no-dynamic-require': 'off',
-          // Keep all the original Airbnb indentation rules except MemberExpression,
-          // which messes up our pretty chaining indentations
-          indent: ['error', 2, {
-            SwitchCase: 1,
-            VariableDeclarator: 1,
-            outerIIFEBody: 1,
-            MemberExpression: 'off',
-            FunctionDeclaration: {
-              parameters: 1,
-              body: 1
-            },
-            FunctionExpression: {
-              parameters: 1,
-              body: 1
-            },
-            CallExpression: {
-              arguments: 1
-            },
-            ArrayExpression: 1,
-            ObjectExpression: 1,
-            ImportDeclaration: 1,
-            flatTernaryExpressions: false,
-            ignoredNodes: ['JSXElement *']
-          }],
-          // Specify the maximum length of a line in your program
-          // JSX can get lengthy, so this helps alleviate that a bit
-          // http://eslint.org/docs/rules/max-len
-          'max-len': ['error', 120, 2, {
-            ignoreUrls: true,
-            ignoreComments: false,
-            ignoreStrings: true,
-            ignoreTemplateLiterals: true
-          }],
           // Allow console during development, otherwise throw an error
           'no-console': 'off',
-          // Allow extra parentheses since multiline JSX being wrapped in parens is considered idiomatic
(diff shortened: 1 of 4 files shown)


edmorley:
Another issue not dealt with here (since it requires hundreds of fixes) is that whilst the prettier is being used to turn off conflicting AirBnB rules, the actual prettier rules to replace them aren't enabled. To do so requires adding:
```js
        rules: {
          // ...
          'prettier/prettier': 'error'
        }
```

.neutrinorc.js
```diff
@@ -1,103 +1,37 @@
 module.exports = {
   use: [
-    ['./packages/airbnb-base', {
+    ['./packages/airbnb', {
```
eliperelman: As mentioned in the other comment, we don't actually use any react in the repo, and aren't linting the create-project JSX files in this repo using our own configuration. In my opinion, I think this should stay as airbnb-base.

.neutrinorc.js
```diff
@@ -1,103 +1,37 @@
...
+      // Excludes are managed vis `.eslintignore` since `exclude` doesn't support globs.
```
eliperelman: s/vis/via/

.neutrinorc.js
```diff
@@ -1,103 +1,37 @@
...
         rules: {
           // Algebraic and functional types should allow capital constructors without new
           'babel/new-cap': 'off',
-          // Disable necessitating return after a callback
-          'callback-return': 'off',
-          // Allow using class methods with static/non-instance functionality
-          // React lifecycle methods commonly do not use an instance context for anything
-          'class-methods-use-this': 'off',
           // Disallow trailing commas on arrays, objects, functions, et al
           'comma-dangle': ['error', 'never'],
```
eliperelman: I'm fine with removing this if you like.

packages/create-project/commands/init/index.js
```diff
@@ -10,7 +10,7 @@ const { projects, packages, isYarn } = require('./utils');
 
 /* eslint-disable no-underscore-dangle */
 module.exports = class Project extends Generator {
-  _logo() {
+  static _logo() {
```
eliperelman: ++

.neutrinorc.js
edmorley: If that's the case, we should remove the react eslint plugin references from `,neutrinorc.js` then - since we are already using a React based linting config - this change was keeping the status quo.

That said, I wonder if we should actually eventually remove the create-project templates from `.eslintignore` and start linting them like we do the rest of the repository. Yes we'll still need the one off lint step during create-project bootstrap, but at least we'd catch/fix more obvious issues in the code we have checked in - plus it would improve consistency with the rest of the monorepo from a code style/best practices perspective.

eliperelman: > but at least we'd catch/fix more obvious issues in the code we have checked in

I think this is a compelling reason to at least lint it ourselves...somehow. It would be nice to catch eslint **errors** at patch-time instead of finding them in release. Stylistic fixes will still be done during create-project.

.neutrinorc.js
edmorley: Ah so when I looked into `comma-dangle` it was a bit confusing. Airbnb enables it as 'always', the eslint-config-prettier preset then turns it off entirely, and ordinarily the 'prettier/prettier' rule would then turn it back on as 'never' - but we're not using 'prettier/prettier' at the moment.

So removing this line would make it `'comma-dangle': 'off'`. (Which may or may not be what you meant?) :-)

.neutrinorc.js
edmorley: I was hoping the eslint `overrides' feature would let us do this, however it seemed like it only works via the ESLint API, only via the CLI sadly.

.neutrinorc.js
eliperelman: Grr, I'm dumb. I guess I was trying to align with:

https://github.com/mozilla-frontend-infra/neutrino-preset-mozilla-frontend-infra/blob/master/lint.js#L43

Thoughts?

edmorley: Zooming out a bit, my thoughts are:
* I like the idea of Neutrino dogfooding it's own linting configs
* To get the most value of dogfooding we should limit the amount we override them (and if things really are an issue, fix the preset directly)
* If we want to use prettier, perhaps medium-long term we should create/use a prettier neutrino preset
* Long term I'd like to see most of the style rules in neutrino-preset-mozilla-frontend-infra go away (I don't really think they belong there - ideally projects should use stock presets and not deviate too much. And I don't think that preset should turn off things like `react/prefer-stateless-function` - the individual projects should make those choices IMO)

edmorley: > And I don't think that preset should turn off things like react/prefer-stateless-function - the individual projects should make those choices IMO

To clarify on this part - I agree that for legacy project some rules (eg `react/prop-types`) are too burdensome - however by hiding the setting of it to `'off'` in neutrino-preset-mozilla-frontend-infra, it means brand new projects (which have no reason to not do things right from the start) don't know to turn it on.

eliperelman: Good points all around. Most of the rules in our preset are stylistic choices to improve readability/maintainability, prevent mutation bugs, or get in the way of doing FP.

Let's raise these as issue in that preset.

edmorley: Fwiw the AirBnB style guide are sometimes receptive to suggestions for rule changes. By filing issues against them and linking to those issues when making adjustments, it would (a) show that upstreaming has been attempted, (b) give breadcrumbs to follow for understanding why the deviation :-)

eliperelman: Yeah, I should also try and document why we currently have those deviations in the interim.

edmorley:
I've made the `s/vis/via/` change. Regarding airbnb vs airbnb-base, should I:
* leave the PR as is
* update the PR to use airbnb-base again, but still remove the react plugin/rule references etc

eliperelman:
> update the PR to use airbnb-base again, but still remove the react plugin/rule references etc

Probably this, so it's not confusing that it's not doing anything right now.

PR #778, opened 2018-03-28
Use verdaccio to test create-project with proxied registry

eliperelman:


eliperelman:
I split the serial jobs into separate Travis build matrices, but the benefit still isn't there. Since Travis only gives us 5 concurrent builds, it still ends up taking 30 minutes, even though each build is only about 3-4 minutes long now.

If we had more concurrent builds, this would be a whole lot better.

PR #789, opened 2018-04-14
Update yarn lockfile

edmorley:
Generated via `rm -rf node_modules yarn.lock && yarn` using yarn 1.6.0.

(Renovate's lockfile maintenance feature is/was broken, so updating manually in the meantime.)


edmorley:
I'm going to just merge this given it passes tests, only affects dev/testing and is a mechanical change.

PR #792, opened 2018-04-14
Drop support for Node.js 6

edmorley:
Since by the time Neutrino 9 is released, Node 6 will already be in maintenance mode (and anyone who needs to continue using it can just keep using Neutrino 8 instead):
https://github.com/nodejs/Release#release-schedule

This also changes the default babel targets for the `library` preset (when used in `node` mode) and the `node` preset, from `6.10` to `8.0`. (I'm happy to drop this part of the PR if preferred.)


docs/packages/node/README.md
```diff
@@ -291,24 +291,15 @@ module.exports = {
...
 ```
 
-_Example: Override the Node.js Babel compilation target to Node.js v8:_
+_Example: Override the Node.js Babel compilation target to Node.js v6:_
 
 ```js
 module.exports = {
   use: [
     ['@neutrinojs/node', {
-      // Add additional Babel plugins, presets, or env options
```
eliperelman: Nice catch here!

Issue #748, opened 2018-03-14
Compare speed/size of babel-minify vs uglify-es


edmorley:
Currently Neutrino 8 uses babel-minify (formerly babili), in part because uglifyjs previously didn't have ES6 support. 

However with webpack 4, [uglify-es](https://github.com/mishoo/UglifyJS2/tree/harmony) is now used (via [uglifyjs-webpack-plugin](https://github.com/webpack-contrib/uglifyjs-webpack-plugin)), which supports ES6 features - plus caching and parallelisation is supported out of the box.

For projects like mozilla/treeherder, JS minification currently accounts for 80%+ of the production build time, so if there was any way to speed it up (without having a big impact on bundle size), it would make a massive difference to our end to end CI and deploy times.

As such once we're using webpack 4, we should perform another comparison of the speed/output size/features of babel-minify vs uglify-es, to see:
(a) if there are settings we could adjust to improve the speed (eg enabling parallisation with babel-minify, to match that now supported with uglify-es
b) whether we should change the Neutrino default
c) whether we should provide a Neutrino web preset option to make it easier for uses to switch between the two

edmorley:
Meant to say - in addition speed/output size considerations, we should also factor in:
* correctness/known bugs (for example webpack-contrib/babel-minify-webpack-plugin#68)
* whether it's safer to stick to the webpack default, which is a more tested path

edmorley:
Although it's looking like uglify-es's future is up in the air at the moment:
https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/262

evilebottnawi:
@edmorley Please read https://github.com/webpack-contrib/babel-minify-webpack-plugin/issues/68#issuecomment-379210998

PR #809, opened 2018-04-20
Update to webpack 4

### LINKED ISSUE #737: Add support for webpack 4
Webpack 4 has now been released. Whilst their docs have not yet been updated (eg webpack/webpack.js.org#1856 and friends) - it would be good to start looking at what we need to do to support webpack 4 soon.

I propose something like:
1. Review/merge any open non-breaking-change PRs (including those from Renovate)
2. Publish a new Neutrino 8.x release and make sure the `release/v8` branch is updated
3. Alter `master` to be Neutrino 9 dev/alpha/whatever, making it ok for breaking changes (we can still ship urgent Neutrino 8.x bugfix releases later, but from the release branch)
4. Merge any remaining Renovate PRs, so we pick up webpack 4 fixes to deps
5. Update webpack-chain to support any new webpack 4 options
6. Vet all of our deps to see if there are any that don't yet support webpack 4
7. Open a PR to add webpack 4 support to Neutrino master
8. After a few iterations on the PR (and once it's clear the webpack 4 docs/best practices have stabilised), merge to master
9. Release a Neutrino 9 alpha
10. Once Babel 7 is released and we've updated to that, release Neutrino 9 beta/final...

Anything I've missed? :-)

### LINKED ISSUE #802: CSS extraction includes css-hot-loader even in production
The `web` preset supports extraction of CSS to a separate file as of #443, via `@neutrinojs/style-loader` and then `extract-text-webpack-plugin`.

The `extract-text-webpack-plugin` loader/plugin doesn't support HMR hot reloading itself (unlike `style-loader`, which works since it's just JS) - so if one wants to use it in development, then `css-hot-loader` must be used alongside it.

However at the moment the web preset:
* only enables extraction in production:
  https://github.com/mozilla-neutrino/neutrino-dev/blob/ee7b3d90fe93da0b6664b194da2994d4ce79f26e/packages/web/index.js#L37
* but includes `css-hot-loader` whenever extraction is enabled, even if that's in production:
https://github.com/mozilla-neutrino/neutrino-dev/blob/ee7b3d90fe93da0b6664b194da2994d4ce79f26e/packages/style-loader/index.js#L102-L104

This combination is confusing, since by default `css-hot-loader` is only being included in production - where it will never be used.

We should either:
1) Enable extraction in both development and production, plus in development only, also use `css-hot-loader`
2) Continue to enable extraction only in production by default, and remove the use of `css-hot-reloader` entirely
3) Continue to enable extraction only in production by default, and make the inclusion of `css-hot-reloader` conditional on production too. (This would allow people to force extraction to be enabled in development, and yet still have working hot loading)

Also relevant:
* `extract-text-webpack-plugin` doesn't support webpack 4 - so for that we're going to need to switch to `mini-css-extract-plugin`. For now the latter also requires the use of `css-hot-reloader` if being used in development, however it appears that they are hoping to add native support for HMR thereby making `css-hot-reloader` redundant (webpack-contrib/mini-css-extract-plugin#34).
* the discussion in webpack-contrib/extract-text-webpack-plugin#30 seems to suggest that currently the recommended solution is to not use CSS extraction in development (and instead rely on `style-loader`)

As such, I'm leaning towards option (2), with the understanding that we will likely enable extraction everywhere once `mini-css-extract-plugin` supports it natively.

@timkelty @eliperelman - thoughts?

Whilst we're on this topic - it seems like we could improve the way we set default options. This issue in part exists because at the moment, we set the `style.extract` default according to production, but yet for the `hot` option, we always set it to `true` and instead later use a `.when()` clause to decide whether to invoke `devServer` etc. But this means that `@neutrinojs/style-loader` is passed `hot: true` even in production. I was wondering too if we should move defaults one layer deeper? (ie in `@neutrinojs/style-loader`)?

edmorley:
Notable changes:

* webpack 4's new [splitChunks](https://webpack.js.org/plugins/split-chunks-plugin/) feature replaces the `@neutrinojs/chunk` middleware - so the fake `vendor` entrypoint should now be removed from `.neutrinorc.js` to avoid a bigger build then necessary (a check has been added to the web preset to enforce this).
* the default javascript minifier has been switched from [babel-minify](https://github.com/babel/minify) to [uglify-es](https://github.com/mishoo/UglifyJS2/tree/harmony) (to override, pass a non-falsey value to the web preset's `minify.babel` option).
* [extract-test-webpack-plugin](https://github.com/webpack-contrib/extract-text-webpack-plugin) has been replaced by [mini-css-extract-plugin](https://github.com/webpack-contrib/mini-css-extract-plugin).

For more details see the individual commit messages (most of the intermediate commits will fail to run standalone, but it still seemed preferable to split them up for easier review).

To try this branch out locally:

1. `git clone https://github.com/mozilla-neutrino/neutrino-dev.git`
2. `cd neutrino-dev && git checkout webpack-4`
3. `yarn && yarn link:all`
4. `cd <path to existing Neutrino 8 project>`
5. `yarn && yarn link neutrino && for d in node_modules/@neutrinojs/*; do yarn link "${d#*/}"; done` (there will be one or two link errors eg for `@neutrinojs/babel-minify-webpack-plugin` - that's fine)
6. Try out `yarn {start, build, lint, test}` as you would normally
7. Once finished testing, to switch back to the published packages, by:
    - in the project that uses Neutrino: `rm -rf node_modules && yarn`
    - in the neutrino-dev repository: `yarn lerna exec yarn unlink`

Alternatively create a new template project by ensuring the `yarn global bin` directory is on `PATH`, and then replacing step 4 with:
`cd .. && create-project <new directory name> && cd <new directory name>`

I'd love to have any and all feedback - including what affect this branch has on build times/sizes compared to Neutrino 8, and how uglify-es vs babel-minify compares for your use-cases (by toggling `minify.babel` back and forth).

Useful resources for reviewing:

* [WIP official migration docs](https://github.com/webpack/webpack.js.org/pull/1939)
* [`mode` docs](https://webpack.js.org/concepts/mode/)
* [`splitChunks` docs](https://webpack.js.org/plugins/split-chunks-plugin/)
* webpack [default optimization options](https://github.com/webpack/webpack/blob/v4.6.0/lib/WebpackOptionsDefaulter.js#L196-L294) and [resultant plugins enabled](https://github.com/webpack/webpack/blob/v4.6.0/lib/WebpackOptionsApply.js#L301-L338)
* [mini-css-extract-plugin docs](https://github.com/webpack-contrib/mini-css-extract-plugin#usage)
* [create-react-app's webpack 4 PR](https://github.com/facebook/create-react-app/pull/4077)

Left to do:

- [ ] Decide whether to use `uglify-es` or `babel-minify` by default (I'm leaning towards the former, see #748).
- [ ] Decide whether to allow choosing minifier in the web preset, or if `@neutrinojs/babel-minify` (presuming we pick `uglify-es`) should be completely separated to reduce dependencies/complexity. (Especially since a future [minimizer-webpack-plugin](https://github.com/webpack-contrib/babel-minify-webpack-plugin/issues/68#issuecomment-379210998) might mean any custom handling in the web preset has to be rewritten anyway.)
- [ ] Make the relevant minification changes based on those decisions.
- [ ] Keep an eye on jantimon/html-webpack-plugin#880 and webpack-contrib/karma-webpack#322, and remove the workarounds for them if fixed prior to this being merged.

Fixes #737.
Fixes #802.
Closes #768.
Closes #769.
Closes #766.


edmorley:
I've updated the above instructions for trying out this branch to prevent leftover packages in the "existing project" case. (If you see errors like `TypeError: dep.getResourceIdentifier is not a function` or `Error: Cannot find module 'json-loader'` it means there's a mixture of webpack 3 and 4 packages being used - the revised steps should avoid this.)

eliperelman:
Preliminary builds using current and updated neutrino-preset-mozilla-frontend-infra on taskcluster-web:

screen shot 2018-04-21 at 10 16 00 am

screen shot 2018-04-21 at 10 24 10 am

Build time went from 154 seconds to 46 seconds. Incredible.

---

I noticed that in my updated preset, since the chunking logic is gone, I can no longer do this:

screen shot 2018-04-21 at 10 28 14 am

Does this mean that `.chunkFilename(`[name].[chunkhash].${cacheVersion}.js`)` should work properly for those chunks now?

edmorley:
Thank you for trying it out - nice improvement in build time! :-)

I think there's more caching going on by default now (`node_modules/.cache`) - how do the second build times compare? Also, how long does webpack 4 + `{minify: { babel: true }}` take compared to the default of uglify-es?

> Does this mean that .chunkFilename([name].[chunkhash].${cacheVersion}.js) should work properly for those chunks now?

Yeah I'm pretty sure that should work. 

Re the automatically generated chunk filenames - since this PR sets `optimization.splitChunks.name` to `false` in production, we end up with filenames like `1.ceddedc0defa56bec89f.js` rather than the default of `vendors~index~page2.b694ee990c08e6be6a33.js`. Now the latter avoids cache busting just because another entrypoint was added, however isn't very friendly. 

As such, I was thinking of filing an issue against webpack asking for the default when `name` is false to instead include the name of the `cacheGroup` - which in this case is `vendor`. This would give filenames more like `vendors~1.b694ee990c08e6be6a33.js`. I haven't yet played around with passing functions to `name`, to see if we could do this ourselves in the meantime.

eliperelman:
taskcluster-tools, using the same current and updated preset, removing the vendor bundle this time, prior to the updated build:

screen shot 2018-04-21 at 10 49 44 am

The second screenshot is rather large, so splitting in 4:

screen shot 2018-04-21 at 11 04 44 am

screen shot 2018-04-21 at 11 05 20 am

screen shot 2018-04-21 at 11 05 36 am

screen shot 2018-04-21 at 11 05 51 am

---

From 431 seconds to 61 seconds. **From 7 minutes to 1 minute**. Amazing.

eliperelman:
Build summaries:

taskcluster-web:

| project | webpack | primed | minifer | time | v3 improvement |
| --- | --- | --- | --- |
| taskcluster-web | v3 | no | babel | 154s | - |
| taskcluster-web | v4 | no | uglify | 46s | 70.1% |
| taskcluster-web | v4 | yes | uglify | 25s | 83.8% |
| taskcluster-web | v4 | no | babel | 88s | 42.9% |
| taskcluster-web | v4 | yes | babel | 84s | 45.5% |
| taskcluster-tools | v4 | no | babel | 431s | - |
| taskcluster-tools | v4 | no | uglify | 61s | 85.9% |
| taskcluster-tools | v4 | yes | uglify | 41s | 90.5% |
| taskcluster-tools | v4 | no | babel | 168s | 61.0% |
| taskcluster-tools | v4 | yes | babel | 162s | 62.4% |

---

Takeaways:

- Uglify is significantly faster than Babel minify.
- Uglify is also significantly better with incremental builds over non-primed build. Babel minify shows no real statistical improvement in incremental builds.

edmorley:
PR rebased and conflicts from master resolved, then 65f77a3 and 8bc0d3b added to remove `@neutrinojs/babel-minify` and the `webpack-sources` pinning.

eliperelman:
Updated stats

| project | webpack | primed | minifer | time | v3 improvement |
| --- | --- | --- | --- | --- | --- |
| taskcluster-web | v3 | no | babel | 154s | - |
| taskcluster-web | v4 | no | uglify | 46s | 70.8% |
| taskcluster-web | v4 | yes | uglify | 23s | 85.1% |
| taskcluster-web | v4 | no | babel | **FAILED** | - |
| taskcluster-web | v4 | yes | babel | **FAILED** | - |
| taskcluster-tools | v3 | no | babel | 431s | - |
| taskcluster-tools | v4 | no | uglify | 76s | 82.4% |
| taskcluster-tools | v4 | yes | uglify | 36s | 91.7% |
| taskcluster-tools | v4 | no | babel | **FAILED** | - |
| taskcluster-tools | v4 | yes | babel | **FAILED** | - |

---

Looks like numbers are comparable within margin of error.

---

Looks like the Babel minify error is back with the sources unpinned:

> Error: original.line and original.column are not numbers -- you probably meant to
> omit the original mapping entirely and only map the generated position. If so,
> pass null for the original mapping instead of an object with empty or null values.

packages/web/index.js
```diff
@@ -6,17 +6,15 @@ const compileLoader = require('@neutrinojs/compile-loader');
...
-const chunk = require('@neutrinojs/chunk');
 const clean = require('@neutrinojs/clean');
-const babelMinify = require('@neutrinojs/babel-minify');
 const styleMinify = require('@neutrinojs/style-minify');
 const loaderMerge = require('@neutrinojs/loader-merge');
 const devServer = require('@neutrinojs/dev-server');
 const { join } = require('path');
 const { resolve } = require('url');
 const merge = require('deepmerge');
+const HtmlWebpackIncludeSiblingChunksPlugin = require('html-webpack-include-sibling-chunks-plugin')
```
eliperelman: Missing trailing semicolon. Thanks eslint!

edmorley:
Or alternatively removing the `babel-minify` instructions entirely and replacing them with a mention that it's just too buggy and not recommended.

edmorley:
So re-reading the babel-minify-webpack-plugin issue I see that it occurs for all source map variants apart from cheap (and cheap is pointless when minifying, since everything is on one line), so I guess that leaves telling people to just disable sourcemaps in production if using babel-minify. (Which seems reasonable; it's not our fault it's buggy)

eliperelman:
OK, I can agree with that. Let's just not support it for now with a warning.

PR #810, opened 2018-04-22
Extract css without contenthash by default (react-components preset)

Didericis:
Changes the default options for styles so that the content hash isn't included in filenames when css is extracted. This makes stylesheets easier to import.

EX:

```js
// BEFORE
import 'my-component/build/MyComponent.qwerihqwenoasidf.css'
// AFTER
import 'my-component/build/MyComponent.css'
```


eliperelman:
Yeah, we should probably port to v8 if we can.

edmorley:
This has just been released in v8.2.3 :-)

PR #814, opened 2018-04-23
Remove default CopyPlugin from src/static/ to build/static/

edmorley:
Since for most assets users should be importing in the build, and for the valid use-cases of not doing that (eg `robots.txt`), this default plugin didn't help anyway.

Another step forwards in making Neutrino default to fulfilling the 80% case, and then leaving the 20% to opt-in middleware, rather than having too much enabled by default.

### DIFF
--- a/docs/packages/node/README.md
+++ b/docs/packages/node/README.md
@@ -332,7 +332,6 @@ _Note: Some plugins are only available in certain environments. To override them
 | Name | Description | Environments and Commands |
 | --- | --- | --- |
 | `banner` | Injects source-map-support into the mains (entry points) of your application if detected in `dependencies` or `devDependencies` of your package.json. | Only when `source-map-support` is installed |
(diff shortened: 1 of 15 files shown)


constgen:
But `@neutrinojs/copy` will stay in the core in case somebody will use it, will it?

constgen:
Makes sense

davidje13:
For those poor saps like me experimenting with updating to v9, could you add the old configuration to the migration guide so that it's easy to maintain old behaviour? There are a **lot** of breaking changes and the more I can just patch with quick fixes the better.

From looking at the PR I believe the code is:

```javascript
neutrino.use(copy, {
  patterns: [{
    context: 'src/static',
    from: '**/*',
    to: 'static',
  }],
});
```

Or equivalently:

```javascript
['@neutrinojs/copy', {
  patterns: [{
    context: 'src/static',
    from: '**/*',
    to: 'static',
  }],
}],
```

(plus a new dependency on `@neutrinojs/copy`)

Issue #713, opened 2018-02-19
Scale back default minification in core presets


timkelty:
There have been a number of issues since we began including `@neutrinojs/minify` (minimizing babel, styles, images) in core presets: #678, #673, #700

My preference would be to scale back what the core presets do by default, and instead point users to additional middleware they may want to use (image-minify, style-minify).

That said - we probably also need to make sure the minify middleware include an `enabled` option (defaulting to `NODE_ENV === 'production'`), so that users could just include them without having to configure them.

Especially if we go this route, I'm not convinced theres much value in keeping `@neutrinojs/minify` around. Seems more explicit to have `@neutrinojs/web` include `@neutrinojs/babel-minify` directly, and direct users to include `@neutrinojs/image-minify` and `@neutrinojs/style-minify` if they desire.

I can work on a PR for v9, but curious to hear what others think first.

/cc @edmorley @mozilla-neutrino/core-contributors

edmorley:
Sorry for the delayed reply! Yeah I agree it would be good to scale back the default packages included with `@neutrinojs/web`, and let people include them as needed (I've been contemplating skipping web and using the component pieces individually recently, due to the number of things now included in it that we don't use).

timkelty:
@edmorley anything besides what is mentioned here (image/style-minify) you might want considered for removal from web?

edmorley:
It's tricky, because I'm sure some of the things [mozilla/treeherder](https://github.com/mozilla/treeherder) doesn't use, others do - such as:
* `fast-async` (since we only support more modern browsers; though in fact the Neutrino default targets all support async now I think...so maybe this can be removed anyway?)
* `webpack-manifest-plugin`
* the `@neutrinojs/copy` copying to `staticDir`
* `worker-loader`
* `html-webpack-template` (technically we use it at the moment, but would prefer to use the upstream template; or if that has bugs get them fixed first)
* `@neutrinojs/image-minify`
* `@neutrinojs/style-minify`

(And then in things like `@neutrinojs/karma` we don't use any of the Chrome related `dependencies`, overriding them with Firefox instead; but that's a different story)

timkelty:
To some degree it matters if you're talking about the requiring dependencies themselves, or whether anything is actually done with them.

For example, a config of `{html: false, manifest: false}` would disable those features, though those packages would still be deps.

edmorley:
[...]

PR #815, opened 2018-04-24
Split up the minify preset

### LINKED ISSUE #807: The library preset enables all types of minification by default
The library preset `.use()`s the minify preset without passing any options, which activates all of {babel-minify, style-minify, image-minify}:
https://github.com/mozilla-neutrino/neutrino-dev/blob/ce7139784507f788ad8509ef5cbc0ad30da34f36/packages/library/index.js#L150

Compare this to the web preset:
https://github.com/mozilla-neutrino/neutrino-dev/blob/ce7139784507f788ad8509ef5cbc0ad30da34f36/packages/web/index.js#L213

Issues:
1) I'm not sure all types of minification be enabled by default. Thoughts?
2) We should allow people to adjust the options, like we do for the web preset.

Filing this since it affects the current implementation of the webpack-4 branch, since that uses (at least for now, to let us A/B test) a toggle to switch between uglify-es and babel-minify.

edmorley:
* Removes the web preset's dependency on `@neutrinojs/image-minify` since it has many heavyweight native code dependencies, and isn't enabled by default anyway (see #713).
* Removes the library preset's dependency on `@neutrinojs/style-minify` and `@neutrinojs/image-minify` (fixes #807).
* Removes the `@neutrinojs/minify` preset in favour of directly depending on the individual `@neutrinojs/*-minify` presets.
* Adds an `enabled` option to `@neutrinojs/image-minify` that defaults to production only, to allow for string form usage of the preset in `.neutrinorc.js`, without the need to specify a `.when()`.


packages/image-minify/test/middleware_test.js
```diff
@@ -2,32 +2,42 @@ import test from 'ava';
 import { Neutrino } from 'neutrino';
 
 const mw = () => require('..');
+const prodEnv = { 'env': { NODE_ENV: 'production' } };
```
eliperelman: We can remove the string from the `env` key.

packages/image-minify/test/middleware_test.js
```diff
@@ -2,32 +2,42 @@ import test from 'ava';
...
+const devEnv = { 'env': { NODE_ENV: 'development' } };
```
eliperelman: We can remove the string from the `env` key.

eliperelman:
I think this also fixes #670 in master.

edmorley:
Though presumably anyone who explicitly adds the `@neutrinojs/image-minify` preset will still be affected?

eliperelman:
Correct.

Issue #306, opened 2017-07-27
WebExtension preset


graingert:
I know you're not actively looking for presets, but a preset supporting WebExtensions would directly further Mozilla's mission.

constgen:
Pull request is welcome.
The most hard in implementation will be to setup the `neutrino start`. What should happen on this command?

graingert:
@constgen launch a Firefox profile with the extension loaded

graingert:
@constgen maybe even use a web socket to uninstall and reinstall the extension when it updates

constgen:
I know that Firefox has a special API for livereloading extensions. But what about Chrome and Edge?

graingert:
@constgen use HMR for those browsers?

constgen:
Background process will not be reloaded in those browsers, will it?
You would like to have a reimplementation of https://github.com/mozilla/web-ext as a preset. Am I right.

eliperelman:
I'm going to close this as something we aren't going to handle in core right now. If someone from the community wants to pick this up, that would be very appreciated!

graingert:
Bummer

eliperelman:
We are considering bringing this into core again. If you are interested in working on this, please let me know and I can point out a few resources for inspiration.

helfi92:
@eliperelman I can look into that. Let me know how I can help :)

eliperelman:
@helfi92 here's a few resources to help you get started:

- Look into the [web-ext tool](https://github.com/mozilla/web-ext). We may only need this tool for the run/start portion of it, since webpack should be able to handle the rest. Maybe we can use [webpack-webext-plugin](https://github.com/rpl/webpack-webext-plugin), maybe not.
- Check out some [webextension examples](https://github.com/mdn/webextensions-examples/tree/master/store-collected-images/webextension-with-webpack). Important things to note are creation of `background.js`, `popup.js`, and content scripts, as well as a manifest.json. Some of these we may be able to generate with create-project, and some we may be able to create at build time, I'm not sure.

eliperelman:
Being worked on in #567.

eliperelman:
Closing as we don't have anyone that could actively maintain this. Feel free to reopen if a patch is submitted with an active maintainer for this.

crimx:
I've created a [neutrino-webextension](https://github.com/crimx/neutrino-webextension) preset which had been successfully used in couple of my own projects over the pass few months.

[...]

PR #831, opened 2018-04-29
Web Extension Preset

Morikko:
## Creating web extension easily over webpack and web-ext
- Use web-ext tool for development (open a specific browser environment)
- Build the final project with web-ext (final source, ZIP and XPI)
- Lint the project with web-ext on top of any classical code lint
- Read the manifest to detect automatically some paths (panel, popup...)

See the README.md in webextension/ for more information.

## Note
- This PR is the continuation of #567 based on issue #306.
- I haven't rebase on 8.2.3 because I got webpack error I don't have with 8.2.0
- A review of the code would be welcomed for knowing if I am doing the preset in the right way  

[Repository for testing: webextension-neutrino-test](https://github.com/Morikko/webextension-neutrino-test)


Morikko:
## Problems

 - [ ] Copied files are not removed in ./dev if removed from ./src
 - [ ] New files in ./src don't trigger a rebuild (but are copied after a rebuild due to a modification on another file)
 - [ ]  Lots of warnings due to the use of `eval`
 - [ ] When building, all js target (without HTML: content scripts and background.js) are not working because `runtime.js` is not imported...
 - [ ]  Lots of error reported in Firefox console from the bundled files. Example below.
```
   ReferenceError: reference to undefined property "disabled"[Learn More]  CustomizableUI.jsm:2480:5
   TypeError: cannot use 'in' operator to search for 'canGoBack' in 'browser'[Learn More]  tabbrowser.xml:2473:1
   ReferenceError: reference to undefined property "additionalBackgrounds"[Learn More]  LightweightThemeConsumer.jsm:161:5
   ReferenceError: reference to undefined property "backgroundsAlignment"[Learn More]  LightweightThemeConsumer.jsm:205:5
   ReferenceError: reference to undefined property "initialPageLoadedFromURLBar"[Learn More]  tabbrowser.xml:697:1
   ...
```

eliperelman:
To update here, we made significant changes in #852 that eliminates the Neutrino CLI. To continue forward with this, we would need to change this to instead output the config required for webext to consume the middleware. Are you still willing to pick up on this?

Morikko:
If I understood well, in webext-config.js, it would be like this: 
```

module.exports = neutrino().webext(config => {
  // Overwrites the config here
});
```
Where neutrino().webext() is the default config  from the middleware (like in `packages/webextension/index.js`) ?

edmorley:
Hi!

Yes that sounds right.

After #852, we no longer have a Neutrino CLI that calls various tools by their node API, and instead rely on the tool's own CLIs / native config files. 

I'm not very familiar with web extensions, but reading through the [web-ext package docs](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Getting_started_with_web-ext) it seems that it will auto-detect a file named `.web-ext-config.js` or `web-ext-config.js`. As such, if the web-ext CLI is to be used, that file should contain the `module.exports = neutrino().webext()`, and the `.webext()` command registered using something like:
https://github.com/mozilla-neutrino/neutrino-dev/blob/b6eb31ffee0b1e1f50504355d74da4472b4a03b8/packages/jest/src/index.js#L21

That said, I can't tell from skimming through this PR whether it used to invoke the webpack build at all? (I guess the `neutrino.on('build', ...)` was cumulative, and so it did invoke webpack as well as the webext API?)

If it did, then perhaps this preset needs to change to using the webpack plugin instead? (Since otherwise two separate CLIs would have to be run sequentially)
https://github.com/rpl/webpack-webext-plugin

A couple of other things I spotted in the PR:
* `config.resolve.modules` and `config.resolveLoader.modules` can be removed (see #926)
* some of the deepmerge usages can be replaced with the spread operator (now that we require node 8.3+)

Hope that helps :-)

edmorley:
Closing this out for now since there's no activity. If you or someone else has a chance to work through the suggested changes above, happy to answer any questions you might have :-)

Issue #836, opened 2018-04-30
Vue preset's babel-preset-vue dependency is unused


edmorley:
The vue preset has a dependency on [babel-preset-vue](https://github.com/vuejs/babel-preset-vue):
https://github.com/mozilla-neutrino/neutrino-dev/blob/6dee32757ee93ec1541124acd59b3c990f5ea853/packages/vue/package.json#L26

However as far as I can tell it's not actually being used:
https://github.com/mozilla-neutrino/neutrino-dev/blob/6dee32757ee93ec1541124acd59b3c990f5ea853/packages/vue/index.js#L11-L27

Should it be removed, or are we meant to be using it?

@helfi92, any ideas?

edmorley:
@helfi92 as the creator of this preset, do you have any ideas? :-)

helfi92:
I believe it can be deleted. PR on the way.

PR #845, opened 2018-05-03
Update to Babel 7

### LINKED ISSUE #316: Investigate moving to Babel 7 when it releases
https://github.com/babel/babel/milestone/14

edmorley:
Notable changes:
* All official packages have moved under the `@babel/` namespace and in some cases further renamed, so package names adjusted accordingly.
* `@babel/preset-env` now includes `@babel/plugin-transform-spread` and `@babel/plugin-transform-classes`, which are the Babel 7 renames of `babel-plugin-transform-object-rest-spread` and `babel-plugin-transform-es2015-classes`.
* `@babel/preset-env`'s `useBuiltIns: true` mode has been renamed to `useBuiltIns: 'entry'` - which is equivalent.
* `@babel/preset-react` now has `development` and `useBuiltIns` options, which we set appropriately:
  https://github.com/babel/babel/tree/v7.0.0-beta.46/packages/babel-preset-react#options
* `babel-plugin-dynamic-import-node` is no longer required by `@neutrinojs/library` for target `node`, since webpack converts the dynamic import to a `require()` itself.
* `@neutrinojs/jest` required more substantial changes since:
  - the custom transformer used the now removed `canCompile()` from Babel's API.
  - `babel-preset-jest` is not fully compatible with Babel 7 (facebook/jest#6126).
* Several packages now have a peer dependency on `@babel/core`, so it's been added where necessary.
* `babel-loader` has been updated to v8 for Babel 7 compatibility.
* `@neutrinojs/vue`'s `babel-preset-vue` dependency doesn't appear to be used, but has been left as-is pending #836.

Closes #316.

### DIFF
--- a/packages/jest/src/transformer.js
+++ b/packages/jest/src/transformer.js
@@ -5,13 +5,9 @@ module.exports = {
   // https://github.com/facebook/jest/blob/v22.4.2/packages/babel-jest/src/index.js#L105-L147
   // And is required due to:
   // https://github.com/facebook/jest/issues/1468
-  // TODO: See if it would be easier to switch to the higher-level babel-jest,
-  // and wrap that instead.
+  // TODO: See if it would be easier to wrap the higher-level babel-jest instead.
   process(src, filename, config) {
     // Babel 7 returns null if the file was ignored.
-    return transform(
-      src,
-      Object.assign({}, { filename }, config.globals.BABEL_OPTIONS)
-    ) || src;
+    return transform(src, { filename, ...config.globals.BABEL_OPTIONS }) || src;
   }
 };
--- a/packages/jest/src/transformer.js
+++ b/packages/jest/src/transformer.js
@@ -1,9 +1,17 @@
-const babel = require('babel-core');
+const { transform } = require('@babel/core');
 
 module.exports = {
+  // This is inspired by:
+  // https://github.com/facebook/jest/blob/v22.4.2/packages/babel-jest/src/index.js#L105-L147
+  // And is required due to:
+  // https://github.com/facebook/jest/issues/1468
+  // TODO: See if it would be easier to switch to the higher-level babel-jest,
+  // and wrap that instead.
   process(src, filename, config) {
-    return babel.util.canCompile(filename) ?
-      babel.transform(src, Object.assign({}, { filename }, config.globals.BABEL_OPTIONS)).code :
-      src;
+    // Babel 7 returns null if the file was ignored.
+    return transform(
+      src,
+      Object.assign({}, { filename }, config.globals.BABEL_OPTIONS)
+    ) || src;
   }
 };
--- a/docs/cli/README.md
+++ b/docs/cli/README.md
@@ -88,8 +88,8 @@ index 3356802..d4d82ef 100644
    devtool: 'source-map',
    entry: {
      index: [
--      'babel-polyfill',
-       '/node/src/index.js'
+-      '/node/src/index.js'
++      '/node/src/app.js'
      ]
    },
 ```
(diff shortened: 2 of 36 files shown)


edmorley:
I'd be curious to know how both the dev start initial compile and the prod build times of this compared to `master`, for `taskcluster-{web,tools}`.

packages/jest/src/transformer.js
```diff
@@ -1,9 +1,17 @@
...
+  // TODO: See if it would be easier to switch to the higher-level babel-jest,
+  // and wrap that instead.
   process(src, filename, config) {
-    return babel.util.canCompile(filename) ?
-      babel.transform(src, Object.assign({}, { filename }, config.globals.BABEL_OPTIONS)).code :
-      src;
+    // Babel 7 returns null if the file was ignored.
+    return transform(
+      src,
+      Object.assign({}, { filename }, config.globals.BABEL_OPTIONS)
```
eliperelman: We dropped support for Node.js v6, right? Does that mean we can use rest/spread here instead of assign?

edmorley:
You're most welcome! The thanks are very much appreciated ❤️

eliperelman:
Thank you so much! We've got some interesting things planned, hopefully I'll get a PR out today.

packages/jest/src/transformer.js
```diff
@@ -1,9 +1,17 @@
-const babel = require('babel-core');
+const { transform } = require('@babel/core');
 
 module.exports = {
+  // This is inspired by:
+  // https://github.com/facebook/jest/blob/v22.4.2/packages/babel-jest/src/index.js#L105-L147
+  // And is required due to:
+  // https://github.com/facebook/jest/issues/1468
+  // TODO: See if it would be easier to switch to the higher-level babel-jest,
```
SimenB: You should be able to. If not, please open up an issue 🙂 

If you wanna hold off on that, you should make sure to return a sourcemap here, Jest handles them nowadays (should allow you to drop `retainLines`)

eliperelman:
@edmorley this still looks good, anything outstanding before this gets merged?

packages/jest/src/transformer.js
edmorley: Hi @SimenB - thank you for taking a look :-)

Yeah there's lots of potential for clean-up in our current Jest config!

The problem I was finding was that mentioned in facebook/jest#1468, in that the only way to pass the babel options would be to call babel-jest's `createTransformer()` in our own `process()` (since that's the earliest we have access to `config.globals`), which seems a little clunky - plus means we can't use babel-jest's `getCacheKey()` implementation and still have to duplicate `canInstrument: true` etc.

eg:
```js
// transformer.js
const { createTransformer } = require('babel-jest')

module.exports = {
  canInstrument: true,
  process(src, filename, config, transformOptions) {
    return createTransformer(config.globals.BABEL_OPTIONS).process(src, filename, config, transformOptions);
};
```

Though perhaps a better way would be to circumvent `config.globals` and pass our babel config via `process.env` like this?
```js
// transformer.js
const { createTransformer } = require('babel-jest')
const babelConfig = JSON.parse(process.env.BABEL_OPTIONS);

module.exports = createTransformer(babelConfig);
```

Or is there a simpler way I'm missing? (Having to use a custom transformer seems pretty heavyweight for just wanting to not use a `.babelrc`, but I realise we're likely an edge-case :-))

edmorley:
@eliperelman I think this is fine to merge now - there's a lot of cleanup we can do on the Jest side, but it's mostly pre-existing and I'd like to keep the scope as small as possible here.

packages/jest/src/transformer.js
edmorley: I've filed #851 for making these changes.

Issue #382, opened 2017-10-27
Reducing the potential for confusion if configuring ESLint outside of Neutrino


edmorley:
### Summary

As seen in #380, users sometimes don't follow the docs recommendations, and store eslint configuration outside of Neutrino. As part of this, they may also override `neutrino-middleware-eslint`'s default of `useEslintrc: false`, which can cause a few problems.

### Background

For linting, the Neutrino docs recommended:
* Keeping all linting configuration in `.neutrinorc.js` (or middleware called from there).
* Favouring `neutrino lint` over calling eslint manually.
* For cases where manual eslint usage is required (eg editor/IDE support), creating an `.eslintrc.js` which does nothing more than call the Neutrino API to generate a config that's equivalent to the one used by `neutrino lint`.

There are then three ways linting is used:
1. Via the `neutrino lint` command
    - This uses the ESLint API (specifically [CLIEngine.executeOnFiles()](https://eslint.org/docs/developer-guide/nodejs-api#executeonfiles)).
2. Via the `neutrino {build,start,...}` commands
    - These use [eslint-loader](https://github.com/MoOx/eslint-loader) to hook up linting to the webpack build, which itself uses the ESLint API (specifically [CLIEngine.executeOnText()](https://eslint.org/docs/developer-guide/nodejs-api#executeontext)).
3. Via running eslint from the CLI or via an editor/IDE plugin
    - These load the `.eslintrc*` files (of which there are several supported formats/filenames: [docs](https://eslint.org/docs/user-guide/configuring#configuration-file-formats)).
    - If there is an `.eslintrc.js` and it's set up as recommended by the Neutrino docs, it exports a mostly equivalent config (with some keys filtered out, since the RC format differs slightly from that accepted by `CLIEngine`).

### Alternative approaches users might take

For users that don't follow the approach recommended by the docs, I can think of a few possible variations they might use:

#### (A) Using an alternate ESLint RC file and `useEslintrc: true`:

For example (from #380)...

```js
// .neutrinorc.js
module.exports = {
    options: {
        output: 'dist'
    },
    use: [
        ['neutrino-middleware-eslint', { eslint: { useEslintrc: true } }],
        ...
    ]
}
```

```js
// .eslintrc (Note: Doesn't have a `.js` extension)
{
    "extends": "eslint:recommended",
    "parserOptions": {
        "sourceType": "script"
    },
    // ...
}
```

The problem with this approach is that:
[...]

PR #852, opened 2018-05-07
Morph Neutrino API and CLI into middleware injectors for external CLI tools

### LINKED ISSUE #708: Neutrino lint and ESLint return different results
Hello,

I'm trying to figure out what's going on here and have been stuck. It seems that when running `neutrino lint` vs `eslint` directly, results differ.

I have created a minimal repro [here](https://github.com/gir/neutrino-lint-test). I am unsure if the problem is with the specific usage of Webpack module resolve (perhaps I use it incorrectly?), a neutrino setup oversight, or a bug. I have quickly tried to trace it through some of the Neutrino source, but don't see anything that jumps out. I also had a look at #380 and #382 but may have missed something.

Essentially I have added `src` as a module directory to the Webpack config so that I can do something like `import foo from 'foo'` (note not `'./foo'`). Then I add the directory to `import/resolver` through ESLint settings so the airbnb-configured `import/no-unresolved` should not fail.

`yarn lint` which runs `neutrino lint` returns errors.
`yarn eslint src --ext .js,.jsx` returns no errors.

I would expect the two runs to return the same results.

Versions:
```
neutrino v8.1.1
node v9.5.0
yarn v1.3.2

@neutrinojs/airbnb v8.1.1
@neutrinojs/jest v8.1.1
@neutrinojs/react v8.1.1
```

### LINKED ISSUE #736: --watch isn't being handled by the mocha preset
I'm currently using "./node_modules/mocha/bin/mocha ./test/**/*.test.js --watch" instead of "neutrino test ./test/**/*.test.js --watch".

Preset referred:
https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/mocha/src/mocha.js

### LINKED ISSUE #773: .neutrinorc.js file clobbers --use option
https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/neutrino/bin/neutrino.js#L88-L89

Was attempting this (with a different `.neutrinorc.js` in cwd: `neutrino build --use .neutrinorc-vue.js`

Config from `.neutrinorc.js` is used when it should be ignored (because of `--use`).

### LINKED ISSUE #839: `options.env` does not set `process.env` values
Bug / `neutrino@8.1.2`

`options.env` is supposed to set environment variables (process.env), but it does not.

It correctly merges default cli options with passed options, but since the `process.env` vals are only set in the constructor ([via `getOptions`](https://github.com/mozilla-neutrino/neutrino-dev/blob/cc412d0c18dbaa66bceb011ed2f4185b2dbad8d2/packages/neutrino/src/api.js#L22)), they only ever get set once with the default cli options.

### LINKED ISSUE #842: neutrino test succeeds even when no test command is registered
**STR:**
1. Create a new React project using `create-project`, but pick "none" for the test preset
2. `yarn test`

**Expected:**
Command exits with non-zero exit code and message about no registered command.

**Actual:**
```
test-react-project $ yarn test
yarn run v1.6.0
$ neutrino test
Done in 2.75s.
test-react-project $ echo $?
0
```

This is using Neutrino from `master`.

### LINKED ISSUE #849: Document that adding webpack to package.json is not recommended
In #830 a Neutrino v8 project had a direct dependency on `webpack` v4, which caused confusing errors, since the `neutrino build` ended up using a combination of webpack 3 from `node_modules/neutrino/node_modules/webpack/` and webpack 4 from `node_modules/webpack/`.

Now this top-level webpack dependency was being used in `.neutrinorc.js`, so I don't think this issue would have been helped by adjusting the modules resolution order (ie #822).

Instead we should document that:
* in general, only neutrino packages should be added to a project's `package.json` (unless directly required in `.neutrinorc.js`)
* in cases where it is required, care must be taken to keep these in sync with the versions used in Neutrino

CC @mrwillis.

### LINKED ISSUE #870: Setting 'config.stats' to a named preset doesn't work
The webpack `config.stats` property supports a number of named configuration presets such as `"verbose"`, `"minimal"` etc:
https://webpack.js.org/configuration/stats/#stats

However setting these in Neutrino has no effect when running `neutrino build`:
```js
    (neutrino) => {
      neutrino.config.stats('verbose');
    },
```

This is due to the way the options are passed here (removing the other settings and passing just the preset name works):
https://github.com/mozilla-neutrino/neutrino-dev/blob/2616583a839812479b63faadb0ed072f75ab3527/packages/neutrino/bin/build.js#L34-L40

This would be another issue fixed by using webpack's CLI instead (#852).

eliperelman:
__Edited to reflect current state of PR.__

Fixes #708.
Fixes #736.
Fixes #870.
Fixes #842.
Closes #773.
Closes #839.
Closes #849.

Outstanding issues to file:

- [x] Update docs to reflect these API changes (#896)
- [x] ~~Determine need for config-loading bins instead of forcing config file boilerplate (still unsure if we should do this because of IDE integration)~~
- [x] Modify stats to produce friendlier build output (#897)

This patch is a work-in-progress, and is quite the overhaul of Neutrino into something different. Right now there is no change to the documentation, which I will change if we decide to move forward with this.

This patch morphs Neutrino into a "preset engine" for transforming managed webpack configuration into a format that can be consumed by other CLI tools. No longer would we wrap external CLI tools like webpack, eslint, jest, karma, mocha, stylelint, etc. Instead, Neutrino could augment these tools by exposing methods for returning usable configuration.

This approach gives the user control of their CLI tool back, and they can programmatically invoke Neutrino to load their configuration, and override however they like:

```bash
webpack --mode production
```

```js
// webpack.config.js

// Load with no config overrides:

// Load with config overrides:
module.exports = neutrino().webpack(config => {
  // return whatever config you want
});

// Load with no config overrides using low-level API:
module.exports = neutrino().output('webpack');

// Load with config overrides using low-level API:
module.exports = neutrino().output('webpack', config => {
  // return whatever config you want
});
```

The same works for ESLint and friends too (except for mocha, since it has no config file we can be required from):

```bash
eslint src
```

```js
// .eslintrc.js

// Load with no config overrides:
module.exports = neutrino().eslintrc();

// Load with config overrides:
module.exports = neutrino().eslintrc(config => {
  // return whatever config you want, like manipulating rules, etc.
});
```

---

- This approach is predicated on having middleware defined in `.neutrinorc.js`, which shouldn't change at all.
- ~~This does do away with a bunch of CLI functionality like `--inspect` or `--use`, but those become less useful if we are just an interface to external CLI tools. We can discuss potentially bringing those back.~~ This is now back in the form of the CLI with `neutrino --inspect`.
- ~~All environment variables are gone. We rely only on webpack's `mode` value.~~ We internally only use webpack's `mode` value, but still set `process.env.NODE_ENV` if not set based on the mode, or defaulted to `production`.
- ~~There is no longer a create-project tool, since we cannot force users to choose between competing tools like webpack-cli or webpack-command, and webpack-dev-server or webpack-serve.~~ The create-project tool is back, and we opt for `webpack-cli` and `webpack-dev-server` as opinions for now.

I'm sure there are other major things I am neglecting to mention. Let's discuss! I see so many benefits to this, including reduced maintenance burden and the ability to allow people to use Neutrino without dumping their existing tools.

cc: @mozilla-neutrino/core-contributors


eliperelman:
Ah, one thing I forgot to mention is that you can override configuration still using neutrinorc with the webpack-chain instance. This means you can now override using whichever method you prefer:

```js
// Use neutrinorc for deterministic configurability with
// the chainable API
module.exports = {
  use: [
    '@neutrinojs/react',
    neutrino => {
      neutrino.config.devtool('source-map');
    }
  ]
};
```

```js
// Override from the configuration file of your CLI tool.
// This requires that you switch away from using the internal
// Neutrino CLI loader to making the tool use its own config file
// again, e.g.
// from:
// neutrino webpack --mode production
// to:
// webpack --mode production

// webpack.config.js

module.exports = neutrino().webpack(config => ({
  ...config,
  devtool: 'source-map'
});
```

timkelty:
Hope to review this week, but I'm a big fan of moving in this direction!

SeeThruHead:
Absolutely love this.

edmorley:
Many thanks for working on this - as you know I'm keen to try and make it easier to use native tools with Neutrino / reduce the cost of maintenance on our side / possibly reduce the overhead of Neutrino in startup time :-)

I tested this by comparing `master` and this PR via the following steps:
1. From monorepo root: `rm -rf node_modules packages/*/node_modules && yarn && yarn link:all`
2. `cd .. && create-project test-react-project`
3. Give the following answers: Web -> React -> Jest -> AirBnb
4. `cd test-<foo>` 
5. Edit the generated `package.json` to remove `neutrino` and `@neutrinojs/*`
6. `rm -rf node_modules yarn.lock && yarn`
7. `yarn link neutrino @neutrinojs/react @neutrinojs/jest @neutrinojs/airbnb`
8. Run `yarn build` 5 times and take the quickest invocation according to `time`

On master I get:
```
$ time yarn build
yarn run v1.6.0
$ neutrino build
√ Building project completed
Hash: 644638fa61693d1f28fe
Version: webpack 4.8.0
Time: 3422ms
Built at: 2018-05-07 21:16:03
                          Asset       Size  Chunks             Chunk Names
runtime.a5d5684f37863467cd6d.js   1.04 KiB       0  [emitted]  runtime
      1.f5d8aa6c2a2a15538a95.js    106 KiB       1  [emitted]
 index.96536b326bb3b272ec66.css   18 bytes       2  [emitted]  index
  index.0705adefdc1c7579150f.js  562 bytes       2  [emitted]  index
                     index.html  565 bytes          [emitted]
Entrypoint index = runtime.a5d5684f37863467cd6d.js 1.f5d8aa6c2a2a15538a95.js index.96536b326bb3b272ec66.css index.0705adefdc1c7579150f.js
Done in 7.73s.

real    0m8.294s
user    0m0.000s
sys     0m0.015s
```

Trying with this PR I got:
```
$ yarn build
yarn run v1.6.0
$ neutrino webpack --mode production
events.js:165
      throw er; // Unhandled 'error' event
      ^

Error: spawn webpack ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:201:19)
    at onErrorNT (internal/child_process.js:379:16)
```

I presume this is due to needing `child_process.spawn()` changes for compatibility with Windows. I tried adding `shell: true`, which got further, but still failed with `Error: Cannot find module '.neutrinorc.js'`, so I skipped this approach for now.

I instead created a `webpack.config.js` like so:
```js
```

And then tried using webpack directly:
```
$ time yarn webpack --mode production
yarn run v1.6.0
$ C:\Users\Ed\src\test-react-new\node_modules\.bin\webpack --mode production
Hash: 6d910272a91fdf310949
Version: webpack 4.8.0
Time: 2891ms
Built at: 2018-05-07 22:01:05
                          Asset       Size  Chunks             Chunk Names
runtime.d3f7781965753ba20f1b.js   1.04 KiB       0  [emitted]  runtime
      1.083d16c533c13131a3f5.js    106 KiB       1  [emitted]
                      index.css   18 bytes       2  [emitted]  index
  index.e329700ad1ece9219d57.js   1.04 KiB       2  [emitted]  index
                     index.html  548 bytes          [emitted]
Entrypoint index = runtime.d3f7781965753ba20f1b.js 1.083d16c533c13131a3f5.js index.css index.e329700ad1ece9219d57.js
 [6] ./src/App.css 39 bytes {2} [built]
 [7] ./src/App.jsx 1.01 KiB {2} [built]
[19] ./src/index.jsx 635 bytes {2} [built]
[20] multi ./src/index 28 bytes {2} [built]
    + 17 hidden modules
Child html-webpack-plugin for "index.html":
     1 asset
    Entrypoint undefined = index.html
    [0] (webpack)/buildin/module.js 497 bytes {0} [built]
    [1] (webpack)/buildin/global.js 489 bytes {0} [built]
        + 2 hidden modules
Child mini-css-extract-plugin ../neutrino-dev/node_modules/css-loader/index.js??ref--7-1!src/App.css:
    Entrypoint mini-css-extract-plugin = *
    [1] ../neutrino-dev/node_modules/css-loader??ref--7-1!./src/App.css 204 bytes {0} [built]
        + 1 hidden module
Done in 6.14s.

real    0m6.677s
user    0m0.000s
sys     0m0.015s
```

It looks like the configuration differs slightly:
* `index.css` no longer has a hashed filename
* `index.js` is larger
* the stats output has changed

...however the build time reduced by 1.5s, which is promising (though might just be the different build options?).

Some other things I spotted:
* `create-project`'s initial lint didn't complete (the `yarn webpack --mode production` is after I fixed the issues)
* I had problems getting the create-project's generated `.eslintrc.js` to work - running eslint failed with:
```
$ yarn eslint --ext=js,jsx,vue,ts,tsx,mjs src --fix
yarn run v1.6.0
$ C:\Users\Ed\src\test-react-new\node_modules\.bin\eslint --ext=js,jsx,vue,ts,tsx,mjs src --fix
Cannot read config file: C:\Users\Ed\src\test-react-new\.eslintrc.js
Error: Cannot find module '.neutrinorc'
Error: Cannot read config file: C:\Users\Ed\src\test-react-new\.eslintrc.js
Error: Cannot find module '.neutrinorc'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:548:15)
```
* when we get to the point of adding more polish to this approach, it would be good to improve the error message for people whose `yarn build` command still tries to run `neutrino build`, since it currently fails with the confusing:
```
$ yarn build
yarn run v1.6.0
$ neutrino build
C:\Users\Ed\src\neutrino-dev\packages\neutrino\bin\neutrino.js:16
spawn(command, [...args, ...commands[command]], {
                                    ^

TypeError: commands[command] is not iterable
    at Object.<anonymous> (C:\Users\Ed\src\neutrino-dev\packages\neutrino\bin\neutrino.js:16:37)
```

edmorley:
> This is similar to how nyc works, by calling a Neutrino CLI which just loads middleware and then spawns the given command passing in a config file.

Could this part be avoided by instead defining the `package.json` build script as:
`webpack --mode production --config node_modules/neutrino/src/foo.js`

I know passing a `node_modules` path isn't pretty, but given `neutrino` is a top-level dependency, we don't need to sorry about hoisting, right? I'm just wondering if we can get away with keeping the number of API usage permutations to a minimum, to reduce complexity/maintenance/potential for bugs.

Also, what are your thoughts on starting to use `.babelrc.js` with this approach too? (Or the new `babel.config.js` which avoids some of `.babelrc*`'s issues; see babel/babel#7358.) It would seem that many tools "just work" when a babelrc is preset (eg babel-jest) - so it might simplify several other presets.

edmorley:
> It looks like the configuration differs slightly:
> * index.css no longer has a hashed filename
> * index.js is larger

Ah the former is because `packages/style-loader/index.js` uses `neutrino.config.mode === 'production'` rather than `neutrino.config.get('mode') === 'production'`.

The latter is because the Jest preset's Babel settings are being applied even for `.webpack()`, and `plugin-transform-modules-commonjs` increases the output size.

eliperelman:
> Could this part be avoided by instead defining the `package.json` build script as:
`webpack --mode production --config node_modules/neutrino/src/foo.js`

Yes, that is certainly possible. If we are fine with a little loss in ergonomics here, we can move forward with it.

> Also, what are your thoughts on starting to use `.babelrc.js` with this approach too?

This should be totally possible now, just need to add a `babel()` method.

> Ah the former is because `packages/style-loader/index.js` uses `neutrino.config.mode === 'production'` rather than `neutrino.config.get('mode') === 'production'`.

Good catch, yay bugs!

> The latter is because the Jest preset's Babel settings are being applied even for `.webpack()`, and `plugin-transform-modules-commonjs` increases the output size.

🤔 Hmm yeah, let me think about this one.

eliperelman:
With my current changeset, the only way to affect configuration is via `.neutrinorc.js` or an override in a respective config file, e.g. `webpack.config.js` or `.eslintrc.js`. I took away the ability to do `webpack node_modules/neutrino/whatever` for now, so you need a `webpack.config.js` with:

```js

```

.eslintignore
```diff
@@ -1,4 +1,6 @@
 packages/**/node_modules/
+# Remove ignore of neutrinorc since ESLint ignores dotfiles by default
+!.neutrinorc.js
 # TODO: Fix lint errors in our tests and remove this line
 packages/**/test/
 # The templates have any lint errors fixed during project creation,
```
edmorley: I can't comment on the exact line, because it's too far our of the diff context - but there's a create-project entry here that can be removed :-)

packages/airbnb-base/package.json
```diff
@@ -24,11 +24,11 @@
   },
   "dependencies": {
     "@neutrinojs/eslint": "^8.2.0",
-    "eslint": "^4.19.1",
     "eslint-config-airbnb-base": "^12.1.0",
     "eslint-plugin-import": "^2.11.0"
   },
   "peerDependencies": {
+    "eslint": "^4.0.0",
```
edmorley: The `@neutrinojs/eslint` package will already have a `peerDependency` on eslint, which will be inherited. Though might be clearer for users inspecting only the top-level package.json to be explicit? I can't decide. (The same also applies to eg the `webpack` peerDep in `@neutrinojs/font-loader` since file-loader has it's own peerDep etc)

packages/eslint/index.js
```diff
@@ -1,46 +1,15 @@
...
-  const destinationRules = pathOr({}, ['eslint', 'rules'], destination);
+  const sourceRules = (source && source.eslint && source.eslint.rules) || {};
+  const destinationRules = (
+    destination &&
+    destination.eslint &&
+    destination.eslint.rules
+  ) || {};
   const rules = deepmerge(sourceRules, destinationRules, {
     arrayMerge(source, destination) {
       return destination;
```
edmorley: I can't select the line, since it's further down below - but there's a leftover `neutrino.options.command !== 'start'`.

I wonder if `command` could be renamed to something else so third-party middleware / end-users' `.neutrinorc.js` conditionals fail hard rather than silently not matching?

edmorley:
> I think we might still need some inspect functionality (eg looking for argv.inspect)

Oh and the debug functionality (for me, main use I have had for it is preset-env's debug output).

packages/airbnb-base/package.json
eliperelman: Yeah, I added this since I figured it would be easier to ensure users would be notified from their top-level peer dependencies, but I am fine with removing some of them if they are bloating.

edmorley: I think given we're now needing users to add a bunch of peerDeps perhaps we need to have it as clear as possible to them :-)

edmorley:
I wonder if we should also make `webpack()` and friends throw if `mode` is not set? (ie if the user forgets to `--mode ...`)

ringods:
@eliperelman testing my `multi-compiler` question from #876 against this rework.

With the new setup, would I have something like this in my `webpack.config.js`?
```

module.exports = neutrino().use('my-preset-generating-multi-compiler')
```

What would the api be then for a preset to return a list of webpack configurations?
I'm not understanding it completely at the moment.

packages/jest/src/index.js
```diff
@@ -86,48 +12,31 @@ module.exports = (neutrino, opts = {}) => {
...
-
-    return jest
-      .runCLI(cliOptions, options.roots || [options.rootDir])
-      .then(({ results }) => results.success ? Promise.resolve() : Promise.reject());
   });
+
+  neutrino.config.when(
+    neutrino.config.module.rules.has('compile') &&
+    neutrino.options.command === 'jest',
+    () => {
```
edmorley: Given that the false case isn't used, and we're not using a passed `(config)`, this might be cleaner as:
```js
  if (
    neutrino.config.module.rules.has("compile") &&
    neutrino.options.command === "jest"
  ) {
    // ...
  }
```

packages/mocha/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@neutrinojs/mocha",
   "version": "8.2.0",
   "description": "Neutrino preset for testing Neutrino projects with Mocha",
-  "main": "src/index.js",
```
edmorley: Similar to the jest preset case, could the file be left under `src/` for now, so that git knows it's existing content and displays the actual diff? (Happy for the file to be moved either in a PR before this lands, or sometime after)

packages/neutrino/karma.js
```diff
@@ -0,0 +1,68 @@
...
+const { join } = require('path');
+
+module.exports = adapter => config => {
+  const { neutrino } = adapter;
+
+  adapter
+    .command('karma')
+    .init();
+
+  if (neutrino.config.module.rules.has('compile')) {
```
edmorley: It feels like these really belong in the preset rather than in the Neutrino package. Could we instead maybe do something like either:
* make these files be mostly a shell that `require()`s the relevant parts from the preset
* have the presets dynamically register new commands, so we don't need to hardcode each test runner in the neutrino package?

eliperelman: Hmm, I hadn't thought of it this way before, and it would take re-working the adapter pattern, but let me see what I can do.

eliperelman:
@ringods I think the idea would be that you could export multiple Neutrino calls:

```js
// webpack.config.js

module.exports = [
  neutrino().webpack(),
  neutrino(middleware).webpack()
];
```

edmorley:
To add to the commit message:

Fixes #708.
Fixes #736.
Fixes #870.
Closes #773.
Closes #849.

We should also check whether it fixes any of:
#842 (whilst Neutrino test won't exist, we should check for say `.mocha()` when no mocha preset)
#839
#382 (though I suspect this one will need extra changes; notably shifting to making eslint-loader use eslintrc as the source of truth)

eliperelman:
Updated description to reflect removal of adapter pattern, and now using:

```js
neutrino().output(name, override);

neutrino().output('webpack');
neutrino().output('webpack', config => {
  return config;
});
```

packages/react/package.json
```diff
@@ -31,7 +31,6 @@
     "@neutrinojs/loader-merge": "^8.2.0",
     "@neutrinojs/web": "^8.2.0",
     "deepmerge": "^1.5.2",
-    "eslint": "^4.19.1",
```
edmorley: So I guess now if people don't add eslint to their `package.json` they will get peer dependency warnings from `eslint-plugin-react` (and similar for the vue/preact presets etc). However maybe that's ok and their choice to save the dependency?

The only option I can think of (other than leaving this dependency in here), is to stop the react preset from including the eslint react plugin, and then airbnb guide users can just use the airbnb preset (which does include the plugin), and for standardjs perhaps we'd have to add a second "standardjs-react" preset? (Maybe not worth it...)

packages/react-components/test/components_test.js
```diff
@@ -0,0 +1,51 @@
+import test from 'ava';
```
edmorley: Awesome for the extra tests!

packages/neutrino/test/api_test.js
```diff
@@ -133,135 +133,53 @@ test('override options.mains', t => {
...
+  api.use(api => {
+    api.config.module
+      .rule('compile')
+      .test(api.regexFromExtensions(['js']))
+  });
 
   t.notDeepEqual(api.config.toConfig(), {});
 });
 
-test('throws when trying to call() a non-registered command', t => {
```
edmorley: Do you think we should turn this into a test checking for the `Unable to find an output handler named` behaviour?

packages/neutrino/package.json
```diff
@@ -2,16 +2,11 @@
   "name": "neutrino",
   "version": "8.2.0",
   "description": "Create and build JS applications with managed configurations",
-  "main": "src/index.js",
-  "bin": {
-    "neutrino": "./bin/neutrino.js"
```
edmorley: Do you think we should leave a bin script that outputs something like "The neutrino command has been removed. See migration guide"?

packages/karma/package.json
```diff
@@ -22,19 +22,11 @@
...
+    "@neutrinojs/compile-loader": "^8.2.0",
     "@neutrinojs/loader-merge": "^8.2.0",
     "babel-plugin-istanbul": "^4.1.6",
     "deepmerge": "^1.5.2",
-    "karma": "^2.0.2",
-    "karma-chrome-launcher": "^2.2.0",
-    "karma-coverage": "^1.1.2",
-    "karma-mocha": "^1.3.0",
-    "karma-mocha-reporter": "^2.2.5",
-    "karma-webpack": "^3.0.0",
```
edmorley: I think some of these are still needed?

packages/standardjs/test/standardjs_test.js
```diff
@@ -9,37 +9,25 @@ test('loads preset', t => {
...
-test('exposes lint command', t => {
-  const api = Neutrino();
-
-  api.use(mw());
-
-  t.is(typeof api.commands.lint, 'function');
-});
-
-test('exposes eslintrc config', t => {
-  t.is(typeof Neutrino().use(mw()).call('eslintrc'), 'object');
```
edmorley: Should we convert this and the other "expose XYZ" to checking that `...output('...')` works?

package.json
```diff
@@ -34,17 +33,21 @@
   },
   "devDependencies": {
     "auto-changelog": "^1.4.6",
-    "ava": "^0.25.0",
+    "ava": "1.0.0-beta.4",
```
edmorley: Sneaking in more changes again hehe.. :-)

packages/create-project/bin/create-project.js
```diff
@@ -22,5 +23,6 @@ const name = basename(directory);
 env.run('create-project', {
   directory,
   name,
+  registry: cli.registry,
```
edmorley: Is this change so that the global yarn registry setting doesn't need to be adjusted? (And so can be removed from the script.) If so, this would be good since means running these tests don't affect global state locally.

packages/react/package.json
eliperelman: Hmm yeah, hadn't thought of this one. I think the easiest thing at the moment would be to leave it until we think of something better, if possible.

packages/neutrino/package.json
eliperelman: Works for me.

packages/create-project/bin/create-project.js
eliperelman: In theory, yes. You can either pass `--registry` during `create-project`, or you can specify the registry in the test generator (see the `project()` function in `test/cli_test`).

packages/web/test/web_test.js
```diff
@@ -66,52 +64,16 @@ test('valid preset development', t => {
...
 
 test('throws when minify.image defined', async t => {
-  const api = Neutrino();
+  const api = new Neutrino();
 
   const err = t.throws(() => api.use(require('..'), { minify: { image: true } }));
   t.true(err.message.includes('The minify.image option has been removed'));
 });
-
-test('throws when vendor entrypoint defined', async t => {
```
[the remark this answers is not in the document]
eliperelman: I moved this test to inside `neutrino` since this wasn't `web`'s concern.

packages/neutrino/test/api_test.js
eliperelman: Added this to the new `package_test`.

packages/karma/package.json
eliperelman: I have moved them to peer deps, although I'm not entirely happy about that.

package.json
eliperelman: Rebase artifact. I'll fix. 😄

eliperelman:
OK, I think have the latest round of review comments addressed.

packages/create-project/commands/init/index.js
```diff
@@ -68,18 +71,31 @@ module.exports = class Project extends Generator {
...
   _initialPackageJson() {
     const installer = isYarn ? 'yarn' : 'npm';
-    const scripts = { build: `${packages.NEUTRINO} build` };
+    const scripts = { build: 'webpack --mode production' };
 
     if (this.data.projectType !== 'library') {
-      scripts.start = `${packages.NEUTRINO} start`;
+      scripts.start = this.data.project === '@neutrinojs/node'
+        ? 'webpack --watch --mode development'
+        : 'webpack-serve --mode development';
```
edmorley: Running `yarn start` on a create-project with options React+Jest+AirBnb, I get:
```
$ yarn start
yarn run v1.6.0
(node:6040) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
$ webpack-serve --mode development
× ｢config｣: An error ocurred while trying to find a config file
              Did you forget to specify a --require?
C:\Users\Ed\src\test-project\node_modules\@webpack-contrib\config-loader\lib\load.js:84
    throw new LoadConfigError(e, configPath);
    ^

LoadConfigError: Cannot find module 'webpack-dev-server/client'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:571:15)
    at Function.require.resolve (C:\Users\Ed\src\test-project\node_modules\webpack-serve\node_modules\v8-compile-cache\v8-compile-cache.js:162:23)
    at config.entry.batch.entry (C:\Users\Ed\src\test-project\node_modules\@neutrinojs\web\index.js:235:43)
```

I guess we have no coverage of this at the moment? 

Perhaps we can do either (or both) of:
* test webpack build using `--mode development`
* see if we can spawn the dev server via `yarn start`, then if it starts successfully, kill it after a few seconds

Until then perhaps we need to hand-test `(using create-project and `--registry`) all of the top level preset's `yarn start` commands?

Also, I think the reason the unit tests didn't catch this is because the repo root `package.json` contains `webpack-dev-server` (when I think it perhaps shouldn't?).

packages/neutrino/src/devServer.js
```diff
@@ -1,11 +0,0 @@
-const merge = require('deepmerge');
-const Future = require('fluture');
-const { serve, validate } = require('./webpack');
-
-// devServer :: (Object config -> Object api) -> Future () Function
-const devServer = (config, { options }) => Future
-  .of(merge({ devServer: { host: 'localhost', port: 5000, noInfo: !options.debug } }, config))
```
edmorley: Does the `noInfo: !options.debug` need to move to `packages/dev-server/index.js` now? Or are we leaving it to the user to pass the options to webpack-serve? 

The latter seems to make sense, I was just wondering if with the current settings things are going to get verbose. I think after this PR we'll probably want to revisit our customisation of `stats` entirely, so perhaps this can wait until then?

packages/neutrino/bin/neutrino.js
```diff
@@ -1,130 +1,9 @@
...
-  .option('quiet', {
-    description: 'Disable console output of CLI commands',
-    boolean: true,
-    default: false,
-    global: true
-  })
-  .option('debug', {
-    description: 'Run in debug mode',
-    boolean: true,
-    default: false,
```
edmorley: With this PR `neutrino.options.debug` now defaults to `undefined` rather than `false`, which means `clean-webpack-plugin` now gives verbose output (its default) when it didn't before. I think we need to make it default to `false` again.

On that note, `process.env.NODE_ENV` is also `undefined` when it wasn't before. I think we need to set that explicitly again too, per:
https://github.com/mozilla-neutrino/neutrino-dev/pull/852#pullrequestreview-119629134

packages/eslint/package.json
```diff
@@ -26,15 +26,13 @@
...
     "eslint-loader": "^2.0.0",
     "eslint-plugin-babel": "^5.1.0",
-    "fluture": "^8.0.2",
     "lodash.clonedeep": "^4.5.0",
-    "ramda": "^0.25.0",
-    "webpack": "^4.7.0"
+    "lodash.omit": "^4.5.0"
   },
   "peerDependencies": {
+    "eslint": "^4.0.0",
```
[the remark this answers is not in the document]
eliperelman: Yeah, I guess so.

packages/jest/package.json
```diff
@@ -22,18 +22,16 @@
...
-    "deepmerge": "^1.5.2",
-    "eslint": "^4.19.1",
     "eslint-plugin-jest": "^21.15.1",
-    "jest-cli": "^22.4.3",
-    "ramda": "^0.25.0",
-    "yargs": "^11.0.0"
+    "lodash.omit": "^4.5.0"
   },
   "peerDependencies": {
+    "jest": "^22.0.0",
```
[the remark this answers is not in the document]
eliperelman: Hmm, this one feel kinda wrong since the addition of the linting is optional based on existing lint rule. Thoughts on leaving this one out?

packages/jest/package.json
```diff
@@ -22,18 +22,16 @@
     "yarn": ">=1.2.1"
   },
   "dependencies": {
-    "@babel/core": "^7.0.0-beta.46",
+    "@babel/core": "^7.0.0-beta.47",
```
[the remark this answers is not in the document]
eliperelman: Sure thing, this was from moving the `.jest()` command from `neutrino` package to `jest`.

packages/neutrino/bin/neutrino.js
```diff
@@ -1,130 +1,9 @@
-#!/usr/bin/env node
+#! /usr/bin/env node
```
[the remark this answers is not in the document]
eliperelman: TIL

packages/fork/README.md
```diff
@@ -1,240 +0,0 @@
-# Neutrino Fork Middleware
```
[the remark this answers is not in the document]
eliperelman: Yeah, going to do a follow-up PR for all docs once this and your docs PR lands.

packages/jest/package.json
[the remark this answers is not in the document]
edmorley: Oh yeah sorry we do the same for the react preset etc, so ignore this

packages/create-project/commands/init/index.js
```diff
@@ -41,7 +42,9 @@ module.exports = class Project extends Generator {
       use: [
         this.data.linter,
         this._getProjectMiddleware(),
-        this.data.testRunner
+        this.data.testRunner && this.data.testRunner.startsWith('@')
```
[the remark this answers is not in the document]
eliperelman: Ooooooo good catch! This is a left-over from when the karma and mocha presets were removed, but when I added them back I didn't remove this.

packages/fork/README.md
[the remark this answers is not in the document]
edmorley: I thought I'd already opened a PR for the new docs, but turns out I'd only pushed the branch previously 😆 

I'll rebase it (after the recent docs fixes) and open a PR tomorrow.

packages/create-project/package.json
```diff
@@ -40,6 +40,7 @@
     "deepmerge": "^1.5.2",
     "fs-extra": "^6.0.0",
     "javascript-stringify": "^1.6.0",
+    "neutrino": "^8.2.0",
```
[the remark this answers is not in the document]
eliperelman: Sorta, this was to fill in the `eslint --ext ${extensions}` section of the lint script, but I am OK with hardcoding it too.

packages/karma/package.json
```diff
@@ -22,21 +22,23 @@
...
-    "ramda": "^0.25.0",
-    "webpack": "^4.7.0"
+    "lodash.omit": "^4.5.0",
+    "mocha-coverage-reporter": "^0.0.1"
   },
   "peerDependencies": {
-    "neutrino": "^8.0.0"
+    "karma": "^2.0.0",
+    "karma-cli": "^1.0.0",
+    "mocha": "^5.0.0",
```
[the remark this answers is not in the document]
eliperelman: It was mostly to enable usage of karma and mocha presets in the same repo.

packages/mocha/package.json
```diff
@@ -20,13 +20,12 @@
...
   "dependencies": {
-    "@babel/core": "^7.0.0-beta.46",
-    "@babel/plugin-transform-modules-commonjs": "^7.0.0-beta.46",
-    "@babel/register": "^7.0.0-beta.46",
+    "@babel/core": "^7.0.0-beta.47",
+    "@babel/plugin-transform-modules-commonjs": "^7.0.0-beta.47",
+    "@babel/register": "^7.0.0-beta.47",
     "@neutrinojs/loader-merge": "^8.2.0",
-    "change-case": "^3.0.2",
-    "mocha": "^5.1.1",
```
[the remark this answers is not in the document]
eliperelman: Sigh, yes.

packages/create-project/package.json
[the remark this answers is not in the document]
edmorley: Oh right, makes sense. I wonder if it still might be worth just duplicating and adding a comment to both places saying to keep in sync?

packages/create-project/commands/init/index.js
eliperelman: Oooooooh, this is a problem. I don't think we can use webpack-serve out-of-the-box because of our hardcoded entries in `devServer`:

https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/web/index.js#L235

packages/neutrino/src/devServer.js
eliperelman: Yeah let's follow up with stats changes.

packages/create-project/commands/init/index.js
edmorley: Perhaps we should switch back to webpack-dev-server for now (also reduces the amount changed in one PR) and we can figure it out later?

packages/create-project/package.json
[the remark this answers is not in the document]
eliperelman: Yeah, will do.

eliperelman:
OK, I think the most recent review comments have all been addressed.

packages/create-project/commands/init/index.js
eliperelman: You read my mind.

ringods:
In comment https://github.com/mozilla-neutrino/neutrino-dev/pull/852#issuecomment-387222910, @edmorley writes:

> Also, what are your thoughts on starting to use .babelrc.js with this approach too? (Or the new babel.config.js which avoids some of .babelrc*'s issues; see babel/babel#7358.) It would seem that many tools "just work" when a babelrc is preset (eg babel-jest) - so it might simplify several other presets.

I hope you can also make it work when a neutrino preset author can provide the **it just works** behaviour by constructing the babel config (or any other tool) from the preset implementation. 

My reason to come to neutrino is that a project using a preset should build with the best practices in web development land implemented by the preset author, without requiring all the different type of configs that the underlying tool require.

My ideal use case for neutrino:

* `npm install neutrino neutrino-<some>-preset`
* Updating `scripts` in `package.json` to invoke integrated tool+neutrino
* Build your project

edmorley:
> I hope you can also make it work when a neutrino preset author can provide the it just works behaviour by constructing the babel config (or any other tool) from the preset implementation.

Ah to clarify, what the approach I was thinking of was:
* still have the babel config be generated by the presets
* but have it be made accessible via a `module.exports = neutrino().webpack()` in `.babelrc.js`
* and then things like `babel-jest` can just point use the `.babelrc.js` without needing the spaghetti here: https://github.com/mozilla-neutrino/neutrino-dev/blob/9e73636cb0fe66967452c049fe62a3ba1c52b4d1/packages/jest/src/index.js#L18-L76

I agree it would be good to reduce the number of steps to create these `module.exports = neutrino().foo()` configs -- and whilst create-project does that for new projects, perhaps we need a way to `neutrino init` existing ones? (create-project could even be made to use this new `neutrino init` itself?)

ringods:
> I agree it would be good to reduce the number of steps to create these `module.exports = neutrino().foo()` configs -- and whilst create-project does that for new projects, perhaps we need a way to `neutrino init` existing ones? (create-project could even be made to use this new `neutrino init` itself?)

@edmorley the way you talk about `neutrino init` feels like generating boilerplate files to me. Wasn't neutrino born to prevent boilerplate? The problem I have with boilerplate code is that it works to bootstrap a project, but that it breaks when best practices and tools evolve and you want to integrate that in existing projects started from boilerplate.

If I reflect that back to neutrino: at one point, as a user I install v1 of a neutrino preset. The preset author continues updating the preset middleware code and creates v2. As a user, the only thing I would need to do is update the preset package to v2 in my `package.json` file to have the new functionality out of the box.

Questions:
* should I have to re-`init`? 
* Are these generated files user-tweakable?
* Must the preset author take this into account at re-`init` time?

If one generates such files, I would prefer to not have these under source control then (and have it documented like that). The workflow would then start like this:

* git clone
* npm install
* run neutrino init (to generate the "standard" config files for cli tools, e.g. babel, jest)
* build your project.

Not having these files like `.babelrc.js` under source control prevents users trying to tweak them outside of neutrino.

Is this what and how you had it in mind?

edmorley:
I think it's a definite advantage to not try and re-invent the CLI of all of the various tools in Neutrino - as seen by the number of bugs this PR will fix. 

If we agree that approach is best, and that the webpack/eslint/... CLIs should be called directly, then it's a question of how do we support passing the config to those CLIs:

1. (As this PR does now) by having config files in the locations the CLIs automatically check (eg repo root  `webpack.config.js` that exports `neutrino().webpack()`), and have the package.json scripts not specify any paths (eg: `"build": "webpack --mode production"`)
2. Have a default config file in the neutrino package that contains the same as the file from (1), but whose path has to be hardcoded in package.json (eg: `"build": "webpack --config node_modules/neutrino/webpack.js"`)
3. Use some wrapper Neutrino script like we have at present, but that spawns the CLI tools rather than the Neutrino 8 approach of invoking them by their API

I am very tempted by option (2) (I actually brought it up in https://github.com/mozilla-neutrino/neutrino-dev/pull/852#issuecomment-387222910), since:
* as you say, it means fewer files to become stale / clutter up the repo + fewer steps for create-project to perform
* users who want to customise can always copy the default file to the repo root, thereby getting (1) for free.

...although the downsides to (2) is that:
* we'd still have to keep the `package.json` scripts up to date with the hardcoded config paths (albeit that's easier than updating the config files)
* users will pretty much need to always use the package.json scripts for anything that refers to the config, otherwise they'll have a lot of typing (not a problem for me, since I think they are preferable anyway)
* for non-standard output commands (eg anything that isn't the webpack config), the hardcoded path will need to be to the particular preset rather than to Neutrino - which might be confusing to the user (eg `"lint": "eslint <options> --config node_modules/@neutrinojs/eslint/eslintrc.js"`)
* if the user has a non-standard node_modules location, or hoisting doesn't work as expected, then the config might not exist at hardcoded location
* certain tools will need the individual config files anyway (eg linting support in IDEs)

Perhaps we can default to (2), but say for any advanced use-case (eg non-standard node_modules/...) the individual config files have to be used?

ringods:
@edmorley to keep the scripts in `package.json` general, something like [nps](https://www.npmjs.com/package/nps) could come in handy.

packages/neutrino/index.js
```diff
@@ -0,0 +1,49 @@
...
+  neutrino.config.mode(mode);
+  neutrino.register('webpack', webpack);
+
+  if (middleware) {
+    neutrino.use(middleware);
+  }
+
+  const adapter = {
+    output(name, override = IDENTITY) {
+      if (name === 'inspect') {
```
edmorley: I think we might want to adjust the way inspect works - at the moment it means changing eg `webpack,config.js` to use `neutrino().inspect()` -- but then that doesn't take into account mode etc.

Other ways might be:
* keep a neutrino bin script for running inspect
* allow users to add `options.inspect: true` to their `.neutrinorc.js` and then the config can be output regardless of how in gets called

For me, inspect is a pretty important feature, so it's something we'll need to improve before we release - but happy to save it to a later PR if prefered (even though it's technically a regression, and one that makes it harder to review/debug PRs).

eliperelman: Yeah, I'm not sure how we want to tackle this, but maybe we can use the neutrino CLI for this.

eliperelman: Added `--inspect` only to the CLI for this.

eliperelman:
Dev server should now be working.

packages/create-project/commands/init/index.js
```diff
@@ -68,18 +68,36 @@ module.exports = class Project extends Generator {
...
 
-    if (this.data.linter) {
-      scripts.lint = `${packages.NEUTRINO} lint`;
-    }
+    // The list of extensions here needs to be kept in sync with the
+    // extension list defined by neutrino/extensions.source. Modifying a value
+    // here should have an accompanying change there as well. We can't pull
+    // in neutrino here as that would potentially give us conflicting versions
+    // in node_modules.
+    const lint = `eslint --ext js,jsx,vue,ts,tsx,mjs src`;
```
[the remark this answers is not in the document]
edmorley: Yes, that rule enforces that template literals only be used when needed, otherwise use single quotes.

package.json
```diff
@@ -29,7 +29,7 @@
     "validate:eslintrc:airbnb-base": "eslint --no-eslintrc --print-config . -c ./packages/airbnb-base/eslintrc.js > /dev/null",
     "validate:eslintrc:standardjs": "eslint --no-eslintrc --print-config . -c ./packages/standardjs/eslintrc.js > /dev/null",
     "validate:eslintrc": "yarn validate:eslintrc:eslint && yarn validate:eslintrc:airbnb-base && yarn validate:eslintrc:airbnb && yarn validate:eslintrc:standardjs && yarn validate:eslintrc:root",
-    "version": "test -v SKIP_CHANGELOG || yarn changelog --package && git add CHANGELOG.md"
+    "version": "[[ -n \"SKIP_CHANGELOG\" ]] || yarn changelog --package && git add CHANGELOG.md"
```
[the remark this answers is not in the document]
eliperelman: Oops yeah, this was just a dumb mistake. 😆

packages/web/index.js
```diff
@@ -220,19 +220,13 @@ module.exports = (neutrino, opts = {}) => {
...
           Object
             .keys(neutrino.options.mains)
             .forEach(key => {
               config
                 .entry(key)
                   .batch(entry => {
                     options.hotEntries.forEach(hotEntry => entry.prepend(hotEntry));
-                    entry
-                      .prepend(require.resolve('webpack/hot/dev-server'))
-                      .prepend(`${require.resolve('webpack-dev-server/client')}?${url}`);
```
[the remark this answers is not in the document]
eliperelman: Hmm possibly, I'll have to try it.

eliperelman:
> Running yarn start for the node preset I get an error:

I have no idea how this didn't happen before, but there was a trailing space in the Node preset for the start script extension:

```js
{ ext: '.js ' }
```

🤷‍♂️

packages/web/index.js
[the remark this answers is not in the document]
eliperelman: No, looks like we still need these entries. I wonder if we still should have `.prepend(require.resolve('webpack/hot/dev-server'))`?

For what it's worth, looks like the next major version of StartServerWebpackPlugin will add the Node.js hot entries automatically.

https://github.com/ericclemmons/start-server-webpack-plugin/pull/19

eliperelman:
Here's another good reason to set `NODE_ENV` prior to loading any middleware:

https://github.com/ericclemmons/start-server-webpack-plugin/pull/19/files#diff-a737110aee135d8c561170c25d1d0dd2R20

packages/web/index.js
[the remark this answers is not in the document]
edmorley: > I wonder if we still should have

Did it work when you tested it?

edmorley:
Think this is unresolved? (Can't see a corresponding change?)

> I get some missing peer dependency warnings when running rm -rf node_modules/ packages/*/node_modules/ && yarn in the monorepo root with this PR checked out:

packages/web/index.js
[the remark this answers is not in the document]
eliperelman: Yeah, my guess is using webpack adds these automatically when hot is set.

edmorley:
The Netlify build on `master` is failing with:
`File not found: /opt/build/repo/docs/packages/fork.md`
(https://app.netlify.com/sites/neutrinojs/deploys/5b05c3691f12b73cb85e935d)

Could you update `mkdocs.yml` for the fork preset removal?

PR #859, opened 2018-05-08
Use spread properties instead of Object.assign()/deepmerge

edmorley:
Now that we require Node 8.3+, we can use spread properties instead of:
* Object.assign()
* deepmerge in cases where we just need a shallow merge.

http://node.green/#ES2018-features-object-rest-spread-properties

### DIFF
--- a/docs/creating-presets.md
+++ b/docs/creating-presets.md
@@ -137,10 +137,11 @@ _Example:_
 
 ```js
 module.exports = (neutrino, opts = {}) => {
-  const options = Object.assign({
+  const options = {
     quiet: false,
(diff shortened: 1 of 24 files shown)


eliperelman:
I haven't yet looked through the code, but does this also account for `Object.assigns` we could eliminate as well?

edmorley:
Ah I forgot to check those too

edmorley:
Bah I used squash and merge and intended to update the commit message to reflect the new summary/description of this PR (ie `Object.assign()` too), but only updated the commit message body not the first line. Not worth force pushing to fix, but annoying it's not accurate 🙁.

PR #863, opened 2018-05-09
Improve regexFromExtensions

### LINKED ISSUE #855: Improve regexFromExtensions()
In several places [regexFromExtensions()](https://github.com/mozilla-neutrino/neutrino-dev/blob/cc412d0c18dbaa66bceb011ed2f4185b2dbad8d2/packages/neutrino/src/api.js#L156-L159) is used to generate the regex used to match files.

However looking at the output from `--inspect` I see a few things that could be improved:

1) If the extension list contains a `.` then it's not escaped. eg:
``neutrino.regexFromExtensions(neutrino.options.extensions.map(ext => `worker.${ext}`))``
...gives:
  `/\.(worker.js|worker.jsx|worker.vue|worker.ts|worker.tsx|worker.mjs)$/`
...rather than:
  `/\.(worker\.js|worker\.jsx|worker\.vue|worker\.ts|worker\.tsx|worker\.mjs)$/`
(though perhaps this is the caller's responsibility? plus ideally the `worker` part of the string would be before the brackets? maybe `regexFromExtensions` should accept a prefix param?)

2) If there is only one extension passed, the brackets are unnecessary. eg:
`neutrino.regexFromExtensions(['html'])`
...gives: `/\.(html)$/`
...rather than: `/\.html$/`

helfi92:
Resolves #855.

### DIFF
--- a/packages/neutrino/test/api_test.js
+++ b/packages/neutrino/test/api_test.js
@@ -258,11 +258,12 @@ test('throws when trying to validate config with non-existent entry point', asyn
   const [err] = await t.throws(result);
 
   t.true(err.includes(`Entry module not found: Error: Can't resolve './src'`));
-})
+});
 
-test('get a regex expression using regexFromExtensions', t => {
+test('regexFromExtensions', t => {
   const api = Neutrino();
 
   t.is(String(api.regexFromExtensions(['js'])), '/\\.js$/');
   t.is(String(api.regexFromExtensions(['js', 'css'])), '/\\.(js|css)$/');
+  t.is(String(api.regexFromExtensions(['worker.js', 'worker.jsx'])), '/\\.(worker\\.js|worker\\.jsx)$/');
 });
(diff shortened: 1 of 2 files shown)


packages/neutrino/test/api_test.js
```diff
@@ -258,4 +258,11 @@ test('throws when trying to validate config with non-existent entry point', asyn
   const [err] = await t.throws(result);
 
   t.true(err.includes(`Entry module not found: Error: Can't resolve './src'`));
+})
+
+test('get a regex expression using regexFromExtensions', t => {
+  const api = Neutrino();
+
+  t.is(String(api.regexFromExtensions(['js'])), '/\\.js$/');
+  t.is(String(api.regexFromExtensions(['js', 'css'])), '/\\.(js|css)$/');
```
edmorley: Could you make one of these also check the escaping of `.`? :-)

packages/neutrino/test/api_test.js
```diff
@@ -258,4 +258,11 @@ test('throws when trying to validate config with non-existent entry point', asyn
   const [err] = await t.throws(result);
 
   t.true(err.includes(`Entry module not found: Error: Can't resolve './src'`));
+})
+
+test('get a regex expression using regexFromExtensions', t => {
```
eliperelman: "regex expression" is redundant.

Issue #788, opened 2018-04-14
Enable the new eslint-plugin-babel 'babel/quotes' rule


edmorley:
`eslint-plugin-babel` v5 (which we updated to in #780) added a new `babel/quotes` rule, that we presumably will want to enable.

> `babel/quotes`: doesn't complain about JSX fragment shorthand syntax (`<>foo</>;`)

https://github.com/babel/eslint-plugin-babel#rules

patiboh:
Hi, I'll be happy to do this. This should be enabled in `.neutrinorc.js`, am I correct?

edmorley:
Hi! 

The rule would need to be added here:
https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/airbnb/index.js

patiboh:
Thanks, I’m relatively new at this :)
I’ve read the project’s development process, I’ve done the modification on my branch and ran `yarn validate:eslintrc` as well as `yarn test`. Is that sufficient testing for submitting my PR?

eliperelman:
@patiboh yes, feel free to submit!

edmorley:
Fixed by #874.

PR #874, opened 2018-05-14
Eslint babel quotes

patiboh:
Fix for issue #788

### DIFF
--- a/packages/airbnb-base/index.js
+++ b/packages/airbnb-base/index.js
@@ -10,11 +10,15 @@ module.exports = (neutrino, opts = {}) => {
         // handled by babel rules
         'new-cap': 'off',
         // handled by babel rules
+        'quotes': 'off',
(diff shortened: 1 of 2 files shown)


edmorley:
Sorry to be a pain, but after chatting to @eliperelman we think it might be best to also make the same change to the `airbnb-base` preset too, just to keep them in sync (and in case people using the Vue preset use `airbnb-base` instead).

As such, would you mind also updating `packages/airbnb/index.js`?

patiboh:
Sure , I'll do it :)

edmorley:
Ah yes that's the one - I pasted the wrong filename haha!

edmorley:
Many thanks for the PR!
This change will be in Neutrino 9 (when it is released) :-)

patiboh:
Great! :)

PR #884, opened 2018-05-18
Misc monorepo testing improvements

edmorley:
* Ensure the monorepo .neutrinorc.js is included in linting
* Remove unused `random` package.json lifecycle script
* Add --registry option to create-project 
* Remove unnecessary 'yarn config set registry' from `test-create-project-ci.sh`
* Fix `test-create-project-ci.sh` shebang whitespace
* Adjust settings used for `lerna publish` in CI
* Make verdaccio more resilient to NPM flakiness + prevent NPM published Neutrino packages from accidentally being used instead of the local versions (by disabling proxying for them)
* Remove neutrino packages from Travis yarn cache after testing  to prevent cache churn

These are some of the changes split out from #852, plus some others to ensure the verdaccio testing works even when used later on release branches. See individual commit messages for more details.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,6 @@
 packages/**/node_modules/
+# Remove ignore of neutrinorc since ESLint ignores dotfiles by default
+!.neutrinorc.js
 # TODO: Fix lint errors in our tests and remove this line
 packages/**/test/
 # The templates have any lint errors fixed during project creation,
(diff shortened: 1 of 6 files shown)


edmorley:
I ended up dropping the ava upgrade for now, since it broke one of the tests and it's not clear if it's an ava bug or something daft in the `master` Neutrino API. The upgrade should be trouble-free after #884, since the problematic test is being rewritten anyway.

Issue #830, opened 2018-04-28
Issue with IgnorePlugin


mrwillis:
### Bug or issue?
Bug

Please try to answer the following questions:

- What version of Neutrino are you using? 8.2.3
- Are you trying to use any presets? If so, which ones, and what versions? neutrinojs/react: "^8.2.3",
- Are you using the Yarn client or the npm client? What version? yarn 1.6.0
- What version of Node.js are you using? 8.11.1
- What operating system are you using? Ubuntu 64
- What did you do? yarn run build (neutrino build)
- What did you expect to happen? It would produce files
- What actually happened, contrary to your expectations? I get this error when including the IgnorePlugin like so in my `.neutrino.js` file:
```js
    neutrino => neutrino.config.plugin('ignore')
      .use(webpack.IgnorePlugin, [/^\.\/locale$/, /moment$/]),
```

```
yarn run build
yarn run v1.6.0
$ neutrino build
⠋ Building project/home/julian/Desktop/cika/cika-client/node_modules/webpack/lib/IgnorePlugin.js:62
                compiler.hooks.normalModuleFactory.tap("IgnorePlugin", nmf => {
                               ^

TypeError: Cannot read property 'normalModuleFactory' of undefined
    at IgnorePlugin.apply (/home/julian/Desktop/cika/cika-client/node_modules/webpack/lib/IgnorePlugin.js:62:18)
    at Compiler.apply (/home/julian/Desktop/cika/cika-client/node_modules/neutrino/node_modules/tapable/lib/Tapable.js:375:16)
    at webpack (/home/julian/Desktop/cika/cika-client/node_modules/neutrino/node_modules/webpack/lib/webpack.js:33:19)
    at MapAction.ApAction$resolved$mapper [as mapper] (/home/julian/Desktop/cika/cika-client/node_modules/fluture/index.js:770:71)
    at MapAction$resolved [as resolved] (/home/julian/Desktop/cika/cika-client/node_modules/fluture/index.js:781:28)
    at resolved (/home/julian/Desktop/cika/cika-client/node_modules/fluture/index.js:220:19)
    at Resolved._fork (/home/julian/Desktop/cika/cika-client/node_modules/fluture/index.js:645:3)
    at drain (/home/julian/Desktop/cika/cika-client/node_modules/fluture/index.js:277:25)
    at settle (/home/julian/Desktop/cika/cika-client/node_modules/fluture/index.js:206:15)
[...]

PR #889, opened 2018-05-19
[v8.x] Really prevent yarn "unmet peer dependency" warnings

edmorley:
The `release/v8` equivalent of #847, since it looks really sloppy to have lots of unmet peer dependency warnings when installing - and I think the warnings contribute to things like #830.

The missing peer dependencies were found by scripting the `yarn add` of `neutrino <SINGLE_PRESET>` for each separate preset/middleware, so that cases where the parent preset hides missing peer deps of the lower level packages could be found.

### DIFF
--- a/packages/babel-minify/package.json
+++ b/packages/babel-minify/package.json
@@ -26,9 +26,7 @@
     "yarn": ">=1.2.1"
   },
   "dependencies": {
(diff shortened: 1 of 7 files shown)


edmorley:
v8.3.0 has just been released, which includes this fix.

Issue #883, opened 2018-05-18
Try to be more explicit in what's (not) supported in presets/middleware


avaragado:
Issue created as a result of a [Spectrum thread](https://spectrum.chat/?t=6beb2d28-5027-4d69-a886-df4f489fe32c).

As someone trying out neutrino for the first time, I quickly hit a problem: it seems that the library preset doesn't support `babel-plugin-transform-object-rest-spread`.

The doc page for the library preset, https://neutrino.js.org/packages/library/, is vague about what's actually supported and not supported by the library. It says:

> Modern Babel compilation supporting ES modules, async functions, and dynamic imports

OK, it doesn't explicitly mention object rest/spread. But that's a feature certainly covered by "modern", and helpers like `create-react-app` support it.

The point is, I had to guess.

Presets should ideally be explicit about the features they support. Stating what is NOT supported is also extremely helpful in many cases.

AlanFoster:
I've ran in to the exact same problem as yourself @avaragado 

Tests passed, linting passed, but compilation failed. I've now updated my `.neutrinorc.js` file to configure babel:

```javascript
const loaderMerge = require("@neutrinojs/loader-merge");

module.exports = {
  use: [
    // ...,

    (neutrino) => {
      neutrino.use(loaderMerge('compile', 'babel'), {
        plugins: ['transform-object-rest-spread']
      });
    }
  ]
};
```

I've sent a PR to correct the example usage of `loaderMerge` now too https://github.com/mozilla-neutrino/neutrino-dev/pull/890

edmorley:
Hi! I agree the docs could be clearer here - there's no point having useful new ES6 features available and not people not realising - and conversely, it's frustrating if something you expect to work doesn't.

Looking at the `release/v8` branch, I see that for some reason `babel-plugin-transform-object-rest-spread` is only enabled when using the `@neutrinojs/react` or `@neutrinojs/preact` presets:
https://github.com/mozilla-neutrino/neutrino-dev/blob/v8.2.3/packages/react/index.js#L31
https://github.com/mozilla-neutrino/neutrino-dev/blob/v8.2.3/packages/preact/index.js#L24

...but not for eg `@neutrinojs/web` or `@neutrinojs/library`:
https://github.com/mozilla-neutrino/neutrino-dev/blob/v8.2.3/packages/web/index.js#L89-L102
[...]

PR #890, opened 2018-05-19
loader-merge: Fix usage of `object-rest-spread`

AlanFoster:
According to the current version of babel, the usage of the object rest spread plugin name should be prefixed with `transform`

```json
{
  "plugins": [
    ["transform-object-rest-spread", { "useBuiltIns": true }]
  ]
}
```

https://babeljs.io/docs/plugins/transform-object-rest-spread/#via-node-api

### DIFF
--- a/docs/migration-guide.md
+++ b/docs/migration-guide.md
@@ -350,7 +350,7 @@ neutrino.config.module
   .rule('compile')
   .loader('babel', 'babel-loader', {
     options: {
-      plugins: ['object-rest-spread']
+      plugins: ['your-babel-plugin']
     }
   });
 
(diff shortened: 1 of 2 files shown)


edmorley:
Also, the docs are in two places in the repo - once under `docs/packages/` and once under `packages/` (an unfortunate side-effect of our current hosting strategy; soon to be fixed). Could you update both locations? :-)

eliperelman:
@AlanFoster would you be willing to update this PR based on @edmorley's comments?

edmorley:
The failed Netlify build is because `master` was broken at the time of the PR, so just going to ignore that result and merge anyway. 

After this PR, I think we should also add object-rest-spread to the web preset on `release/v8` as well as backport this docs fix there - which should help further with #883.

PR #899, opened 2018-05-23
[v7.x] Serve docs using Netlify

edmorley:
Previously the v7 docs were only accessible via browsing the raw markdown on GitHub. After this PR merges, the `release/v7` branch will be activated as a Netlify branch subdomain, and so available at:
https://release-v7.neutrinojs.org

Whilst there probably aren't many people using the v7 docs, this acts as a trial run for what we'll need to do to the `release/v8` docs when the final version of Neutrino 9 is released.

Changes include:
* Adding a link to "Latest Documentation" in the sidebar.
* Adding a banner on the front page, explaining not the latest version. This isn't incredibly pretty, but I don't think it's worth spending time on, given that in the future (when Neutrino v10 supersedes v9) we'll be able to use the material theme's `admonition` extension.
* Updating page titles to include "v7".
* Updating the "edit this page" GitHub base URL to use `release/v7` instead of `master`.
* Adding a `robots.txt` to prevent search engines from directing people to legacy versions.

See the GitHub status checks list for a link to the deploy preview :-)


edmorley:
I've enabled the branch subdomain in Netlify, and it's working well:
https://release-v7.neutrinojs.org/

PR #902, opened 2018-05-23
Migrate react preset and create-project to properly use RHL v4

eliperelman:


edmorley:
(For reviewing reference) Migration guide:
https://github.com/gaearon/react-hot-loader#migrating-from-v3

packages/react/index.js
```diff
@@ -10,16 +10,21 @@ module.exports = (neutrino, opts = {}) => {
   const mode = neutrino.config.get('mode');
   const options = merge({
     hot: true,
-    hotEntries: [require.resolve('react-hot-loader/patch')],
```
edmorley: There are now no default `hotEntries` in any of the presets that inherit `web` (or in `web` itself). Should we now remove `hotEntries` from the web preset entirely?

packages/react/README.md
```diff
@@ -308,30 +310,20 @@ First, install `react-hot-loader` as a dependency, this **must** be React Hot Lo
...
-from `react-hot-loader`. The main file may be named `index.js` or `index.jsx`. The extension is resolved by webpack.
-- Wrap your top-level React component in the `AppContainer`.
-- Perform the application render in a reusable function for initial load and subsequent reloads.
-- Add the `hot` acceptance to call this function.
+Neutrino will then automatically load React Hot Loader then next time a build
+is performed. Next, you need to mark your root component as _hot-exported_.
 
-For example:
+For example, if your `src/index` main entry renders a root component from
+`App.jsx`, then the exported component needs to have a hot export:
```
edmorley: Should we add a link to the RHL docs?

edmorley (approved):
Yey for more things updated! 😃

packages/react/README.md
eliperelman: Yeah, that would be good.

packages/react/index.js
eliperelman: I'm thinking so.

packages/web/index.js
```diff
@@ -219,17 +218,6 @@ module.exports = (neutrino, opts = {}) => {
       config.devtool('cheap-module-eval-source-map');
       config.when(options.hot, () => {
         neutrino.use(hot);
-        config.when(options.hotEntries, (config) => {
```
edmorley: I think we should add a check for `hotEntries` so the user knows to clean up their config (and also why things aren't working if they rely on it) - like we do for the deprecated `options.minify` options above.

eliperelman: Good call.

eliperelman:
Updated PR to upgrade create-project's install of RHL to v4.

PR #910, opened 2018-05-25
Documentation cleanup

edmorley:
* Replaces `neutrino.js.org` URLs with `neutrinojs.org` (like #903 did for `release/v8`).
* Fixes broken URLs as well as those that HTTP 301 redirect (eg due to lack of trailing slashes) - as found by https://validator.w3.org/checklink
* Adds a new `scripts/build-docs.sh` that runs `mkdocs build` followed by a `sed` that turns the `neutrinojs.org` in `packages/*/README.md` (that are required so the READMEs work on NPM) into relative URLs that will point at the current branch's version of the docs.
* Removes redundant `CNAME` file.
* Makes the Neutrino logo be loaded from the Netlify hosted file, rather than from Github.


scripts/build-docs.sh
```diff
@@ -0,0 +1,14 @@
...
+# Switch to the root of the repository.
+cd "$(dirname "$(dirname "${BASH_SOURCE[0]}")")"
+
+mkdocs build
+
+# Some of the docs are generated from the individual package READMEs,
+# which have to use absolute URLs so that they work on the NPM package
+# pages. This converts them to site-root relative URLs, so that links
+# between packages target the current branch's version of the docs.
+find build/ -type f -name '*.html' -exec sed -i 's|https://neutrinojs\.org/|/|g' {} +
```
eliperelman: If I'm reading this line right, this says:

Find all `*.html` files in the build folder and replace all occurrences of `https://neutrinojs.org` with a `/`. Is that right?

edmorley: Yeah but with the trailing slash on the original -- it converts `https://neutrinojs.org/foo/bar/` to `/foo/bar/`.

eliperelman: Ah cool.

PR #917, opened 2018-05-29
create-project: Use specific major versions of dependencies

### LINKED ISSUE #911: create-project should install specific major versions of deps
Jest 23 just came out, and so create-project now tries to install it instead of Jest 22, which gives:
`warning " > @neutrinojs/jest@8.3.0-alpha.8f855f46" has incorrect peer dependency "jest@^22.0.0".`
(though most users won't see that, since it is only shown when using `--debug`)

The same thing is happening with RHL.

We should specify the major version in:
https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/create-project/commands/init/matrix.js

edmorley:
To help prevent breakage when newer versions of the dependencies are released, but the corresponding Neutrino preset hasn't yet been updated to support them. Ideally we'd have some way of auto-generating these, but for now this at least avoids future breakage.

I was hoping to pin the Neutrino monorepo packages too, but that would require a significant refactor, so that will have to wait until another time.

Fixes #911.

### DIFF
--- a/packages/create-project/commands/init/matrix.js
+++ b/packages/create-project/commands/init/matrix.js
@@ -1,6 +1,7 @@
 const LINTING = 'linting';
 const PROJECT = 'project';
 const TESTING = 'testing';
+
 const N = 'neutrino';
 const REACT = '@neutrinojs/react';
 const PREACT = '@neutrinojs/preact';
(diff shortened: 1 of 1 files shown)


edmorley:
PR updated to use `@^N`

PR #919, opened 2018-05-29
Speed and reliability improvements to Verdaccio configuration

edmorley:
1. Switch Verdaccio from the NPM registry to Yarn's
  -> For consistency with standard yarn usage & to hopefully avoid NPM registry specific infra issues in the future.
2. Tweak `timeout` / `max_fails`
  -> Since 10s might not be long enough to connect, and 20 retries before giving up means the broken jobs take ages to appear as complete in Travis.
3. Increase Verdaccio `maxage` to 10m
  -> So that Verdaccio doesn't try and refresh package metadata when we're only partway through through the ~6 minute CI run.
4. Disable Verdaccio's caching of tarballs
  -> Since it duplicates the local yarn cache and won't be kept once Verdaccio is restarted anyway.
5. Clean up Verdaccio config
  -> Add comments and remove unused publish permission.

This PR reduces the overall duration of each Travis job by about 1 minute - mainly due to the `maxage` change. 

See:
https://www.verdaccio.org/docs/en/uplinks.html
https://www.verdaccio.org/docs/en/packages.html

I also tried many other permutations of tmpfs for temp directory or yarn cache directory, but it gave virtually no improvement. Ultimately I think double digit yarn invocations during the create-project matrix is just slow due to there being no `yarn.lock` in the new projects, so each time it has to do a lot of module resolution / graph generation.

### DIFF
--- a/verdaccio.yml
+++ b/verdaccio.yml
@@ -2,8 +2,8 @@ store:
   memory:
     limit: 10000
 uplinks:
-  npmjs:
-    url: https://registry.npmjs.org/
+  yarnpkg:
+    url: https://registry.yarnpkg.com/
     # Make request fail sooner if NPM being flaky (default is 30s).
     timeout: 10s
     # Increase number of failed requests before disabling uplink (default 2).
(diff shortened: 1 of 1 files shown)


edmorley:
I've since morphed this PR to include more Verdaccio changes than just the registry URL change.

This PR now reduces the overall duration of each Travis job by about 1 minute - mainly due to the `maxage` change than anything else.

PR #922, opened 2018-05-31
Only test against Node 9 on master

edmorley:
Since it's very unlikely we're ever going to have a Node 9 only CI failure if Node 8 and Node 10 both pass. This means each PR/branch only uses 2 jobs rather than 3, out of our 5 concurrent job allowance - but yet we still have coverage once they are merged to `master`.

Node 9 will also no longer be supported after the end of June.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,11 @@ language: node_js
 cache: yarn
 node_js:
 - '8'
-- '9'
 - '10'
+jobs:
+  include:
+    - if: type != pull_request AND branch = master
+      node_js: '9'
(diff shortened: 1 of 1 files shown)


edmorley:
The Node 9 job was still scheduled. Might be a Travis bug - have emailed their support.

edmorley:
It turns out that for PRs, `branch` means "base branch", so the conditional also needs to check for `type` too. I've updated the PR and it now works as expected. I've also opened a PR against the Travis docs (travis-ci/docs-travis-ci-com#1873).

PR #926, opened 2018-06-01
Use default webpack module resolution

### LINKED ISSUE #822: Fix the resolve.modules directory order
After several hours of debugging I've managed to work out why the Jest 22 upgrade PR was failing when Jest was used with Preact, but not when used with React.

When reproducing locally (inside a project created with `create-project`), the failure looked like:
```
test-preact-jest $ yarn test
yarn run v1.6.0
$ neutrino test
 FAIL  test\simple_test.js
  ● Test suite failed to run

    TypeError: Cannot read property 'AsymmetricMatcher' of undefined

      at Object.<anonymous> (../neutrino-dev/node_modules/jest-matcher-utils/build/index.js:31:49)
      at Object.<anonymous> (../neutrino-dev/node_modules/expect/build/index.js:3:25)
```

The cause was this:
* `jest-matcher-utils` (used by jest and present in the yarn linked jest preset directory) uses `pretty-format`, which gets hoisted to the node_modules in the monorepo root
* `preact-compat` (which is in the test project's node_modules) indirectly uses a much older version of `pretty-format`, which gets hoisted to the root of the test project's `node_modules`
* because of the current Neutrino `resolve.modules` settings, `jest-matcher-utils` ended up using the older `pretty-format` from the test project `node_modules` instead of the monorepo `node_modules`

Doing a `yarn test --debug` showed the generated modules list being passed to Jest, as:
```js
  "moduleDirectories": [
    "C:\\Users\\Ed\\src\\neutrino-dev\\packages\\jest\\node_modules",
    "C:\\Users\\Ed\\src\\neutrino-dev\\packages\\eslint\\node_modules",
    "node_modules",
    "C:\\Users\\Ed\\src\\test-preact-jest\\node_modules",
    "C:\\Users\\Ed\\src\\neutrino-dev\\packages\\web\\node_modules",
    "C:\\Users\\Ed\\src\\neutrino-dev\\node_modules",
    "C:\\Users\\Ed\\src\\neutrino-dev\\packages\\preact\\node_modules"
  ],
```

Doing any of the following stopped the test from failing:
1. Moving `"C:\\Users\\Ed\\src\\neutrino-dev\\node_modules"` higher up the list (by making the corresponding monorepo-specific call in the web preset a `.prepend()`)
2. Removing `"C:\\Users\\Ed\\src\\test-preact-jest\\node_modules"` (by removing the `.add(neutrino.options.node_modules)` from the web preset)
3. Removing all entries other than `node_modules` (the webpack default). This surprisingly worked for both test & build.

It also seems wrong to not have `node_modules` as the first entry?

I'll probably end up picking (1) or (2) as a workaround for now (since it will prevent any risk to non-monorepo), however I think we really need to come up with a better order/approach here (within the constraints of having to work with monorepo + standalone + standalong w/no hoisting), since this is not the first time I've spent hours debugging an issue to find it's been due to the Neutrino modules resolution order ([another example](https://bugzilla.mozilla.org/show_bug.cgi?id=1426741#c2)).

edmorley:
Since the custom module resolution is not required, and has been the cause of at least 3 bugs that I have spent hours debugging.

The vast majority of modules referenced by Neutrino already use `require.resolve()` and so are unaffected by this change. Those that do not (for example eslint presets, since eslint doesn't support specifying them as full paths), were not helped by this custom module resolution anyway, since the resolution happens outside of webpack, and so relied on hoisting even before this change.

The `node_modules` option has been removed in favour of using the Neutrino API (or else `NODE_PATH`). Customising module resolution is a massive footgun and should not be used in most cases.

Fixes #822.

### DIFF
--- a/packages/jest/src/index.js
+++ b/packages/jest/src/index.js
@@ -71,14 +71,14 @@ module.exports = neutrino => {
       source,
       tests,
       root,
-      node_modules, // eslint-disable-line camelcase
       debug
     } = neutrino.options;
+    const modulesConfig = neutrino.config.resolve.modules.values();
     const aliases = neutrino.config.resolve.alias.entries() || {};
 
     return override({
       rootDir: root,
-      moduleDirectories: neutrino.config.resolve.modules.values(),
+      moduleDirectories: modulesConfig.length ? modulesConfig : ['node_modules'],
       moduleFileExtensions: neutrino.config.resolve.extensions
         .values()
         .map(extension => extension.replace('.', '')),
@@ -93,8 +93,6 @@ module.exports = neutrino => {
             [extensionsToNames(style)]: require.resolve('./style-mock')
           }),
       bail: true,
-      // eslint-disable-next-line camelcase
-      coveragePathIgnorePatterns: [node_modules],
       collectCoverageFrom: [join(
         relative(root, source),
         `**/*.{${extensions.join(',')}}`
--- a/docs/api.md
+++ b/docs/api.md
@@ -163,24 +163,6 @@ Neutrino({
 })
 ```
 
-### `options.node_modules`
-
-Set the directory which contains the Node.js modules of the project. If the option is not set, Neutrino defaults it to
-`node_modules`. If a relative path is specified, it will be resolved relative to `options.root`; absolute paths will be
-used as-is.
-
-```js
-Neutrino({
-  // if not specified, defaults to options.root + node_modules
-
-  // relative, resolves to options.root + modules
-  node_modules: 'modules',
-
-  // absolute
-  node_modules: '/code/website/modules'
-})
-```
-
 ## Other options
 
 ### `options.debug`
--- a/docs/creating-presets.md
+++ b/docs/creating-presets.md
@@ -306,28 +306,6 @@ module.exports = {
 };
 ```
 
-### `options.node_modules`
-
-Set the directory which contains the Node.js modules of the project. If the option is not set, Neutrino defaults it to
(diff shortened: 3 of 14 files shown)


eliperelman:
I think the major reason we used these node_modules settings was to test create-project with `yarn link`, but with verdaccio that's really not necessary now. ++

edmorley:
True :-)

Though I've tested this with a `yarn link`ed Neutrino + presets into Treeherder, and other than the `yarn test` part (which didn't work before either) everything else still worked for me (thanks to our thorough `module.resolve()` usage). So doesn't seem to be any downside :-)

edmorley:
sorry `yarn lint` not `yarn test` (post edited above to correct)

packages/jest/src/index.js
```diff
@@ -93,8 +93,6 @@ module.exports = neutrino => {
             [extensionsToNames(style)]: require.resolve('./style-mock')
           }),
       bail: true,
-      // eslint-disable-next-line camelcase
-      coveragePathIgnorePatterns: [node_modules],
```
eliperelman: We don't need this line any more?

edmorley: The default is the regexp pattern `"/node_modules/"` which matches all node_modules:
https://facebook.github.io/jest/docs/en/configuration.html#coveragepathignorepatterns-array-string

The previous value was the custom `node_modules` setting which no longer exists.

eliperelman: OK perfect.

Issue #96, opened 2017-03-05
Feature request: command to output final webpack config


wmadden:
I recently started a project to try out Neutrino and while the initial setup was pretty easy, as soon as I wanted to make a change to the (eventual) webpack configuration I struggled to work out how to do it.

My first attempt to understand exactly what was ending up in webpack was to try to get ahold of the configuration object that webpack would receive. I was surprised to find there was no way to make `neutrino` output it, either using a `--verbose` flag or with a separate command. In the end I just added a custom preset which would print it out.

One of the biggest problems I observed using Grunt in production is that plugins you'd add would each modify the grunt configuration object, so that the actual configuration of tasks that'd be run would pass through many (often invisible) intermediate states before being used, making debugging impossible. I see a similarity here, in that after applying a couple of presets I had no way to know what webpack configuration was actually being used.

I think it'd be a good idea to make the underlying configuration more transparent to users - give them a way to see what it is before and after applying presets, and with different config properties switched on/off, and encourage them to use it in the documentation.

Moreover it'd give people a way to "eject" from Neutrino (borrowing the terminology from `create-react-app`), should they need to.

psachs21:
I had some similar issues and found it helpful to output the config object. You can do this by adding adding a custom neutrino config like so:

`
\\ neutrino-custom.js
module.exports = (neutrino) => {
  console.log(neutrino.config.toConfig())
};
`

And adding that to the presets (as the last one)

It's not ideal, and a flag on neutrino is cleaner, but it may unblock you.

tj:
dumping the config after each modification would be cool. I haven't used it enough yet to judge if it would be better, but doing `config = middleware(config)` with "vanilla" webpack might be a bit easier to grok than chained webpack, dunno, could be a lot worse, probably more error-prone

eliperelman:
As far as debugging a preset, I usually use `getWebpackOptions` which just maps to `config.toConfig` as @psachs21 mentiond:

```js
module.exports = neutrino => {
  console.log(neutrino.getWebpackOptions());
};
```

I think having a `--debug` flag would be nice, and could probably be used to show an object diff between presets:

```js
const alpha = ({ config }) => config.entry('index').add('src/index.js');
const beta = ({ config }) => config.entry('index').prepend('babel-polyfill');
const gamma = ({ config }) => config.entry('index').add('webpack/hot/dev-server');

// Neutrino v5.alpha .use()
neutrino.use(alpha);
[...]

PR #928, opened 2018-06-05
Improve stringified config output by --inspect

### LINKED ISSUE #328: Plugin class names missing from --inspect output
Currently the output when using `--inspect` includes a `plugins` property, which contains an array of plugin options but not the actual name of the plugin to which the options belong.

For example:
```js
  plugins: [
    {
      keys: [
        'NODE_ENV'
      ],
      defaultValues: {}
    },
    {
      options: {
        template: 'ui/index.html',
        filename: 'index.html',
        hash: false,
        inject: false,
        compile: true,
        favicon: false,
        minify: {
          useShortDoctype: true,
          keepClosingSlash: true,
          collapseWhitespace: true,
          preserveLineBreaks: true
        },
        cache: true,
        showErrors: true,
        chunks: 'all',
        excludeChunks: [],
        title: 'Webpack App',
        xhtml: true,
        appMountId: 'root',
        mobile: true
      }
    },
```

To work out the mapping of plugins to options, I have to cross-reference against the output from `neutrino.config.toConfig()`:
```js
  plugins: 
   [ EnvironmentPlugin { keys: [Array], defaultValues: {} },
     HtmlWebpackPlugin { options: [Object] },
```

This is using neutrino 6.2.0 and webpack-chain 3.3.0 with node 8.5.0.

edmorley:
This take advantage of the new webpack-chain `toString()` added in mozilla-neutrino/webpack-chain#53.

The output from `neutrino --inspect` now lists the correct plugin declarations and arguments, annotates plugins/loaders with hints about how to reference them in a custom Neutrino config, and supports using `__expression` to override the stringified output when needed.

The usage of `deep-sort-object` has been removed since it breaks the new `toString()` comment annotations, and really if sorted output is considered important, it should be handled by webpack-chain itself.

Example output for Treeherder's WIP Neutrino 9 config:
https://emorley.pastebin.mozilla.org/9086982

After this merges, I was thinking we could backport this to Neutrino 8 under a separate flag (eg `--inspect-v9`) which would make it easier for people to diff Neutrino 8 and 9 configs.

Fixes #328.
Refs #96.

### DIFF
--- a/packages/eslint/index.js
+++ b/packages/eslint/index.js
@@ -110,12 +110,16 @@ module.exports = (neutrino, opts = {}) => {
     }
   };
   const options = merge(defaults, opts);
-  const loaderOptions = typeof options.eslint.formatter === 'string'
-    ? deepmerge(options.eslint, {
-      // eslint-disable-next-line global-require
-      formatter: require(`eslint/lib/formatters/${options.eslint.formatter}`)
-    })
-    : options.eslint;
+  const loaderOptions = options.eslint;
+
+  if (typeof loaderOptions.formatter === 'string') {
+    const formatterPath = `eslint/lib/formatters/${loaderOptions.formatter}`;
+    // eslint-disable-next-line global-require
+    loaderOptions.formatter = require(formatterPath);
+    // Improve the stringified output when using --inspect.
+    // eslint-disable-next-line no-underscore-dangle
+    loaderOptions.formatter.__expression = `require('${formatterPath}')`;
+  }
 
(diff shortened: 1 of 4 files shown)


eliperelman:
Note: I don't think we should port back to v8 since that means adding back deep-object-sort, since non-sorted objects is a breaking change.

edmorley:
> under a separate flag (eg --inspect-v9)

eliperelman:
Oops haha.

PR #932, opened 2018-06-06
Stop leaking pluginId into the options passed to plugins

### LINKED ISSUE #929: @neutrinojs/html-template passes invalid pluginId to HtmlWebpackPlugin
The internal `pluginId` option is currently passed to the plugin, when it shouldn't be (doesn't cause an error at the moment, but would if the plugin were to start validating options). 

This also adds more noise to the `--inspect` output.

eg:

```js
    /* neutrino.config.plugin('html-intermittent-failures') */
    new HtmlWebpackPlugin(
      {
        template: 'C:\\Users\\Ed\\src\\neutrino-dev\\node_modules\\html-webpack-template\\index.ejs',
        inject: false,
        appMountId: 'root',
        xhtml: true,
        mobile: true,
        minify: {
          useShortDoctype: true,
          keepClosingSlash: true,
          collapseWhitespace: true,
          preserveLineBreaks: true
        },
        pluginId: 'html-intermittent-failures',
        filename: 'intermittent-failures.html',
        chunks: [
          'intermittent-failures'
        ],
        title: 'Intermittent Failures View'
      }
    ),
```

edmorley:
Previously the Neutrino-specific `pluginId` option was being passed along with the plugin-specific options to the plugin constructor in many of the Neutrino presets.

Fixes #929.

### DIFF
--- a/packages/banner/index.js
+++ b/packages/banner/index.js
@@ -1,8 +1,8 @@
 const { BannerPlugin } = require('webpack');
 
-module.exports = (neutrino, options = {}) => {
+module.exports = (neutrino, { pluginId = 'banner', ...options } = {}) => {
   neutrino.config
-    .plugin(options.pluginId || 'banner')
+    .plugin(pluginId)
     .use(BannerPlugin, [{
       banner: 'require(\'source-map-support\').install();',
       test: neutrino.regexFromExtensions(),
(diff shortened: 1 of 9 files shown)


edmorley:
I wonder whether we actually need the ability to pass `pluginId` on all of these anyway? We don't support it in several other presets, and it's only for specific use-cases that customisation is required (eg html-template). 

Thoughts?

eliperelman:
I'm indifferent on the plugin IDs. There are some where it's definitely helpful, but in some cases I'm wondering if it's hurting us or helping us.

edmorley:
I guess we can revisit later :-)

PR #933, opened 2018-06-06
[v8.x] Correct the URL used for the devServer 'open' feature

### LINKED ISSUE #927: Open browser doesn't work
### Issue

- What version of Neutrino are you using?
8.2.3
- Are you trying to use any presets? If so, which ones, and what versions?
`@neutrinojs/react` (8.2.3)
- Are you using the Yarn client or the npm client? What version?
Yarn 1.7.0.
- What version of Node.js are you using?
8.9.1
- What operating system are you using?
Windows 10
- What did you do?

`.neutrinorc.js`:
```js
module.exports = {
  use: [
    [
      '@neutrinojs/react',
      { devServer: { open: true } }
    ]
  ]
}
```
`$ neutrino start`
- What did you expect to happen?
It should open browser.
- What actually happened, contrary to your expectations?
Error: 'System cannot find http://localhost:5000:5000.'

edmorley:
Previously the port was listed twice in the URL, eg:
`http://localhost:5000:5000`

This PR fixes the issue on `release/v8` and `master` is already fixed due to the #852 refactor.

Fixes #927.

### DIFF
--- a/packages/dev-server/index.js
+++ b/packages/dev-server/index.js
@@ -62,7 +62,7 @@ module.exports = (neutrino, opts = {}) => {
     neutrino.options.https ? { https: neutrino.options.https } : {}
   ]);
   const protocol = options.https ? 'https' : 'http';
-  const url = `${protocol}://${publicHost}:${options.port}`;
+  const url = `${protocol}://${publicHost}`;
 
   neutrino.config
     .devServer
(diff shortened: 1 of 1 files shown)


eliperelman:
Is this true in every `host`, `post`, `public` configuration scenario?

edmorley:
> Is this true in every host, post, public configuration scenario?

Ah it didn't in the `neutrino.options.host` case - PR updated. (The `getPublic()` implementation on `master` doesn't support `neutrino.options.host`, so doesn't need adjusting for this.)

edmorley:
There are a couple of other fixes I'd like to get into `release/v8` before we release a new version. I'm going to try and do those in the next 24 hours.

PR #938, opened 2018-06-07
karma: Update karma-webpack to 4.0.0-beta.0

edmorley:
Since it has improved webpack 4 support and fixes the deprecation warnings. (Sadly it still needs the `optimization` workaround to prevent hangs after compilation.)

An exact version has been used instead of a caret range in case future beta releases make further breaking changes. Once v4 final is released, we can switch back to caret.

See:
https://github.com/webpack-contrib/karma-webpack/blob/next/CHANGELOG.md#400-beta0-2018-03-19


eliperelman:
Oh, we should also set this in create-project.

Issue #865, opened 2018-05-10
Allow passing html-template options for each entry in 'mains'


edmorley:
For multi-entrypoint projects, it's possible to add multiple entries to the Neutrino `mains` option, like so:

```js
    mains: {
      index: 'entry-index.js',
      foo: 'entry-foo.js',
      bar: 'entry-bar.js',
    },
```

However in the web preset, the same `html` options property is used for each, which means there is no way to give the pages different titles or vary the template used.

For mozilla/treeherder's Neutrino v8 WIP, I'm currently working around this with:

```js
    (neutrino) => {
      const entryHtmlOptions = {
        index: {
          // Page uses custom template instead of `html-webpack-template`'s
          template: 'index.html',
          inject: true,
        },
        foo: {
          title: 'Custom Page Title',
        },
        bar: {
          title: 'Another Page Title',
        },
      };
      Object.entries(entryHtmlOptions).forEach(([entry, options]) =>
        neutrino.config
          .plugin(`html-${entry}`)
          .tap(([defaultOptions]) => [{ ...defaultOptions, ...options }]));
    }
```

...but it seems likely that anyone using Neutrino's multi-entrypoint support is having to do something similar - so perhaps should be supported natively.

One way of doing so might be to permit the main's value to be an object (whilst still supporting the string form for the simple case). ie:

```js
    mains: {
      index: {
        source: 'entry-index.js',
        html: {
          template: 'index.html',
          inject: true,
        },
      },
      foo: {
        source: 'entry-foo.js',
        html: {
          title: 'Custom Page Title',
        },
      },
      bar: {
        source: 'entry-bar.js',
        html: {
          title: 'Another Page Title',
        },
      },
    },
```

@mozilla-neutrino/core-contributors - thoughts?

eliperelman:
I like the approach, but would probably hope for a flatter structure rather than nesting in an `html` property. Maybe we should consider dropping html-webpack-template before doing this?

tangdw:
like this 
[vue-cli3-pages](https://cli.vuejs.org/config/#pages)

thany:
Exactly what I need, but if I use the `entryHtmlOptions` approach, how do I access those options in my mains? I was hoping they would go into `process.env` or even straight into `window`, but they are nowhere to be found... So how to use them? @edmorley

timkelty:
[...]

Issue #682, opened 2018-01-23
Allow configuring html-loader via web preset options


edmorley:
The font, image, style and babel loaders support being passed options via the web preset to configure their behaviour:
https://github.com/mozilla-neutrino/neutrino-dev/blob/v8.0.18/packages/web/index.js#L132-L134

However the html-loader doesn't:
https://github.com/mozilla-neutrino/neutrino-dev/blob/v8.0.18/packages/web/index.js#L104

For context, the option I wish to adjust is [html-loader](https://github.com/webpack-contrib/html-loader)'s `attr` option, which defaults to `['img:src']`, but I'd like to change to `['img:src', 'link:href']` so it picks up favicons (`<link href="...">`).

If we did add such an option, we'd probably have to call it `htmlLoader` or similar, since `html` is already taken for the options passed to `html-webpack-plugin`.

My workaround in the meantime:
```js
neutrino.config.module
  .rule('html')
  .use('html')
  .tap(options => ({
    ...options,
    // Override html-loader's default of `img:src`,
    // so it also parses favicon images (`<link href="...">`).
    // TODO: Remove once we update to html-loader 1.0.0+:
    // https://github.com/webpack-contrib/html-loader/issues/17
    attrs: ['img:src', 'link:href'],
  }));
```

edmorley:
Now that the default value for `attrs` has been updated in #943, Treeherder's has no need to be able to pass custom options to `html-loader` when used via the react preset.

As such, I'm happy to close this out, unless we feel it's a feature we really should have regardless?

eliperelman:
We can close, I don't think it's too crazy to ask for an options tap here.

PR #943, opened 2018-06-07
html-loader: Add 'link:href' to attrs so favicons are parsed

edmorley:
By default `html-loader@0.x` has a default `attrs` of `['img:src']`, which means only `` tags are followed when parsing the HTML:
https://github.com/webpack-contrib/html-loader/blob/v0.5.5/README.md#usage
https://github.com/webpack-contrib/html-loader/blob/v0.5.5/index.js#L29

With this change, favicons are included too, and have all the benefits of being included in the webpack build - such as being being inlined by url-loader (if small enough), or else being copied to the output directory automatically and given hashed filenames.

This matches the default that will be used in `html-loader` v1.0.0:
webpack-contrib/html-loader#17

...and saves projects from having to resort to the manual Neutrino API to adjust the options used by `html-loader` (see #865).

### DIFF
--- a/packages/html-loader/index.js
+++ b/packages/html-loader/index.js
@@ -3,4 +3,11 @@ module.exports = (neutrino, options = {}) => neutrino.config.module
   .test(neutrino.regexFromExtensions(['html']))
   .use('html')
     .loader(require.resolve('html-loader'))
-    .options(options);
+    .options({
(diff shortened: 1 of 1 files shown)


edmorley:
Oops there was a typo in the commit message, #865 was meant to say #682. (I've now corrected in the PR description.)

Issue #781, opened 2018-04-02
@neutrinojs/library - compile to standard ES5


flash1293:
IMHO the library preset should configure babel to compile to standard ES5 to be consumable by uglify, incuding async/await and other language features.

eliperelman:
I think this is reasonable, and I am thinking the same thing about react-components as well so they play well with CRA by default. The tradeoff here comes with the size expense.

@mozilla-neutrino/core-contributors thoughts?

edmorley:
Or ideally compile to both - one for the project's `main` and the other for `module`.

eliperelman:
See also #722.

edmorley:
This doesn't block a Neutrino v9 alpha - however since it's a breaking change, someone would need to start working on this soon or else it's not going to make Neutrino 9 final, and would then have to wait until Neutrino 10.

PR #948, opened 2018-06-09
Library package compiles down to ES5

acconrad:
This solves #781 by using [`babel-preset-env#uglify`](https://babeljs.io/docs/plugins/preset-env#targetsuglify) to compile the library package down to ES5

### DIFF
--- a/packages/library/index.js
+++ b/packages/library/index.js
@@ -31,7 +31,7 @@ module.exports = (neutrino, opts = {}) => {
           useBuiltIns: 'entry',
           targets: options.target === 'node' ?
             { node: '8.3' } :
-            { browsers: [] }
+            { uglify: true, browsers: [] }
         }]
       ]
     }, options.babel)


acconrad:
Hi @edmorley ! The PR is fixed per your instructions. `library` and `react-component` both now feature `forceAllTransforms: true` to compile down to ES5. Should make it in time for v9 release :)

eliperelman:
I believe the way CRA solves this is to use `ie9` as the browser to compile to. Any more thoughts here?

packages/react-components/index.js
```diff
@@ -12,10 +13,44 @@ module.exports = (neutrino, opts = {}) => {
       title: 'React Preview'
     },
     manifest: mode === 'development',
+    babel: {},
     externals: opts.externals !== false && {},
     style: { extract: { plugin: { filename: '[name].css' } } }
   }, opts);
 
+  Object.assign(options, {
+    babel: compileLoader.merge({
```
edmorley: The react-components preset inherits from the react preset, which already configures `@babel/preset-env`. As such I don't think we need to add this here, and instead could just override the targets passed to the react preset.

acconrad: @edmorley so the entire `Object.assign` portion can be safely ignored?

edmorley: Yeah I believe so. I'd just set default values for `options.targets` and let those be passed down to the react (and thus web) presets.

acconrad: Not sure this is entirely true.

So I see that basically every package starts with an empty babel object and builds on it ->

eliperelman:
@acconrad could you rebase this please?

acconrad:
all set @eliperelman !

packages/react-components/index.js
edmorley: The reason the other presets start with an empty babel object, is that they need to customise the list of plugins used.

However for this PR, the only thing that needs changing, is the list of browser/node targets, which can be passed down via the standard options, and will cause the existing web preset functionality to use them:
https://github.com/mozilla-neutrino/neutrino-dev/blob/e597c615e2c76e9986b6262249e35001a00efc0f/packages/web/index.js#L94

packages/library/index.js
```diff
@@ -27,6 +27,7 @@ module.exports = (neutrino, opts = {}) => {
       presets: [
         [require.resolve('@babel/preset-env'), {
           debug: neutrino.options.debug,
+          forceAllTransforms: true,
```
edmorley: I don't think we want `forceAllTransforms` for the same reasons as mentioned above

acconrad: so what are those targets then?

eliperelman: CRA uses `ie9` to compile for ES5.

acconrad: do there need to be targets for other browsers? i thought you needed to specify them for each type

eliperelman: For the browser targets, Babel essentially reduces support down to the least common denominator. With `ie9` being pretty low, it outputs code to the least common denominator of ES5.

eliperelman: All that to say, by specifying `ie9`, all the output code is ES5 which all of the other browsers will happily use.

acconrad:
@edmorley I think that takes care of adding in the browser support to `ie 9`, is there anything else that needs modification? I think I added in the options piece for `react-components`

edmorley:
Could you also update the README for the two packages?
https://github.com/mozilla-neutrino/neutrino-dev/blame/master/packages/react-components/README.md#L18
https://github.com/mozilla-neutrino/neutrino-dev/blame/master/packages/library/README.md#L11
https://github.com/mozilla-neutrino/neutrino-dev/blame/master/packages/library/README.md#L261

For the change to the react-components README, I think the browser versions reference needs to be moved out of the "Modern Babel compilation ..." line (since it's under the "partially extends" section), and to a new bullet specific to that package.

packages/react-components/README.md
```diff
@@ -14,8 +14,9 @@ other Neutrino middleware, so you can build, test, and publish multiple React co
 - Support for React Hot Loader
 - Write JSX in .js or .jsx files
 - Support for importing web workers with `.worker.*` file extensions
+- Compiles down to ES5 for all major browsers with support up to IE 9
```
eliperelman: Can you change this to say "Compiles to ES5 to support major browsers, publishing to npm, and library interoperability"

acconrad: all set!

edmorley:
@acconrad, there are conflicts with `master` (I think due to the use of merge rather than rebase, otherwise GitHub would normally be able to handle them) - could you rebase?

PR #951, opened 2018-06-13
eslint: Lint the tests directory by default too

### LINKED ISSUE #331: Add the tests directory to default lint include
Currently neutrino-middleware-eslint defaults the directory `include` list to just `[neutrino.options.source]`:
https://github.com/mozilla-neutrino/neutrino-dev/blob/10819ad89d9d408f79e29487af5ea430701ac6df/packages/neutrino-middleware-eslint/index.js#L30

It seems like linting should also be run against tests - and so `include` default instead to `[neutrino.options.source, neutrino.options.test]`. (At least for `neutrino lint` or `neutrino test`, perhaps not for start/build?)

Also it looks like the include/exclude from neutrino-preset-airbnb-base can be removed, since it's just repeating what is in neutrino-middleware-eslint:
https://github.com/mozilla-neutrino/neutrino-dev/blob/10819ad89d9d408f79e29487af5ea430701ac6df/packages/neutrino-preset-airbnb-base/index.js#L24

Thoughts? :-)

edmorley:
The `tests` directory was already added to the `lint` script command by create-project, however it was not yet listed in the `include` for `eslint-loader`.

The testing presets already update `envs` accordingly, to prevent errors from mocha/jest/... globals.

The redundant `eslint.extensions` option has been removed, since that was only used by the ESLint Node API, which we no longer use.

Fixes #331.

### DIFF
--- a/packages/eslint/README.md
+++ b/packages/eslint/README.md
@@ -38,14 +38,14 @@ const eslint = require('@neutrinojs/eslint');
 
 // Usage shows default values
 neutrino.use(eslint, {
-  test: /\.(js|jsx)$/,
-  include: [], /* Should specify either include or exclude */
-  exclude: [], /* Should specify either include or exclude */
+  test: neutrino.regexFromExtensions(), // Uses extensions from neutrino.options.extensions
+  include: [neutrino.options.source, neutrino.options.tests],
+  exclude: [],
   eslint: {
+    failOnError: neutrino.config.get('mode') === 'production',
     cwd: neutrino.options.root,
     useEslintrc: false,
     root: true,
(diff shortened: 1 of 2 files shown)


edmorley:
I've updated the PR to remove the dotfiles addition, since it causes `import/no-extraneous-dependencies` lint errors like so:
https://travis-ci.org/mozilla-neutrino/neutrino-dev/jobs/391908556#L6384

...which would mean us having to override the airbnb rule here:
https://github.com/airbnb/javascript/blob/04ab225858b75b9fb1180075b92845700849ea5b/packages/eslint-config-airbnb-base/rules/imports.js#L71-L93

PR #958, opened 2018-06-21
Use lerna exact mode & bump version to 9.0.0-0 prerelease

edmorley:
* Switches to lerna exact mode (making it possible to install historic versions of monorepo packages without them using a mixture of old and new minor/patch versions).
* Bumps the monorepo package versions to 9.0.0-0 prerelease (to make it clearer that `master` is for the upcoming Neutrino 9; this won't be the actual version that gets published).
* Cleans up the lerna configuration.

See individual commit messages for more information.


eliperelman:
Question: what affect does removing the force publish have here?

edmorley:
Force publish wasn't removed, it was moved to `lerna.json`

eliperelman:
Oh oh oh, I see! Carry on. 🤣

PR #967, opened 2018-06-25
[v8.x] create-project: Install a matching version of Neutrino packages

edmorley:
The Neutrino 8 equivalent of #957, minus the refactoring and without using `Object.values()` (since the `release/v8` branch must remain compatible with Node 6).


edmorley:
After merging this I think we should be ready for a new Neutrino 8.x release (possibly `8.3.0` given the new `--inspect-new`). I'm happy to try that (I'll test using Verdaccio and diff the archive contents/permissions to avoid the previous issue, if so), or if you want to do it that's fine too.

PR #972, opened 2018-06-29
Improve mode/NODE_ENV defaults and interaction

### LINKED ISSUE #900: Properly set `mode` and `NODE_ENV` for test middleware

### LINKED ISSUE #971: @neutrinojs/stylelint doesn't set failOnError
The docs say `failOnError` is set:
https://github.com/mozilla-neutrino/neutrino-dev/blame/e597c615e2c76e9986b6262249e35001a00efc0f/packages/stylelint/README.md#L44

...however the preset doesn't actually do so:
https://github.com/mozilla-neutrino/neutrino-dev/blob/e597c615e2c76e9986b6262249e35001a00efc0f/packages/stylelint/index.js#L5-L11

edmorley:
Previously the only way to override `mode` was by passing `--mode` on the command line. However this is not possible with all tools, since some (such as karma) reject unrecognised arguments.

Now:
* `mode` is derived from `NODE_ENV` if `--mode` wasn't passed, and !production `NODE_ENV` falls back to mode `development`.
* if `--mode` is passed, it takes priority over `NODE_ENV`.
* if neither `mode` nor `NODE_ENV is defined, then `NODE_ENV` is set to `production`, however `mode` is left undefined, which causes webpack to output a helpful warning about relying on defaults.
* the template test runner configs set a default `NODE_ENV` of `test`.
* `@neutrinojs/stylelint` now also correctly sets `failOnError`.

Fixes #900.
Fixes #971.
Closes #955.

### DIFF
--- a/packages/eslint/index.js
+++ b/packages/eslint/index.js
@@ -84,7 +84,8 @@ module.exports = (neutrino, opts = {}) => {
     include: !opts.include ? [neutrino.options.source, neutrino.options.tests] : undefined,
     eslint: {
       cache: true,
-      failOnError: neutrino.config.get('mode') === 'production',
+      // Make errors fatal not just for 'production' but also 'test'.
+      failOnError: process.env.NODE_ENV !== 'development',
       cwd: neutrino.options.root,
       useEslintrc: false,
       root: true,
--- a/packages/neutrino/index.js
+++ b/packages/neutrino/index.js
@@ -8,13 +8,25 @@ const configPrefix = 'neutrino.config';
 module.exports = (middleware = { use: ['.neutrinorc.js'] }, options = {}) => {
   const neutrino = new Neutrino(options);
   const { argv } = yargs;
-  const mode = argv.mode || 'production';
+  let { mode } = argv;
 
-  if (!process.env.NODE_ENV) {
+  if (mode) {
+    // If specified, --mode takes priority and overrides any existing NODE_ENV.
     process.env.NODE_ENV = mode;
+  } else if (process.env.NODE_ENV) {
+    // Development mode is most appropriate for a !production NODE_ENV (such as `NODE_ENV=test`).
+    mode = (process.env.NODE_ENV === 'production') ? 'production' : 'development';
+  } else {
+    // Default NODE_ENV to the more strict value, to save needing to do so in .eslintrc.js.
+    // However don't set `mode` since webpack already defaults it to `production`, and in so
+    // doing outputs a useful message informing users that they are relying on the defaults.
+    process.env.NODE_ENV = 'production';
   }
 
-  neutrino.config.mode(mode);
+  if (mode) {
+    neutrino.config.mode(mode);
+  }
+
   neutrino.register('webpack', webpack);
 
   if (middleware) {
--- a/packages/stylelint/index.js
+++ b/packages/stylelint/index.js
@@ -6,6 +6,8 @@ module.exports = (neutrino, { pluginId = 'stylelint', ...opts } = {}) => {
     configBasedir: neutrino.options.root,
     files: '**/*.+(css|scss|sass|less)',
     context: neutrino.options.source,
+    // Fail for all of 'production', 'lint' and 'test'.
+    failOnError: process.env.NODE_ENV !== 'development',
     formatter: formatters.string,
     ...opts
   };
--- a/packages/create-project/commands/init/templates/jest/jest.config.js
+++ b/packages/create-project/commands/init/templates/jest/jest.config.js
@@ -1,3 +1,5 @@
 const neutrino = require('neutrino');
 
+process.env.NODE_ENV = process.env.NODE_ENV || 'test';
+
 module.exports = neutrino().jest();
--- a/packages/create-project/commands/init/templates/karma/karma.conf.js
+++ b/packages/create-project/commands/init/templates/karma/karma.conf.js
@@ -1,3 +1,5 @@
 const neutrino = require('neutrino');
 
+process.env.NODE_ENV = process.env.NODE_ENV || 'test';
+
 module.exports = neutrino().karma();
--- a/packages/create-project/commands/init/templates/mocha/mocha.config.js
+++ b/packages/create-project/commands/init/templates/mocha/mocha.config.js
@@ -1,1 +1,3 @@
+process.env.NODE_ENV = process.env.NODE_ENV || 'test';
+
 require('neutrino')().mocha();
--- a/packages/eslint/README.md
+++ b/packages/eslint/README.md
@@ -43,7 +43,7 @@ neutrino.use(eslint, {
   exclude: [],
   eslint: {
     cache: true,
-    failOnError: neutrino.config.get('mode') === 'production',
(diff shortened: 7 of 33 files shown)


edmorley:
Once we agree on the behaviour/UX I'll add some tests :-)

.eslintrc.js
```diff
@@ -1,3 +1,5 @@
 const neutrino = require('./packages/neutrino');
 
+process.env.NODE_ENV = process.env.NODE_ENV || 'lint';
```
edmorley: Possible values instead of "lint" might be "test" or "production". Though given we use "test" for `karma.config.js` etc, it seems wrong to just call this "production". 

ESLint config conditionals that choose to rely on `NODE_ENV` (eg enabling/disabling the `no-console` rule) should be using `NODE_ENV !== 'development'` rather than `NODE_ENV === 'production'` for compatibility with `NODE_ENV='test'` anyway -- so using `NODE_ENV='lint'` doesn't seem like it would cause any more problems.

packages/neutrino/index.js
```diff
@@ -4,14 +4,21 @@ const webpack = require('./webpack');
...
-  if (!process.env.NODE_ENV) {
+  let mode = args.mode;
+
+  if (mode) {
     process.env.NODE_ENV = mode;
+  } else if (process.env.NODE_ENV) {
+    // Development mode is most appropriate for a !production NODE_ENV (such as `NODE_ENV=test`).
+    mode = (process.env.NODE_ENV === 'production') ? 'production' : 'development';
+  } else {
+    throw new Error('Either --mode must be specified via the CLI, or else NODE_ENV be defined.');
```
edmorley: Alternatives to throwing are:
* default to `NODE_ENV='production'` and `mode='production'`
* default to `NODE_ENV='production'` and don't set `mode` at all. This would mean for workflows that include a webpack compile (ie not `yarn lint`), webpack would output the "no mode was specified, using production" warning message, which might help solve the "prevent people from accidentally omitting mode/NODE_ENV` concern.

Both of the alternatives would mean we could drop the `process.env.NODE_ENV = ...` boilerplate from the template `.eslintrc.js` files -- though we would still need it for `karma.config.js`.

packages/stylelint/index.js
```diff
@@ -6,6 +6,8 @@ module.exports = (neutrino, { pluginId = 'stylelint', ...opts } = {}) => {
     configBasedir: neutrino.options.root,
     files: '**/*.+(css|scss|sass|less)',
     context: neutrino.options.source,
+    // Fail for all of 'production', 'lint' and 'test'.
+    failOnError: process.env.NODE_ENV !== 'development',
```
edmorley: This fixes #971.

packages/eslint/index.js
```diff
@@ -84,7 +84,8 @@ module.exports = (neutrino, opts = {}) => {
     include: !opts.include ? [neutrino.options.source, neutrino.options.tests] : undefined,
     eslint: {
       cache: true,
-      failOnError: neutrino.config.get('mode') === 'production',
+      // Fail for all of 'production', 'lint' and 'test'.
+      failOnError: process.env.NODE_ENV !== 'development',
```
edmorley: I've flipped this, so it is set to true for `yarn test` too.

packages/neutrino/index.js
```diff
@@ -8,10 +8,15 @@ const configPrefix = 'neutrino.config';
 module.exports = (middleware = { use: ['.neutrinorc.js'] }, options = {}) => {
   const neutrino = new Neutrino(options);
   const { argv } = yargs;
-  const mode = argv.mode || 'production';
+  let { mode } = argv;
 
-  if (!process.env.NODE_ENV) {
+  if (mode) {
     process.env.NODE_ENV = mode;
```
eliperelman: Here we force `NODE_ENV` to always be the same value as `--mode`, but that's not what we want right? What about `NODE_ENV=development webpack --mode production`?

edmorley: This was intentional (there's a note in the PR description), for the reasons under scenario G here:
https://github.com/mozilla-neutrino/neutrino-dev/pull/955#issuecomment-399957959

...though open to counter-arguments :-)

edmorley:
@eliperelman, any more thoughts on the UX choices? I'm keen to get them finalised, so I can updated the tests in this PR, then press on with the overall docs update in another PR. 

(Context: I really want to get Treeherder builds faster soon (ie next week or two), which means a Neutrino 9 alpha by then, or else stopping using Neutrino and just switching to plain webpack)

eliperelman:
I still a little uneasy about https://github.com/mozilla-neutrino/neutrino-dev/pull/972#discussion_r199492880 but I can't articulate my issue with it. I'd say let's roll with it, test it out, and revise if necessary.

.eslintrc.js
edmorley: @eliperelman thoughts about this part?

packages/neutrino/index.js
edmorley: @eliperelman and this?

.eslintrc.js
eliperelman: I'm OK with this, although my initial gut reaction wishes this existed in `eslint()` instead of being dotfile boilerplate.

edmorley: Yeah me too -- though I guess that would mean we would have to change the whole way the middlewares are instantiated.

eliperelman: Yeah which is too much right now. The thing that prevents us from doing this is not being able to reliably infer when the `NODE_ENV` was set by Neutrino vs. the environment.

edmorley: That definitely gets us closer, however if I follow correctly, it would mean that the web preset (and any others) see the NODE_ENV as `'production'` and not `'test'` -- and that even for `.karma()`, parts of the config still comes from other presets.

Therefore it seems that the only way to solve this is make the meat of the presets be a step that occurs **after** they register themselves.

ie:
```js
// registers the presets and what output commands they include, but doesn't actually generate any config or adjust NODE_ENV
neutrino.use(some_lint_preset);
neutrino.use(some_build_preset);
neutrino.use(some_test_preset);

// The output function then sets NODE_ENV if undefined (in this case to `"test"`)
// and then actually generates the config, where all presets get to see `NODE_ENV=test`
neutrino.karma();
```

edmorley: (but that's a major change and would mean backwards incompatibility with all existing presets)

eliperelman: I think keeping middleware compat is more important here.

edmorley:
👍

PR #983, opened 2018-07-05
Move @neutrinojs/env into the web preset & support default values

edmorley:
Since `EnvironmentPlugin` doesn't require any non-webpack dependencies or complex configuration, it's simpler to inline the plugin directly in presets that use it (which is currently just the web preset).

The argument passed to the plugin is now correctly wrapped in an array (as expected by `webpack-chain`), meaning the object form that allows setting defaults is now also supported:
https://webpack.js.org/plugins/environment-plugin/#usage-with-default-values

The options change is non-breaking, however the removal of `@neutrinojs/env` is, for people who used it directly.

Closes #657.


edmorley:
Neutrino 8, even

PR #990, opened 2018-07-10
Ensure output override is a function

### LINKED ISSUE #908: Make Neutrino adapter more strict about arguments passed
I was creating my own `karma.conf.js` to work around #907, and had something like:

```js

module.exports = (config) => {
  config.set({
    // ...
    webpack: neutrino().webpack({
      mode: 'development',
    }),
  });
};
```

This was still producing the `The 'mode' option has not been set, webpack will fallback to 'production'` warning - and I eventually figured out that `.webpack()` takes a function not an object, so should be used like so:

```js
    webpack: neutrino().webpack(config => ({
      ...config,
      mode: 'development',
    })),
```

...although that didn't seem to work either.

Regardless, Neutrino should throw if invalid arguments are passed, rather than silently ignoring the config.

eliperelman:
Fixes #908.

### DIFF
--- a/packages/neutrino/index.js
+++ b/packages/neutrino/index.js
@@ -28,6 +28,10 @@ module.exports = (middleware = { use: ['.neutrinorc.js'] }, options = {}) => {
         process.exit();
       }
 
+      if (typeof override !== 'function') {
+        throw new Error(`The output override for "${name}" must be a function`);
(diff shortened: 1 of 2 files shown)


edmorley (approved):
Many thanks!

PR #992, opened 2018-07-11
Add `transform-object-rest-spread` plugin for Node

Ameobea:
Since the Node preset already includes support for most modern Ecmascript features including async functions, `import`, etc., the lack of support for the spread operator is confusing and seems out of place.  This commit adds support for the spread operator by default, allowing for it to be parsed properly when users use it in their source code.

### DIFF
--- a/packages/node/index.js
+++ b/packages/node/index.js
@@ -39,7 +39,8 @@ module.exports = (neutrino, opts = {}) => {
     include: [neutrino.options.source, neutrino.options.tests],
     babel: compile.merge({
       plugins: [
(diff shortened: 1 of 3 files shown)


eliperelman:
In the v8 branch, you would be correct that the Node.js settings by default don't include anything additional to support object rest/spread syntax, however in master the support matrix is for Node.js v8+ and the `current` transpilation target, which Node.js natively supports object rest/spread.

This means the upcoming v9 release can use object rest/spread natively. Regardless, thank you so much for taking the time to submit this contribution, we still appreciate it!

PR #994, opened 2018-07-13
Target IE9 as the browser to support ES5 compilation

acconrad:
A continuation of #948 without all of the rebasing issues


acconrad:
@edmorley @eliperelman should be good to go now

PR #999, opened 2018-07-17
Bug fix: Respect externals options in library plugin

### LINKED ISSUE #987: @neutrinojs/library: the externals options is ignored
The `externals` options is documented here:

https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/library/README.md#preset-options

but it appears to me that it is completely ignored by the implementation.
See in particular this line (current master branch):

https://github.com/mozilla-neutrino/neutrino-dev/blob/46b88910a621a4e23b36ae55412b4094504770cc/packages/library/index.js#L74

where `nodeExternals` is invoked without forwarding any options.

sloria:
Previously, externals was ignored. This passes along
the options to webpack-node-externals.

Fixes #987 

I verified that this fix in a project I was working on. Without this change, `externals.whitelist` had no effect. With this change, whitelisted modules get bundled into the output file.

### DIFF
--- a/packages/library/index.js
+++ b/packages/library/index.js
@@ -14,6 +14,7 @@ module.exports = (neutrino, opts = {}) => {
     target: 'web',
     libraryTarget: 'umd',
     babel: {},
+    externals: opts.externals !== false && {},
     clean: opts.clean !== false && {
       paths: [neutrino.options.output]
     }
@@ -56,7 +57,7 @@ module.exports = (neutrino, opts = {}) => {
   neutrino.config
     .when(hasSourceMap, () => neutrino.use(banner))
     .devtool('source-map')
-    .externals([nodeExternals()])
+    .externals([nodeExternals(options.externals)])
     .target(options.target)
     .context(neutrino.options.root)
     .output


packages/library/index.js
```diff
@@ -56,7 +56,7 @@ module.exports = (neutrino, opts = {}) => {
   neutrino.config
     .when(hasSourceMap, () => neutrino.use(banner))
     .devtool('source-map')
-    .externals([nodeExternals()])
+    .externals([nodeExternals(options.externals)])
```
edmorley: `options.externals` isn't defined by default, so this will be calling `nodeExternals(undefined)`. This is currently handled ok, due to:
https://github.com/liady/webpack-node-externals/blob/cfa1c5b33752fb8cb72360167ae89b23816cdefe/index.js#L22

...however it doesn't feel right to be relying on an implementation detail that they could possibly change in a minor version.

As such, perhaps the implementation should copy what `react-components` does? (With optional cleanup there, since the current `opts.externals !== false && {}` overly complex, unless I'm mis-remembering how deepmerge handles merging different data types.)

sloria:
OK, I've gone ahead and updated the code to use the same logic that react-components is using. Let me know if there's anything else that needs to be done before this is merged.

edmorley:
Hi! Travis is failing due to an unrelated Jest/JSDom regression (facebook/jest#6766). We're holding off merging any more things to `master` until it's resolved.

sloria:
@edmorley OK. I ran into that Jest issue in some of my projects. The fix is to set the environment, like so:

```
  "jest": {
    "testEnvironment": "node"
  },
```

edmorley:
Whilst that would work around the issue, it seems like this should be fixed in Jest, so once they do so, we'd want to remove the workaround again anyway.

sloria:
I understand. But shouldn't neutrino override `"testEnvironment"` anyway? Jest's default is to run against jsdom.

https://github.com/facebook/jest/blob/fb2a6ac4d3f4e9530941e24030b52f3d506aa2de/packages/jest-config/src/defaults.js#L66

edmorley:
> I understand. But shouldn't neutrino override "testEnvironment" anyway? Jest's default is to run against jsdom.

Ah good point! So the default of `'jsdom'` is fine for the browser-targetting Neutrino presets, but perhaps not what is wanted for `@neutrinojs/node` etc. I've filed #1020 for this.

In the meantime, a new Jest minor version has been released, which fixes the test failures - I've retriggered and Travis is now green :-)

packages/library/index.js
edmorley: Thank you for updating the PR - it's missing the `.when()` of the react-components implementation though?
https://github.com/mozilla-neutrino/neutrino-dev/blob/ed21fa6074354609389710f051286356b073cb4e/packages/react-components/index.js#L53

sloria:
@eliperelman Sorry for the delay in my response. I'm not able to work on this more right now, as I'm no longer using neutrino.

I'd be happy if someone else brought this one home.

eliperelman:
Continuing in #1040, thank you so much for your help here!

PR #1011, opened 2018-08-01
[v8.x] Fix Babel minification documentation

sammarks:
When using the `react-components` preset, I wanted to not minify class names and spent quite a bit of time following code trails to find out that the documentation actually appears to be incorrect with regard to Babel minification options.

Looks like all we need is just an extra `minify` level, and with that change the configuration started working for me in local tests on version 8.

I also targeted this toward version 8 rather than master because it appears the minification configuration is changed a lot in master.

### DIFF
--- a/docs/packages/web/README.md
+++ b/docs/packages/web/README.md
@@ -287,7 +287,7 @@ module.exports = {
             removeConsole: true,
             removeDebugger: true,
           },
-          preset: {}
+          plugin: {}
         }
       },
 
--- a/docs/packages/web/README.md
+++ b/docs/packages/web/README.md
@@ -283,8 +283,11 @@ module.exports = {
       // Example: Remove console and debugger from output
       minify: {
         babel: {
-          removeConsole: true,
-          removeDebugger: true,
+          minify: {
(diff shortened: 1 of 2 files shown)


docs/packages/web/README.md
```diff
@@ -283,8 +283,11 @@ module.exports = {
       // Example: Remove console and debugger from output
       minify: {
         babel: {
-          removeConsole: true,
-          removeDebugger: true,
+          minify: {
+            removeConsole: true,
+            removeDebugger: true,
+          },
+          preset: {}
```
edmorley: I think this should be `plugin` rather than `preset`. See:
https://github.com/sammarks/neutrino-dev/tree/03689f6a60e896e173bb8e1f47141a483ecc8a2e/docs/packages/babel-minify#usage
https://github.com/mozilla-neutrino/neutrino-dev/blob/release/v8/packages/babel-minify/index.js#L5

edmorley: @sammarks do you know if you'll have a chance to make the change above, or would you prefer me to update it? :-)

PR #1029, opened 2018-08-21
Support per-page HTML template customisation via 'mains'

### LINKED ISSUE #865: Allow passing html-template options for each entry in 'mains'

edmorley:
This adds support for defining each entry point in `mains` as an object, where the path to the entry point is now defined under an `entry` property, and any other properties can be used by presets for page-specific options. (The short form using a string is still supported.)

In the case of `@neutrinojs/web` (and presets that inherit from it), these additional properties are then used to override the options passed to `html-webpack-plugin`, allowing for page-specific customisation of the generated HTML template.

For example:

```js
module.exports = {
  options: {
    mains: {
      index: {
        entry: './index',
        // Options here take priority over the preset's `html` options below.
        title: 'Site Homepage',
      },
      admin: {
        entry: './admin',
        title: 'Admin Dashboard',
      },
      account: {
        entry: './user',
        inject: true,
        template: './my-custom-template.html',
      },
    }
  },
  use: ['@neutrinojs/web', {
    // Customise the defaults used for all pages.
    html: {
      minify: false,
    }
  }]
}
```

For a list of the available `html-webpack-plugin` options, see:
https://github.com/jantimon/html-webpack-plugin#options

Fixes #865.


docs/api.md
```diff
@@ -138,27 +138,34 @@ Neutrino({
 ### `options.mains`
 
 Set the main entry points for the application. If the option is not set, Neutrino defaults it to:
```
edmorley: There is a lot of pre-existing duplication between `api.md`, `creating-presets.md` and `customization.md` that is out of scope for this PR, but something we should figure out how to avoid at some point.

packages/neutrino/Neutrino.js
```diff
@@ -138,38 +142,46 @@ module.exports = class Neutrino {
...
+            };
           },
           set(newValue) {
-            value = newValue;
+            normalizedConfig = normalizeMainConfig(newValue);
           }
         });
       });
 
     this.mainsProxy = new Proxy(options.mains, {
```
edmorley: `mainsProxy` appears to be unused. Do we still need it?

eliperelman: We need the Proxy itself, but only exposed it on the Neutrino API just in case. We can certainly remove it if we don't foresee exposing it as being useful.

docs/creating-presets.md
```diff
@@ -272,35 +272,42 @@ Set the main entry points for the application. If the option is not set, Neutrin
...
+
 Notice the entry point has no extension; the extension is resolved by webpack. If relative paths are specified,
 they will be computed and resolved relative to `options.source`; absolute paths will be used as-is.
- 
+
 By default these main files are not required to be in JavaScript format. They may also potentially be JSX, TypeScript,
 or any other preprocessor language. These extensions should be specified in middleware at
 `neutrino.config.resolve.extensions`.
 
+
```
eliperelman: Are there 2 blank lines here?

edmorley: Ah yes I'll remove this additional newline

eliperelman:
I think this is still technically a breaking change since the value exposed by the API is now never a string.

PR #1031, opened 2018-08-21
Migrate to new org, repo, and license

eliperelman:


packages/jest/src/index.js
```diff
@@ -97,6 +97,7 @@ module.exports = neutrino => {
         relative(root, source),
         `**/*.{${extensions.join(',')}}`
       )],
+      testEnvironment: neutrino.config.get('target') === 'node' ? 'node' : 'jsdom',
```
edmorley: This change shouldn't be here - could you rebase on master before merging if the PR branch has gotten muddled?

PR #1041, opened 2018-08-23
Specify root option for monorepo scripts and project creation

### LINKED ISSUE #973: Template .eslintrc.js files no longer set 'root'
#825 was regressed when #852 landed, since `root` is no longer set in the template files, eg:
https://github.com/mozilla-neutrino/neutrino-dev/blob/e597c615e2c76e9986b6262249e35001a00efc0f/packages/create-project/commands/init/templates/airbnb/.eslintrc.js#L3

To test that this has regressed, change to a directory a couple of levels deeper, and run eg:
```
node ../../node_modules/eslint/bin/eslint.js --cache --ext js,jsx --report-unused-disable -directives ".*.js" "*.js" . -c ../../.eslintrc.js
Cannot read config file: C:\Users\Ed\src\treeherder\.eslintrc.js
Error: Cannot find module '.neutrinorc.js'
Error: Cannot read config file: C:\Users\Ed\src\treeherder\.eslintrc.js
Error: Cannot find module '.neutrinorc.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:594:15)
    at Function.Module._load (internal/modules/cjs/loader.js:520:25)
    at Module.require (internal/modules/cjs/loader.js:650:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at requireFromRoot (C:\Users\Ed\src\neutrino-dev\packages\neutrino\Neutrino.js:33:10)
    at Neutrino.use (C:\Users\Ed\src\neutrino-dev\packages\neutrino\Neutrino.js:201:16)
    at middleware.use.map.usage (C:\Users\Ed\src\neutrino-dev\packages\neutrino\Neutrino.js:216:44)
    at Array.map (<anonymous>)
    at Neutrino.use (C:\Users\Ed\src\neutrino-dev\packages\neutrino\Neutrino.js:216:26)
    at module.exports (C:\Users\Ed\src\neutrino-dev\packages\neutrino\index.js:21:14)
```

eliperelman:
Fixes #973.

I'm not sure how this wasn't happening before, but when running without the `--no-eslintrc` bits in the precommit I would get errors around eslint being unable to load the config file. I think this was related to eslint trying to load these nested eslintrc files to lint their subdirectories, which we definitely don't want during development.

So now I added `--no-eslintrc -c ./.eslintrc.js` so it will lint the monorepo using **only** our top-level eslintrc. When you run it like this though, eslint gives you warnings that the create-project eslintrc files are ignored by default, so the eslintignore line addition is to add these back to being linted, but not used as config.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,8 +1,3 @@
 const neutrino = require('./packages/neutrino');
 
-module.exports = neutrino({
-  use: require('./.neutrinorc'), // eslint-disable-line global-require
-  options: {
-    root: __dirname
-  }
-}).eslintrc();
+module.exports = neutrino().eslintrc();
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,8 @@
 const neutrino = require('./packages/neutrino');
 
-module.exports = neutrino(require('./.neutrinorc')).eslintrc();
+module.exports = neutrino({
+  use: require('./.neutrinorc'), // eslint-disable-line global-require
+  options: {
+    root: __dirname
+  }
+}).eslintrc();
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,3 +3,8 @@ packages/**/node_modules/
 
 # This file intentionally contains invalid syntax so results in a parse error.
 packages/neutrino/test/fixtures/test-module/errorMiddleware.js
+
+# By using --no-eslintrc in yarn lint, eslint no longer lints the nested
+# RC files. We then negate this ignore so they are still linted, but no longer
+# used as configuration by our development lint command.
+!packages/create-project/bin/commands/init/templates/*/.eslintrc.js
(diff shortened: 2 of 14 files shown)


eliperelman:
I know this is similar to the other PR, but I did have an extra change I wanted to check.

.eslintrc.js
```diff
@@ -1,3 +1,8 @@
 const neutrino = require('./packages/neutrino');
 
-module.exports = neutrino(require('./.neutrinorc')).eslintrc();
+module.exports = neutrino({
+  use: require('./.neutrinorc'), // eslint-disable-line global-require
```
edmorley: Perhaps we should move the `require()` to the top of the file and save needing the `eslint-disable-line`? (in this file and the others)

eliperelman: Sure thing.

PR #1042, opened 2018-08-23
Web: Raise splitChunks.maxInitialRequests to 5 in production

edmorley:
Since it reduces build times, and the greater number of requests is mitigated by use of long term caching (and a non-issue if using HTTP2). The webpack 4 default value is `3` and in webpack 5 it will be `4`.

See:
https://webpack.js.org/plugins/split-chunks-plugin/#splitchunks-maxinitialrequests
https://github.com/webpack/webpack/blob/v4.17.1/lib/WebpackOptionsDefaulter.js#L238-L240
https://github.com/webpack/webpack/projects/5#card-10750335

### DIFF
--- a/packages/web/index.js
+++ b/packages/web/index.js
@@ -14,6 +14,7 @@ const HtmlWebpackIncludeSiblingChunksPlugin = require('html-webpack-include-sibl
 const ManifestPlugin = require('webpack-manifest-plugin');
 
 module.exports = (neutrino, opts = {}) => {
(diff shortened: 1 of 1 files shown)


edmorley:
The webpack value is being conservative, since by default they don't use hashed filenames or other long-term caching techniques - and when using HTTP1, browsers typically have a maximum of 6 parallel requests per domain. (But with long term caching, and the `immutable` header, it's more likely only a subset of requests will be required, plus with HTTP2 there is no maximum per domain.)

PR #1043, opened 2018-08-25
Docs: Add a Content-Security-Policy and other best practices

edmorley:
This implements the recommendations from:
https://infosec.mozilla.org/guidelines/web_security

Which raises the grade given by https://observatory.mozilla.org/ from a `D` to an `A+`:
https://observatory.mozilla.org/analyze/neutrinojs.org
https://observatory.mozilla.org/analyze/master.neutrinojs.org

The headers are configured according to:
https://www.netlify.com/docs/headers-and-basic-auth/#structured-configuration
https://github.com/toml-lang/toml#string

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,5 @@
 mkdocs==1.0.2
+# When updating the theme, check the Netlify PR preview to see if any changes
+# are required to the Content-Security-Policy header in netlify.toml.
 mkdocs-material==3.0.3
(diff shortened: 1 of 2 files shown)


edmorley:
This needs a few tweaks before it's ready. (And awaiting replies to some tickets against mkdocs-material)

edmorley:
I don't think this is worth the effort, given the chance for breakage as we update the theme in the future.

PR #1049, opened 2018-08-28
html-template: Replace html-webpack-template with our own template

edmorley:
The HTML template that comes with `html-webpack-plugin` [1] is very minimal and so until now we've used the more fully-featured template from `html-webpack-template` [2] instead.

However that template currently reimplements the core functionality of `html-webpack-plugin` in template markup and so has to be used with `inject` mode disabled [3]. This means:
- many `html-webpack-plugin` plugins don't work (eg the one for SRI)
- the template depends much more strongly on the internal data structures of `html-webpack-plugin`, and so can be broken by new major versions (as will be the case with `html-webpack-plugin` v4)
- the template doesn't benefit from upstream improvements made to the injected content.

In addition, `html-webpack-plugin` now supports several features that were previously only possible with a custom template (eg additional meta tags), and there are now many third party plugins to provide additional functionality (and are the preferred approach).

As such, this switches `@neutrinojs/html-template` to using its own custom template, that is virtually the same as [1], other than adding `lang` and `appMountId` support. The `mobile` option has been replaced by adding the same viewport tag via the new `meta` option.

If there are features that `html-webpack-template` supported, that our new simpler template does not, users can use their own template (`html-webpack-plugin` pushes this approach strongly) - Neutrino should cater for the 80% out of the box, not the 99%).

Unrelated to the switch away from `html-webpack-template`, I've also stopped setting `xhtml` to `true`, since XHTML compliant output isn't something that most people need.

[1] https://github.com/jantimon/html-webpack-plugin/blob/master/default_index.ejs
[2] https://github.com/jaketrent/html-webpack-template/blob/master/index.ejs
[3] https://github.com/jaketrent/html-webpack-template/issues/61

### DIFF
--- a/packages/html-template/README.md
+++ b/packages/html-template/README.md
@@ -35,16 +35,20 @@ and plug it into Neutrino:
 
 ```js
 // Using function middleware format
-const template = require('@neutrinojs/html-template');
 
 // Usage shows default values
-// Accepts options specified by HtmlWebpackTemplate
-// https://github.com/jaketrent/html-webpack-template
+// Accepts options specified by html-webpack-plugin:
+// https://github.com/jantimon/html-webpack-plugin#configuration
 neutrino.use(template, {
-  inject: false,
+  // @neutrinojs/html-template includes a custom template that has more features
+  // (eg appMountId and lang support) than the default html-webpack-plugin template:
+  // https://github.com/jantimon/html-webpack-plugin/blob/master/default_index.ejs
+  template: require.resolve('@neutrinojs/html-template/template.ejs'),
   appMountId: 'root',
-  xhtml: true,
-  mobile: true,
+  lang: 'en',
+  meta: {
+    viewport: 'width=device-width, initial-scale=1'
+  },
   minify: {
     useShortDoctype: true,
     keepClosingSlash: true,
@@ -65,15 +69,20 @@ neutrino.use(template, {
 // Using object or array middleware format
 
 // Usage shows default values
-// Accepts options specified by HtmlWebpackTemplate
-// https://github.com/jaketrent/html-webpack-template
+// Accepts options specified by html-webpack-plugin:
+// https://github.com/jantimon/html-webpack-plugin#configuration
 module.exports = {
   use: [
     ['@neutrinojs/html-template', {
-      inject: false,
+      // @neutrinojs/html-template includes a custom template that has more features
+      // (eg appMountId and lang support) than the default html-webpack-plugin template:
+      // https://github.com/jantimon/html-webpack-plugin/blob/master/default_index.ejs
+      template: require.resolve('@neutrinojs/html-template/template.ejs'),
       appMountId: 'root',
-      xhtml: true,
-      mobile: true,
+      lang: 'en',
+      meta: {
+        viewport: 'width=device-width, initial-scale=1'
+      },
       minify: {
         useShortDoctype: true,
         keepClosingSlash: true,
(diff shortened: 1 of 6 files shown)


packages/html-template/README.md
```diff
@@ -35,16 +35,20 @@ and plug it into Neutrino:
...
-// Accepts options specified by HtmlWebpackTemplate
-// https://github.com/jaketrent/html-webpack-template
+// Accepts options specified by html-webpack-plugin:
+// https://github.com/jantimon/html-webpack-plugin#configuration
 neutrino.use(template, {
-  inject: false,
+  // @neutrinojs/html-template includes a custom template that has more features
+  // (eg appMountId and lang support) than the default html-webpack-plugin template:
+  // https://github.com/jantimon/html-webpack-plugin/blob/master/default_index.ejs
+  template: require.resolve('@neutrinojs/html-template/template.ejs'),
```
eliperelman: Should we instead use `path.resolve` here?

packages/web/test/web_test.js
```diff
@@ -1,5 +1,8 @@
+import { resolve } from 'path';
+
```
eliperelman: Let's remove the extra blank lines between imports.

packages/html-template/README.md
edmorley: Yeah quite possibly

packages/web/test/web_test.js
edmorley: I intentionally added the newline, since I think it's cleaner to order (and separate) imports based on stdlib vs third-party vs local imports (we already do this for Treeherder). Ideally in the future all Neutrino imports would be sorted this way, and enforced via eslint-plugin-import.

edmorley:
> Did you say we can also still inject link and script tags via the config?

meta tags yes, via the `meta` option, `link` / `script` tags no but there are plugins for at least `script` tags and people can use a custom template if needed (see ref in commit message about 80% case etc). Open to adding support for links iff an absolute blocker, but might be best to see what feedback we get without it first? (Easy to add retrospectively)

packages/html-template/README.md
edmorley: Updated

packages/html-template/template.ejs
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html lang="<%= htmlWebpackPlugin.options.lang %>">
+  <head>
+    <meta charset="utf-8">
```
eliperelman: It's super picky, but can you make this tag self-closing? The reasoning being `xhtml` mode. If they enable xhtml mode, this would be the one place in the template that isn't compliant, while a self-closing tag will be compliant with either mode.

edmorley: Neither the stock template, nor the one we've been using until now, nor CRA's do that:
https://github.com/jantimon/html-webpack-plugin/blob/55abb36ee8cad6e8eb32fcc0848c785099316f2d/default_index.ejs#L4
https://github.com/jaketrent/html-webpack-template/blob/71257cad015f0e507da13435adef6b648d8eadc9/index.ejs#L10
https://github.com/facebook/create-react-app/blob/a671462c1089c1afe95cc1f9d4f2432d3a885806/packages/react-scripts/template/public/index.html#L4

I think XHTML compliant markup falls into "the 1%". See also a recent thread on the Django developers list about how it's not really worth achieving:
https://groups.google.com/forum/#!topic/django-developers/G4Rz1_Bmh7w

PR #1061, opened 2018-08-30
create-project: Update webpack-cli from v2 to v3

edmorley:
We hadn't received a Renovate PR for this `webpack-cli` release (which would have reminded us to bump the version in `matrix.js`), since the dependency is intentionally not listed in `peerDependencies` so that people can install their own choice of CLI if desired. To prevent this problem in the future, `webpack-cli` has now been added to the monorepo `devDependencies`, so we'll get the reminder PR next time.

https://github.com/webpack/webpack-cli/compare/v2.1.5...v.3.1.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "verdaccio": "^3.0.2",
     "verdaccio-memory": "^1.0.1",
     "webpack": "^4.8.3",
+    "webpack-cli": "^3.1.0",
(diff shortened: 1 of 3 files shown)


edmorley:
It's worth noting as well that webpack-command is no longer recommended and likely to be archived soon, given webpack/webpack#7966 - and so us sticking with webpack-cli (but not blocking people trying others) seems like the best approach for us for now.

PR #1062, opened 2018-08-30
Add devtool option to web preset

### LINKED ISSUE #676: Web preset option to enable sourcemaps in production
Currently the Neutrino web preset only enables sourcemaps (specifically `cheap-module-eval-source-map`) in development:
https://github.com/mozilla-neutrino/neutrino-dev/blob/e52919b5ebcd694857bb7a5244afe07ca1299fd1/packages/web/index.js#L187

I imagine some people wouldn't want sourcemaps in production (eg non-open-source), however perhaps it would be worth having an option to enable for those who do? (Which would save users having to figure out which sourcemap type is best for production by reading through all of https://webpack.js.org/configuration/devtool/)

eliperelman:
Fixes #676.

I will update the docs for this once we are good with the approach, unless you would like to hold off due to the in-progress docs updates.

This option can be used like so:

```js
// disable for all environments
['@neutrinojs/web', {
  devtool: false
}]

// use the same value for all environments
['@neutrinojs/web', {
  devtool: 'source-map'
}]

// change per NODE_ENV
['@neutrinojs/web', {
  devtool: {
    development: 'cheap-module-eval-source-map',
    production: 'source-map',
    test: 'cheap-eval-source-map'
  }
}]
```

### DIFF
--- a/packages/web/index.js
+++ b/packages/web/index.js
@@ -23,7 +23,7 @@ module.exports = (neutrino, opts = {}) => {
     html: {},
     devtool: {
       development: 'cheap-module-eval-source-map',
-      production: false,
+      production: undefined,
       test: 'source-map'
     },
     devServer: {
@@ -113,12 +113,10 @@ module.exports = (neutrino, opts = {}) => {
       .use(EnvironmentPlugin, [options.env]);
   }
 
-  if (options.devtool) {
-    const devtool = options.devtool[process.env.NODE_ENV];
+  const devtool = options.devtool[process.env.NODE_ENV];
 
-    if (devtool) {
-      neutrino.config.devtool(devtool);
-    }
+  if (devtool !== undefined) {
+    neutrino.config.devtool(devtool);
   }
 
   neutrino.use(htmlLoader);
--- a/packages/web/index.js
+++ b/packages/web/index.js
@@ -21,6 +21,11 @@ module.exports = (neutrino, opts = {}) => {
     env: false,
     hot: true,
     html: {},
+    devtool: {
+      development: 'cheap-module-eval-source-map',
+      production: false,
+      test: 'source-map'
+    },
     devServer: {
       hot: opts.hot !== false,
       publicPath: resolve('/', publicPath)
@@ -52,6 +57,14 @@ module.exports = (neutrino, opts = {}) => {
 
   if ('style' in options.minify) {
     throw new Error('The minify.style option has been removed. To enable style minification use the @neutrinojs/style-minify preset.');
+  }
+
+  if (typeof options.devtool === 'string' || typeof options.devtool === 'boolean') {
+    options.devtool = {
+      development: options.devtool,
+      production: options.devtool,
+      test: options.devtool
+    };
   }
 
   if (typeof options.devServer.proxy === 'string') {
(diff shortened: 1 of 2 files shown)


packages/web/index.js
```diff
@@ -54,6 +59,14 @@ module.exports = (neutrino, opts = {}) => {
     throw new Error('The minify.style option has been removed. To enable style minification use the @neutrinojs/style-minify preset.');
   }
 
+  if (typeof options.devtool === 'string' || typeof options.devtool === 'boolean') {
```
edmorley: This will also accept `true`, which will result in a webpack schema validation error (since `devtool` can only be `string` or `false`), but that's probably the best bet since at least they'll get a nice schema validation error message.

packages/web/index.js
```diff
@@ -100,6 +113,14 @@ module.exports = (neutrino, opts = {}) => {
       .use(EnvironmentPlugin, [options.env]);
   }
 
+  if (options.devtool) {
+    const devtool = options.devtool[process.env.NODE_ENV];
+
+    if (devtool) {
```
edmorley: This will mean that for `false`, `neutrino.config.devtool()` isn't called at all, which for `mode: 'development'` will mean the webpack default of `'eval'` is used, which is probably not what the user intended. I think this should be a `if (process.env.NODE_ENV in options.devtool)` or else `if (devtool !== undefined)`?

packages/web/index.js
```diff
@@ -100,6 +113,14 @@ module.exports = (neutrino, opts = {}) => {
       .use(EnvironmentPlugin, [options.env]);
   }
 
+  if (options.devtool) {
```
edmorley: I don't think this conditional is needed, since the handling above will mean that `devtool: false` will be converted to an object anyway, and thus `options.devtool` will always be truthy (other than for spurious values like `0`, `""`, `null` etc, which we shouldn't worry about handling anyway.

packages/web/index.js
```diff
@@ -21,6 +21,11 @@ module.exports = (neutrino, opts = {}) => {
     env: false,
     hot: true,
     html: {},
+    devtool: {
+      development: 'cheap-module-eval-source-map',
+      production: false,
+      test: 'source-map'
```
edmorley: This changes the `NODE_ENV=test` devtool from the previous value of `undefined` (which would have given a value of `eval` in `mode: 'development'` due to the webpack defaulter), to `'source-map'`. Have you done comparisons on the quality vs speed trade-offs of the two? (for both one-off test runs and watching; guessing it needs to be for karma since jest/mocha circumvent webpack) Since we don't minify for tests, I suppose there is slightly less benefit of 'source-map' over the cheaper options.

eliperelman: I haven't done any benchmarking, I just added this to make testing easier to debug.

edmorley: Trying with Treeherder, the webpack compilation part of `yarn test` (as judged by the duration output to the console) varied accordingly (lowest of several repeats listed):
* `devtool: false`: 9060ms
* `devtool: 'cheap-module-eval-source-map'`: 9641ms
* `devtool: 'source-map'`: 11265ms

'source-map' means a 25% increase in compile time, but still not that significant proportion of the total `yarn test` time (which was ~20s), given the time taken to spin up the browser and run the tests.

As such, `source-map` seems fine to me :-)

PR #1064, opened 2018-08-31
Added .web.js{x} resolve extensions to webpack for better react-native-web support

### LINKED ISSUE #1056: Add `.web.js` and `.web.jsx` to resolve extension
### Feature request or enhancement

We have some `react-native-web` support [here](https://github.com/neutrinojs/neutrino/blob/master/packages/react/index.js#L52) in the react-preset.
For multi-platform support react-native-web need `.web.js{x}` files to be loaded like `.android.js` and `.ios.js` files. See [here](https://github.com/necolas/react-native-web/blob/master/packages/website/guides/getting-started.md#multi-platform-applications).
We should add those extensions somewhere in a preset or middleware (`neutrino.config.resolve.extensions`).

piu130:
As described in the `react-native-web` [doc](https://github.com/necolas/react-native-web/blob/master/packages/website/guides/getting-started.md#multi-platform-applications) there is the new file extension `.web.js{x}` needed for significant platform differences.
This PR adds those extensions.
Discussion see [here](https://github.com/neutrinojs/neutrino/issues/1056).

Fixes #1056.

### DIFF
--- a/packages/react/index.js
+++ b/packages/react/index.js
@@ -50,4 +50,7 @@ module.exports = (neutrino, opts = {}) => {
   });
 
   neutrino.config.resolve.alias.set('react-native', 'react-native-web');
+
+  neutrino.config.resolve.extensions.prepend('.web.js');
+  neutrino.config.resolve.extensions.prepend('.web.jsx');
 };
--- a/packages/react/test/react_test.js
+++ b/packages/react/test/react_test.js
@@ -3,6 +3,7 @@ import { validate } from 'webpack';
 import Neutrino from '../../neutrino/Neutrino';
 
 const mw = () => require('..');
+const newExtensions = ['.web.jsx', '.web.js'];
 const originalNodeEnv = process.env.NODE_ENV;
 
 test.afterEach(() => {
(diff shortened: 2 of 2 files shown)


packages/react/index.js
```diff
@@ -50,4 +50,7 @@ module.exports = (neutrino, opts = {}) => {
   });
 
   neutrino.config.resolve.alias.set('react-native', 'react-native-web');
+
+  neutrino.config.resolve.extensions.prepend('.web.js');
```
edmorley: Could you add a comment linking to the react-native-web docs, and also a `Fixes #1056` to the commit message/PR description? :-)

piu130: Done.

piu130:
Missed that :) Done.

piu130:
Should we create a test `valid common config` so we do not have to duplicate code? Same for `web`-preset?

edmorley:
Re creating a shared test config (saw the email notification, though the message isn't above any more) - we're thinking of switching to Jest at some point (#635), which will mean we can use the snapshots feature and so avoid the manual duplication/boilerplate :-)

packages/react/index.js
kopax: Hello, I am trying hard to distribute react-native packages with extensions support from `node_modules`,  `.web.js`, `.android.js`, `.js` , I don't get how to tell react-native application to import the correct file when `node_modules`. I am not able to tell the webapp to import `.web.js` first,  too  when import happen from `node_modules`, I initialized my project with expo SDk36, do you know how I can solve my issue? Thanks again for reading.

kopax:
Hello, I am trying hard to distribute react-native packages with extensions support from `node_modules`,  `.web.js`, `.android.js`, `.js` , I don't get how to tell react-native application to import the correct file when `node_modules`. I am not able to tell the webapp to import `.web.js` first,  too  when import happen from `node_modules`, I initialized my project with expo SDk36, do you know how I can solve my issue? Thanks again for reading.

PR #1069, opened 2018-09-02
web/library: Remove worker-loader

edmorley:
Since the top-level presets aim to cover the 80% case, and web workers are still less common. If after Neutrino 9's release there ends up being demand for them, we'd be better off having a separate opt-in preset (similar to `@neutrinojs/pwa`).

### DIFF
--- a/packages/library/README.md
+++ b/packages/library/README.md
@@ -428,7 +428,6 @@ The following is a list of rules and their identifiers which can be overridden:
 | Name | Description | Environments and Commands |
 | --- | --- | --- |
 | `compile` | Compiles JS files from the `src` directory using Babel. Contains a single loader named `babel` | all |
-| `worker` | Allows importing Web Workers automatically with `.worker.*` extensions. Contains a single loader named `worker`. | all |
 
 ### Plugins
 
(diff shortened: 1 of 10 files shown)


eliperelman:
Could you elaborate on the benefit to removing this? Do you feel like our assumptions here make it difficult to override this in the future?

edmorley:
Ah good question. The reason was just that I don't think many users will be using it, and that it's another dependency that has to be downloaded and then `require()`ed during the build. It also cumulatively adds to the overall cognitive overhead of understanding the web preset and the resultant `--inspect` output.

ie: This was just a continuation of my ongoing work to slim down Neutrino 9 as much of possible compared to the accumulation of features/complexity that we'd ended up with Neutrino 8 (whose `neutrino build --inspect-new` webpack config for the React preset is 400 lines long, compared to 240 lines for the equivalent on `master`).

If you felt strongly about leaving it then, then happy to do so. Do you know of many projects that use web workers with Neutrino? :-)

eliperelman:
@edmorley I have in the past, which is why I had added it, since it gave a way to use workers with zero config. That said, I haven't used it in a while, and don't have a major issue removing it. My biggest concern was being able to import as worker with no config. 😄

edmorley:
I think in the future a combined opt-in preset for various worker types might be nice to have? (And consolidate with `@neutrinojs/pwa`?)

PR #1082, opened 2018-09-06
Test create-project buildable in development mode

### LINKED ISSUE #886: Add create-project test coverage of development configs
As found in #852, create-project currently doesn't have any test coverage of the development (ie `yarn start`) configs. Whilst they are unit-tested in the individual presets, this can still miss issues:
https://github.com/mozilla-neutrino/neutrino-dev/pull/852#discussion_r189422295

Testing the dev server means dealing with running the server, then killing it after N seconds, and detecting the difference between startup errors and the process being killed by the test runner.

A simpler alternative that might help in the meantime, would be to test a webpack build using `--mode development`, to give at least some of the coverage.

eliperelman:
Fixes #886.

This patch is a stop-gap for testing create-project output works in development mode by only overriding the build command, not start.

### DIFF
--- a/packages/create-project/test/cli_test.js
+++ b/packages/create-project/test/cli_test.js
@@ -51,7 +51,7 @@ const spawnP = (cmd, args, options) => new Promise((resolve, reject) => {
   child.stderr.on('data', data => { output += data.toString(); });
   child.on('close', code => (code === 0 ? resolve(code) : reject(output)));
 });
-const buildable = async (t, dir, args) => {
+const buildable = async (t, dir, args = []) => {
   try {
     await spawnP('yarn', ['build', ...args], { cwd: dir, stdio: 'pipe' });
     t.pass();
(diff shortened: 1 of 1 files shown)


packages/create-project/test/cli_test.js
```diff
@@ -97,16 +97,23 @@ Object.keys(tests).forEach(projectName => {
...
     assert.file(join(dir, '.neutrinorc.js'));
     assert.file(join(dir, 'webpack.config.js'));
     assert.file(join(dir, '.eslintrc.js'));
 
     await lintable(t, dir);
     await buildable(t, dir);
 
+    const pkg = require(pkgPath); // eslint-disable-line import/no-dynamic-require
+
+    if ('start' in pkg.scripts) {
```
edmorley: As far as I can tell, the `@neutrinojs/library` preset is the only one that create-project does not configure a `start` script for. 

However the library preset does have prod-only code:
https://github.com/neutrinojs/neutrino/blob/1b4f92b2f4510aa71852638a7cca59993e7d21b1/packages/library/index.js#L101-L103

...so perhaps should still be tested in development mode?

Also, is it worth adding a comment saying something about "building in development mode to emulate start, until we can come up with something better"?

eliperelman: Definitely. I'll add the comment.

PR #1087, opened 2018-09-10
Update to html-webpack-plugin 4.0.0-alpha.2

edmorley:
Notable changes:
* fully supports webpack 4, so no longer needs the sibling chunks plugin workaround.
* caches generated HTML in watch mode, significantly reducing rebuild times when using webpack-dev-server.
* includes many other performance improvements, which are particularly noticeable for multi-entrypoint projects.

For Treeherder (which has 7 entrypoints), this update reduces incremental build times by 65%!

See:
https://github.com/jantimon/html-webpack-plugin/pull/953


edmorley:
The Travis run has finished/succeeded:
https://travis-ci.org/neutrinojs/neutrino/builds/426889722

...it's just that Travis isn't syncing to GitHub:
https://www.traviscistatus.com/incidents/8ld9n73sg3h3

packages/react-components/index.js
```diff
@@ -44,7 +44,6 @@ module.exports = (neutrino, opts = {}) => {
       neutrino.use(react, options);
 
       // Remove the html generation plugins added by the react preset
```
edmorley: @eliperelman do you know why the html plugins are manually removed here, rather than just always setting the `html` preset option to `false` (which would stop the plugins being added in the first place)?

eliperelman: I'm not sure, my guess was just not thinking about it.

edmorley: I've opened #1093 for this.

PR #1105, opened 2018-09-17
Remove image-minify middleware

### LINKED ISSUE #670: `image-minify` plugin breaks json-loader

### LINKED ISSUE #1104: Decide what to do with @neutrinojs/image-minify
The Neutrino monorepo currently contains `@neutrinojs/image-minify` (added in #459, discussions also in #446), which is a preset that minifies images as part of the webpack build. It used to be also pulled into the web preset, however the integration with it was removed in #815 due to the impact on the dependency tree.

Even with that change, I still have a number of reservations about the preset.

I'm not convinced minifying images as part of the webpack build makes sense, given that:
- minifying images is time-consuming, and for most projects the images they use will change very infrequently (compared to their JS/CSS), so most of the time it adds to the build time unnecessarily.
- deciding what settings to use for image compression really needs someone to sit down and look at the before/after for each image/image type, and try a few settings - whereas it's likely with the webpack plugin someone will set it up once, and then not think to look at the build output again.
- when developing, people won't be seeing the actual images that will be used in the final build.

Even if there are cases where build-time minification makes sense, I'm not sure we should have a core Neutrino preset for it, since:
- it implies to users (many of them may be newish to the webpack ecosystem) that build-time minification is the right way to do it, which I don't think it is (see above), and they may add it to their project without second thought (given it's an official preset), and not realise it's what is slowing down their builds considerably.
- it's very hard for us to pick minification defaults that will work for everyone (eg lossless vs lossy, what compression level, ...), so many/most people will need to heavily customise the options used by the preset anyway, reducing the benefit of a preset.
- we don't know what image types (jpeg/gif/...) will be used by projects, so have to include the minification plugins for all of them (which are typically heavyweight and slow to install due to native compiled code), which defeats the advantage of [imagemin-webpack](https://github.com/itgalaxy/imagemin-webpack)'s "decide which plugins/dependencies you need" feature.
- it doesn't currently fully lazy-load plugins (it required more invasive refactoring than was worth it in #1101) so impacts eslint/test runner load times.
- it needs more `__expression` annotations before the `--inspect` feature will work correctly (all other monorepo presets support the new fully-self-contained `--inspect` other than this one and `@neutrinojs/stylelint`).
- it's currently not actively maintained (eg long-running #670, #673, #801 and a `imagemin-gifsicle` deprecation warning seen when installing `@neutrinojs/image-minify`) and the core developers are already stretched pretty thinly as it is on the rest of Neutrino.

I think for Neutrino we need to be aiming for:
- the top-level monorepo presets (eg web/react/airbnb/...) to satisfy the ~80% case.
- certain very carefully chosen specific features to be available via opt-in monorepo presets, taking into account:
  - how many projects might use it
  - ease of us picking defaults that will work for most projects
  - the complexity of people configuring the feature on their own (ie would it just be a one-liner, or are we saving them time/hassle?)
- everything else to be covered by either third-party presets, or else Neutrino API usage in people's own `.neutrinorc.js` -- and for the Neutrino docs to be clear enough to make this easy.

@eliperelman, @timkelty, thoughts?

Also, on the topic of what we include in the monorepo, perhaps it would be useful to have a policy (or at least an informal process) for how we decide whether to include new presets? For example, we could do what the Django project does, and say that first features need to be proven via a third-party package, and once any issues ironed out/a suitable demand seen, only then can it be included in core?

eliperelman:
Fixes #1104.
Closes #801.
Closes #670.


edmorley:
Could you update `yarn.lock`? (yarnpkg/yarn#5840 means CI doesn't fail if redundant lockfile entries aren't removed, which is unfortunate)

Also, could you update the commit message to include `Closes #670`?

PR #1107, opened 2018-09-18
Respect passing false to style-loader modules

### LINKED ISSUE #1068: @neutrinojs/style-loader still sets 'exclude' when 'modules: false'
Passing `{ modules: false }` to `@neutrinojs/style-loader` correctly prevents the addition of the module rules, however the non-module rule still has:

```js
      /* neutrino.config.module.rule('style') */
      {
        test: /\.css$/,
        exclude: [
          /\.module\.css$/
        ],
        // ...
```

This is because of:
https://github.com/neutrinojs/neutrino/blob/985e17992ea40618089940ccaba832cce347ec1c/packages/style-loader/index.js#L5

ie: `modules` is set to `true` even if it's passed as `false` (but then fixed after the `merge()`, which is why the rest of the functionality works as expected).

eliperelman:
Fixes #1068.

### DIFF
--- a/packages/style-loader/index.js
+++ b/packages/style-loader/index.js
@@ -44,7 +44,7 @@ module.exports = (neutrino, opts = {}) => {
         }
       })
     );
-  };
+  }
 
   rules.forEach(options => {
     const styleRule = neutrino.config.module.rule(options.ruleId);
--- a/packages/style-loader/index.js
+++ b/packages/style-loader/index.js
@@ -1,7 +1,7 @@
 const merge = require('deepmerge');
 
 module.exports = (neutrino, opts = {}) => {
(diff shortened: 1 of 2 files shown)


packages/style-loader/index.js
```diff
@@ -1,7 +1,7 @@
 const merge = require('deepmerge');
 
 module.exports = (neutrino, opts = {}) => {
-  const modules = opts.modules || true;
+  const modules = opts.modules === false ? false : opts.modules;
```
edmorley: I think this stops `modules` from defaulting to `true`?

eliperelman: Gah right, I'll fix.

eliperelman:
Tests added.

PR #1110, opened 2018-09-19
Remove stylelint middleware

### LINKED ISSUE #982: Dependency deprecation warning: stylelint-processor-html (npm)
On registry `https://registry.npmjs.org/`, the "latest" version (v1.0.0) of dependency `stylelint-processor-html` has the following deprecation notice:

`Please use stylelint-processor-arbitrary-tags https://github.com/mapbox/stylelint-processor-arbitrary-tags `

Marking the latest version of an npm package as deprecated results in the entire package being considered deprecated, so contact the package author you think this is a mistake.

Please take the actions necessary to rename or substitute this deprecated package and commit to your base branch. If you wish to ignore this deprecation warning and continue using `stylelint-processor-html` as-is, please add it to your [ignoreDeps](https://renovatebot.com/docs/configuration-options/#ignoredeps) array in Renovate config before closing this issue, otherwise another issue will be recreated the next time Renovate runs.

eliperelman:
Closes #805.
Closes #982.

We cannot seem to get a recent release of stylelint-webpack-plugin [1] in order to provide any value.

[1] https://github.com/webpack-contrib/stylelint-webpack-plugin/issues/149


edmorley:
To add some additional reasons from IRC:
1. unlike the eslint related presets in this repository, the `@neutrinojs/stylelint` preset doesn't actually have much to configure, and doesn't really change it's behaviour based on other presets (compared to say the interactions between the linting presets when using the react/preact/vue/node presets). As such it doesn't save much work over people just adding `stylelint-webpack-plugin` directly in their `.neutrinorc.js`.
2. for CSS I don't think there's as much value for "linting during the build", compared to JS. For example, one advantage of eslint-loader is that it can give a useful linting error that might prevent/explain a runtime exception that would otherwise mean opening the browser console and figuring out why the app broke. That kind of scenario is much less likely to occur with CSS. As such a `yarn lint:style` that runs the styelint CLI outside of the webpack build, or an editor/IDE plugin would likely be sufficient for many cases.
3. Given (2), one option would be for use to ditch the `stylelint-webpack-plugin` part of `@neutrinojs/stylelint` and just keep the `neutrino().stylelintrc()` config generation for use with the stylelint CLI. However given (1), it doesn't seem like that adds any value over people just using a static config in `.stylelintrc.js` and not using Neutrino for it at all.

timkelty:
@edmorley @eliperelman late response again guys, but I totally agree. Trim the fat!

PR #1117, opened 2018-09-20
Remove old docs snippets that reference webpack 3 build output

eliperelman:


eliperelman:
Hmm yeah, not sure what we should put there, if anything for output.

PR #1118, opened 2018-09-20
Add webpack-dev-server to react-components' dependencies to make it startable, make note in react-components README about conflict with eslint-plugin-import

### LINKED ISSUE #962: webpack-dev-server missing from react-components project on master
Using create-project to create a react-components project, then using `yarn start`, I get:
```
yarn run v1.7.0
$ webpack-dev-server --mode development
'webpack-dev-server' is not recognized as an internal or external command,
operable program or batch file.
error Command failed with exit code 1.
```

We should either add webpack-dev-server to the `devDependencies` matrix in create-project, or else not define the `start` script entry for react-component projects' `package.json`.

Also, if I manually install `webpack-dev-server`, I then get a bunch of `import/no-extraneous-dependencies` errors relating to deps being in `devDependencies` rather than `dependencies`.

eliperelman:
Fixes #962.

### DIFF
--- a/packages/create-project/commands/init/matrix.js
+++ b/packages/create-project/commands/init/matrix.js
@@ -63,7 +63,7 @@ const presets = {
       'prop-types@^15',
       'react@^16',
       'react-dom@^16',
-      WEBPACK, WEBPACK_CLI
+      WEBPACK, WEBPACK_CLI, WEBPACK_DEV_SERVER
     ]
   },
   [STANDARDJS]: {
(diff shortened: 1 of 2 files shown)


edmorley:
> Fixes #962.

For that issue I think we also need to either (a) stop create-project adding a `start` script entry, or (b) install `webpack-dev-server` so the `start` script works?

eliperelman:
Ah right, I'll add the dev server (b).

eliperelman:
Will do!

PR #1124, opened 2018-09-20
Add documentation notes about compiling node_modules

### LINKED ISSUE #1100: Document how to Babel-transpile dependencies in node_modules too
Currently Neutrino intentionally does not compile `node_modules` by default, since packages should be compiling down to ES5 themselves (even when using the `module` property in `package.json`). There is an increasing (and unfortunate) trend within the react-**native** packages to not do this - which resulted in confusion in #1046. (See facebook/create-react-app#1125 and facebook/create-react-app#3776 for more history on this area)

I believe the Neutrino default is currently the right thing to do (particularly since unlike CRA, it's easy to override), however it would be good to document how a project can enable the transpiling of `node_modules` - either via a single section on `@neutrinojs/compile-loader` or else a mention on all of the top-level presets.

The new section should include something like:
https://github.com/neutrinojs/neutrino/issues/1046#issuecomment-416052505

eliperelman:
Fixes #1100.

### DIFF
--- a/packages/web/README.md
+++ b/packages/web/README.md
@@ -463,6 +463,9 @@ module.exports = {
 }
 ```
 
+If the need arises, you can compile `node_modules` by referring to the relevant
+[`compile-loader` documentation](https://neutrinojs.org/packages/compile-loader/#compiling-node_modules).
+
 ### Rules
 
 The following is a list of rules and their identifiers which can be overridden:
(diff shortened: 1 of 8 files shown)


packages/web/README.md
```diff
@@ -497,6 +497,9 @@ By following the [customization guide](https://neutrinojs.org/customization/) an
 you can override and augment the build by by providing a function to your `.neutrinorc.js` use array. You can also
 make these changes from the Neutrino API in custom middleware.
 
+If the need arises, you can also compile `node_modules` by referring to the relevant
+[`compile-loader` documentation](https://neutrinojs.org/packages/compile-loader/#compiling-node_modules).
```
edmorley: For the other presets, this is under the `Customizing` section. Should this be too? (Happy either way)

eliperelman: Good idea, I'll move it.

ackvf:
The link is dead now, is it possible to find the relevant documentation anywhere?

edmorley:
For the in-release-candidate Neutrino 9 (`master` branch), it's here:
https://master.neutrinojs.org/packages/compile-loader/#compiling-node_modules

To try out v9, see #1129.

PR #1135, opened 2018-09-21
Use 'webpack-dev-server --open' for new projects

### LINKED ISSUE #1130: Use 'webpack-dev-server --open' for new projects
One of the things that CRA does that Neutrino doesn't is have the dev server page load in the browser after running `yarn start`. I think we should consider adding `--open` to the default `package.json` `"start"` script entry added to new projects by `create-project`. People can always remove it if they prefer to not have it open automatically.

edmorley:
Fixes #1130.

### DIFF
--- a/docs/migration-guide.md
+++ b/docs/migration-guide.md
@@ -76,7 +76,7 @@ module.exports = neutrino().eslintrc();
 - **BREAKING CHANGE** With the necessity of external CLIs means the removal of most of the Neutrino CLI's
 commands [#852](https://github.com/neutrinojs/neutrino/pull/852):
     - `neutrino build` can typically be replaced with `webpack --mode production`.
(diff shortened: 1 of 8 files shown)


edmorley:
An alternative to this would be to set `devServer.open` via the config:
https://webpack.js.org/configuration/dev-server/#devserver-open

Pros:
* Less clutter in `package.json`
* Would affect existing projects too

Cons:
* Less discoverable if people don't like the behaviour and want to change it
* Would affect existing projects too

I'm probably 60:40 to using `--open` (ie keeping this PR as-is), though could be persuaded either way.

edmorley:
> I am curious, does `--open` always open a new tab or open an existing one if it matches and the dev server is restarted?

As far as I know, there is no way to tell a browser to re-use an existing tab. So if one ctrl+Cs webpack-dev-server and runs it again, another tab will be opened.

PR #1138, opened 2018-09-24
Refresh lockfile & bump in-range dependency versions

edmorley:
Bumping the version ranges themselves in theory shouldn't be necessary, however it reduces the chance of package-managers consolidating deps with an older version & making it harder when we try to debug any issues user encounter.

I've also bumped the `peerDependencies` (hence being a breaking change), which should help remind users to upgrade if they are still using older versions of them.

The lockfile has also been refreshed, which fixes #1137.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,24 +34,24 @@
     "version": "[[ -z \"$SKIP_CHANGELOG\" ]] || yarn changelog --package && git add CHANGELOG.md"
   },
   "devDependencies": {
-    "auto-changelog": "^1.4.6",
+    "auto-changelog": "^1.8.0",
     "ava": "^1.0.0-beta.4",
-    "eslint": "^5.4.0",
-    "eslint-config-prettier": "^3.0.1",
+    "eslint": "^5.6.0",
+    "eslint-config-prettier": "^3.1.0",
     "eslint-plugin-prettier": "^2.6.2",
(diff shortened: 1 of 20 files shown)


eliperelman:
Upgrading `peerDependencies` could essentially force users between choosing between neutrino and a certain version of webpack, and I would prefer not to make that choice. They may be reminded to upgrade their webpack version, but if they can't then it may just be annoying for us to enforce that. I like to use `peerDependencies` to mark the minimum or breaking version of a package that works with neutrino.

edmorley:
Yeah I agree; one of the reason I decided to include it was because of this on IRC - but perhaps I misunderstood?

```
15:52 <•emorley> and I was thinking before final I could bump all the in-range deps to force latest versions
15:52 <•Eli> ++++++++
15:52 <•emorley> but none of those are urgent
16:02 <•Eli> emorley: also peerDeps
```

eliperelman:
Yeah sorry, I meant changing all the peerDeps from `"neutrino": "^9.0.0-0"` to ``"neutrino": "^9.0.0"`.

edmorley:
Ah I thought lerna did that for us - but I guess not given lerna/lerna#1575?
I'll update the PR.

edmorley:
Updated

PR #1139, opened 2018-09-25
Add RFC forwarded header as hint to backend server

### LINKED ISSUE #1089: Do not set changeOrigin on options.devServer.proxy shortcut
https://github.com/neutrinojs/neutrino/blob/master/packages/web/index.js#L72-L77

I'll do some testing, but pretty sure this is vestigial an doesn't make sense to assume.

While I'm at it…I'm wondering if it makes more sense to just use `Forwarded`…https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded

timkelty:
Fixes #1089 

## `changeOrigin: true`

Quickly looking around, it seems like it causes confusion no matter which way you default it:
- https://github.com/facebook/create-react-app/issues/4302
- https://github.com/vuejs/vue-cli/issues/1856
- https://github.com/nodejitsu/node-http-proxy/issues/1132

Originally I was inclined to remove it just to keep the defaults from webpack-dev-sever, as [I was the one who added it](https://github.com/neutrinojs/neutrino/commit/d597cbfd6c755b16a322cbe81704e23f4330e0f9#diff-f7eeb689dea27b680d5b7fd733b51ec9), and since am not sure why I needed it 😝 

However, it seems like perhaps it is more useful as `true` for more people?

CRA, for example, sets this to `true`, but then again there are [people asking that they don't](https://github.com/facebook/create-react-app/issues/4302).

So, I left it as-is. Seems maybe like a toss-up. Any input welcome.

## `X-Dev-Server-Proxy`/ `Forwarded`

This PR removes the invented/non-standard `X-Dev-Server-Proxy` in favor of a `Forwarded` header to indicate to the proxied server that this request has been proxied by webpack-dev-server.

I think I've read this too many times and proxies, being proxies, confuse me quickly…
I'm not 💯 if this should be `for:` or `by:`. Anyone good at reading spec-language? 😁 

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded
- https://tools.ietf.org/html/rfc7239#section-4

### DIFF
--- a/packages/web/index.js
+++ b/packages/web/index.js
@@ -74,7 +74,7 @@ module.exports = (neutrino, opts = {}) => {
         target: options.devServer.proxy,
         changeOrigin: true,
         headers: {
-          'X-Dev-Server-Proxy': options.devServer.proxy
+          Forwarded: 'by=_webpack-dev-server'
         }
       }
     };


timkelty:
@neutrinojs/core-contributors can anyone do a sanity-check and verify we should be using `by:` and not `for:` here?

edmorley:
> So, I left it as-is. Seems maybe like a toss-up. Any input welcome.



> CRA, for example, sets this to `true`, but then again there are [people asking that they don't](https://github.com/facebook/create-react-app/issues/4302).
> 
> So, I left it as-is. Seems maybe like a toss-up. Any input welcome.

I think we should leave it enabled. There seems to only be one person asking for CRA to change the default, and presumably given the issues mentioned in some of the other linked issues/SO posts, there would be many more filed for the reverse, were CRA to ever change their default.

> can anyone do a sanity-check and verify we should be using `by:` and not `for:` here?

Reading the spec I'm pretty sure `by:` is the correct one :-)

PR #1145, opened 2018-09-27
html-template: Update to html-webpack-plugin@4.0.0-beta.1

edmorley:
The new version includes jantimon/html-webpack-plugin#1048, which means we can remove the temporary additions made in #1121.

### DIFF
--- a/packages/html-template/README.md
+++ b/packages/html-template/README.md
@@ -50,16 +50,6 @@ neutrino.use(template, {
   meta: {
     viewport: 'width=device-width, initial-scale=1'
   },
-  // These are copied from the new html-webpack-plugin defaults (not yet released):
-  // https://github.com/jantimon/html-webpack-plugin/pull/1048
-  minify: process.env.NODE_ENV === 'production' && {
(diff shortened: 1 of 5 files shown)


eliperelman:
@edmorley could you tell me how this is a breaking change?

edmorley:
It's only breaking in a pedantic sense - the upstream implementation (that I wrote) uses `mode` to determine whether "in production", whereas my temporary workaround used `NODE_ENV`. Most of the time the two approaches will have the same result since Neutrino forces the two to match, apart from in the `else` case here:
https://github.com/neutrinojs/neutrino/blob/6948fc3907879152627445e06a94dfff9d543e2c/packages/neutrino/index.js#L20-L25

edmorley:
Oh actually, webpack defaults `mode` to production itself, so this can never be breaking.

edmorley:
Merging this now since otherwise Renovate will shortly be opening up a duplicate PR (but without the cleanup changes). Happy to make any changes from further review comments in a follow-up if needed.

PR #1149, opened 2018-09-28
Allow using browserlistsrc for targets

### LINKED ISSUE #1141: Allow browserslistsrc file to configure babel targets
A while back I brought this up: https://github.com/neutrinojs/neutrino/issues/229

It is desirable to have a browserslist file so various tools can pick use it (autoprefixer, stylelint, eslint-plugin-compat, and of course, @babel/preset-env).

`@babel/preset-env` supports this out of the box: https://babeljs.io/docs/en/babel-preset-env#browserslist-integration, however, since Neutrino passes targets directly, I'm assuming it will take those values and not read them from the file.

I'm not sure if this is a situation where we can use `neutrinorc` as the single source of truth, as `.browserslistrc` files, I believe are just plain text (no js).

timkelty:
Resolves #1141

Allows users to pass `false` to `options.targets` or `options.targets.browsers` to have babel-env use a `.browserlistsrc` file.

### DIFF
--- a/packages/web/index.js
+++ b/packages/web/index.js
@@ -89,11 +89,6 @@ module.exports = (neutrino, opts = {}) => {
     };
   }
 
-  // Force @babel/preset-env default behavior (.browserslistrc)
-  if (options.targets === false) {
-    options.targets = {};
-  }
-
   Object.assign(options, {
     style: options.style && merge(options.style, {
       extract: options.style.extract === true ? {} : options.style.extract
--- a/packages/web/index.js
+++ b/packages/web/index.js
@@ -94,7 +94,7 @@ module.exports = (neutrino, opts = {}) => {
       'last 2 iOS versions'
     ];
   }
-  
+
   Object.assign(options, {
     style: options.style && merge(options.style, {
       extract: options.style.extract === true ? {} : options.style.extract
--- a/packages/web/index.js
+++ b/packages/web/index.js
@@ -39,7 +39,16 @@ module.exports = (neutrino, opts = {}) => {
       source: isProduction
     },
     babel: {},
-    targets: {},
+    targets: {
+      browsers: [
+        'last 2 Chrome versions',
+        'last 2 Firefox versions',
+        'last 2 Edge versions',
+        'last 2 Opera versions',
+        'last 2 Safari versions',
+        'last 2 iOS versions'
+      ]
+    },
     font: {},
     image: {}
   }, opts);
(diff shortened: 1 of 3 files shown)


packages/web/index.js
```diff
@@ -90,7 +94,7 @@ module.exports = (neutrino, opts = {}) => {
       'last 2 iOS versions'
     ];
   }
-
+
```
eliperelman: Can you revert the change to this line?

packages/web/index.js
```diff
@@ -80,7 +80,11 @@ module.exports = (neutrino, opts = {}) => {
     };
   }
 
-  if (!options.targets.node && !options.targets.browsers) {
+  if (options.targets === false) {
+    options.targets = {};
+  } else if (options.targets.browsers === false) {
```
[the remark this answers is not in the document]
timkelty: Yeah - I did it this way because it seemed like `options.targets` had other props besides browsers that maybe you'd want to set, but still use the browerslistrc. Not totally sure of that's valid though, will have to check when I'm back on a machine: https://babeljs.io/docs/en/babel-preset-env#options

packages/web/index.js
```diff
@@ -80,7 +80,11 @@ module.exports = (neutrino, opts = {}) => {
     };
   }
 
-  if (!options.targets.node && !options.targets.browsers) {
+  if (options.targets === false) {
```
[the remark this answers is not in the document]
timkelty: Will do

packages/web/index.js
timkelty: Whups

timkelty: Whups

timkelty:
@edmorley I think you're right – we only need to support `options.targets` as `false`, not `options.targets.browsers`. Updated PR.

timkelty:
Actually, if we set the defaults up-top like this, we don't even need to account for `false`, and have users just pass `{}` if they want to overrides defaults. Thoughts?

edmorley:
>  Seem simpler. Thoughts?

The preset default options are merged with the ones the user passes using deepmerge, so I think the new implementation will prevent people from overriding. Some tests would confirm and likely be useful edither way :-)

timkelty:
> The preset default options are merged with the ones the user passes using deepmerge, so I think the new implementation will prevent people from overriding

@edmorley right you were :) Reverted that and added some tests.

PR #1152, opened 2018-10-04
fix resolution of .web.js files

MrLoh:
The config seems to be broken.

### DIFF
--- a/packages/react/index.js
+++ b/packages/react/index.js
@@ -52,8 +52,8 @@ module.exports = (neutrino, opts = {}) => {
   neutrino.config.resolve
     .alias
       .set('react-native', 'react-native-web')
-      .end()
-    .extensions
+      .end();
+  neutrino.config.resolve.extensions
       .prepend('.web.js')
       .prepend('.web.jsx');
 };


edmorley:
Hi! The current test for this is passing on `master`:
https://github.com/neutrinojs/neutrino/blob/66fa422cb1daacc2cb7f19c109ffd93cb8500801/packages/react/test/react_test.js#L6

And inspecting the config for a React project on mine that's using `9.0.0-beta.0`, I see:

```js
  resolve: {
    alias: {
      'react-native': 'react-native-web'
    },
    extensions: [
      '.web.jsx',
      '.web.js',
      '.wasm',
      '.mjs',
      '.jsx',
      '.js',
      '.json'
    ]
  },
```

Is something not working as expected for you?

MrLoh:
Yeah, it didn't work on my project. Was still loading the `.js` and not the `.web.js` file. After I added the above line in my config, it started working. Maybe it's an issue that the prepend doesn't properly work at that point?

edmorley:
What version of Neutrino are you using? (The `.web.js{x}` extensions support was only added in Neutrino 9, currently in beta). If Neutrino 9, could you paste your `.neutrinorc.js`?

PR #1154, opened 2018-10-07
compile-loader: Set 'configFile' to false by default

### LINKED ISSUE #1147: Add 'configFile: false' to babel-loader options
Since the existing `babelrc: false` doesn't apply to the new `babel.config.js` files:
https://babeljs.io/docs/en/next/config-files#project-wide-configuration

See also:
facebook/create-react-app#5109

edmorley:
To prevent stray `babel.config.js` files from affecting the Neutrino babel configuration. (The existing `babelrc` option only applies to `.babelrc` files.)

See:
https://babeljs.io/docs/en/next/config-files#project-wide-configuration

Fixes #1147.

### DIFF
--- a/packages/compile-loader/index.js
+++ b/packages/compile-loader/index.js
@@ -11,6 +11,7 @@ module.exports = (neutrino, options = {}) => {
     .options({
       cacheDirectory: true,
       babelrc: false,
+      configFile: false,
       ...(options.babel || {})
     });
 


edmorley:
(Whilst this has the breaking label, it's not breaking for users upgrading from Neutrino 8, since Babel 6 didn't support `babel.config.js` anyway)

PR #1158, opened 2018-10-09
Switch to terser-webpack-plugin

### LINKED ISSUE #1146: Switch to terser-webpack-plugin
We should switch from the webpack 4 default of [uglifyjs-webpack-plugin](https://github.com/webpack-contrib/uglifyjs-webpack-plugin) to [terser-webpack-plugin](https://github.com/webpack-contrib/terser-webpack-plugin) (which will be the default in webpack 5, see webpack/webpack#8036), since the former uses [uglify-es](https://www.npmjs.com/package/uglify-es), which doesn't have the fix for:
terser-js/terser#50

Apparently the perf issue with uglify-es is particularly bad for AngularJS projects, or people using Font Awesome 5 (adding minutes to the build time).

To do this, we'll need to set `config.optimization.minimizer` for the first time, which means we should fix neutrinojs/webpack-chain#95 first.

edmorley:
Since the webpack 4 default of `uglifyjs-webpack-plugin` uses `uglify-es`, which has been deprecated in favour of `terser`. webpack 5 is due to use `terser-webpack-plugin`, at which point we can remove this.

`terser` includes multiple correctness and performance fixes not present in `uglify-es`.

Fixes #1146.

### DIFF
--- a/packages/web/README.md
+++ b/packages/web/README.md
@@ -536,38 +536,31 @@ module.exports = {
 
 #### Source minification
 
-By default script sources are minified in production only, and using webpack's default of
-[uglifyjs-webpack-plugin](https://github.com/webpack-contrib/uglifyjs-webpack-plugin)
-(which internally uses `uglify-es`). To customise the options passed to `UglifyJsPlugin`
+By default script sources are minified in production only, using
+[terser-webpack-plugin](https://github.com/webpack-contrib/terser-webpack-plugin)
+(which replaces `uglifyjs-webpack-plugin`). To customise the options passed to `TerserPlugin`
 or even use a different minifier, override `optimization.minimizer`.
 
-Note: If switching to [babel-minify-webpack-plugin](https://github.com/webpack-contrib/babel-minify-webpack-plugin)
-ensure that sourcemaps are disabled in production to avoid [this bug](https://github.com/webpack-contrib/babel-minify-webpack-plugin/issues/68).
-
-_Example: Use different options with `uglify-es`:_
+_Example: Adjust the `terser` minification settings:_
 
 ```js
-const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
-
 module.exports = {
   use: [
     '@neutrinojs/web',
     (neutrino) => {
-      neutrino.config
-        .optimization
-          .minimizer([
-            // Based on:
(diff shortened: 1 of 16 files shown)


packages/web/README.md
```diff
@@ -536,38 +536,31 @@ module.exports = {
...
 
-By default script sources are minified in production only, and using webpack's default of
-[uglifyjs-webpack-plugin](https://github.com/webpack-contrib/uglifyjs-webpack-plugin)
-(which internally uses `uglify-es`). To customise the options passed to `UglifyJsPlugin`
+By default script sources are minified in production only, using
+[terser-webpack-plugin](https://github.com/webpack-contrib/terser-webpack-plugin)
+(which replaces `uglifyjs-webpack-plugin`). To customise the options passed to `TerserPlugin`
 or even use a different minifier, override `optimization.minimizer`.
 
-Note: If switching to [babel-minify-webpack-plugin](https://github.com/webpack-contrib/babel-minify-webpack-plugin)
```
edmorley: I've intentionally removed this reference to babel-minify-webpack-plugin, since it's still virtually unmaintained/buggy/slow. If that changes in the future we can re-add.

edmorley:
I was thinking about this PR some more today. Since we manually configure `terser-webpack-plugin`, we have to choose whether to set `sourceMaps` to true/false. This is fine if people use the in-built web preset `devtool` option, however for Neutrino 8 upgraders it's likely they may have:

```js
module.exports = {
  use: [
    ['@neutrinojs/react', {
      // ...
    }],
    (neutrino) => {
      if (process.env.NODE_ENV === 'production') {
        neutrino.config.devtool('source-map');
      }
    },
  ]
};
```

In this case, we'll pass `sourceMaps: false` to `terser-webpack-plugin`.

The only thing I think we can do, is to perhaps add a warning (or throw) in the webpack() handler, like we do for the `vendor` entrypoint check, which checks whether `devtool` and `sourceMap` are out of sync (if the `terser` plugin found).

@eliperelman, thoughts?

eliperelman:
@edmorley I think that solution will work, but we should probably use a warning since it may be desired by the user to have them out of sync for some reason.

edmorley:
I've updated the PR:
* rebased on `master`'s webpack-chain v5 upgrade
* added a check for misconfiguration in `handlers.js` (see comment there for why throwing rather than a `console.error()`)
* added migration guide changes

PR #1160, opened 2018-10-09
Improvements to the Neutrino 9 migration UX

edmorley:
* An error is now shown when using the `links` HTML option without also specifying a custom template, so that users affected by #1049 know what changes they need to make. See:
  https://github.com/neutrinojs/neutrino/issues/1129#issuecomment-428169383
* The migration guide now recommends regenerating the lockfiles.
* The migration section about supported Node.js versions has been made clearer. See:
  https://github.com/neutrinojs/neutrino/issues/1129#issuecomment-423796131
* The `neutrino` bin script error message now mentions `test` being removed, includes `--mode` for the inspect example, and uses the correct URL for the migration guide.
* The `minify.image` option deprecation error message no longer refers to `@neutrinojs/image-minify`, since it was removed. I've linked to the issue since it gives more background as to why compile-time minification is not recommended.
* The `hotEntries` option deprecation check now always occurs, and not only when `options.hot` is enabled. A test has also been added.

### DIFF
--- a/docs/migration-guide.md
+++ b/docs/migration-guide.md
@@ -39,12 +39,15 @@ neutrino --inspect --mode production > v9.config
 git diff --no-index v8.config v9.config
 ```
 
-Please [file an issue](https://github.com/neutrinojs/neutrino/issues) if any issue arises from upgrade
-that may not be outlined here.
+**We strongly recommend that any Yarn/NPM lockfiles (eg `yarn.lock` or `package-lock.json`) are
+removed and regenerated when upgrading, to prevent problems caused by leftover dependencies.**
+
+Please [file an issue](https://github.com/neutrinojs/neutrino/issues) if any issue arises from the
+upgrade that may not be outlined here.
 
 A list of changes is detailed below for migrating:
 
-- **BREAKING CHANGE** The minimum supported Node.js version is 8.10. Node.js 9 is no longer supported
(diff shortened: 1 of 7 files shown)


docs/migration-guide.md
```diff
@@ -39,12 +39,15 @@ neutrino --inspect --mode production > v9.config
 git diff --no-index v8.config v9.config
 ```
 
-Please [file an issue](https://github.com/neutrinojs/neutrino/issues) if any issue arises from upgrade
-that may not be outlined here.
+**We strongly recommend that any Yarn/NPM lockfiles (eg `yarn.lock` or `package-lock.json`) are
```
eliperelman: nit: "npm" is lowercase.

packages/html-template/index.js
```diff
@@ -1,6 +1,16 @@
 const { resolve } = require('path');
 
 module.exports = (neutrino, { pluginId = 'html', ...options } = {}) => {
+  if ('links' in options && !('template' in options)) {
+    throw new Error(
+      'The default Neutrino HTML template no longer supports the "links" option. ' +
+      'To set a favicon use the "favicon" option instead. For stylesheets either ' +
+      'import the equivalent NPM package from JS, or if using a CDN is preferred, ' +
```
eliperelman: nit: s/NPM/npm

Issue #964, opened 2018-06-25
Support configuring multiple instances of @neutrinojs/style-loader from @neutrinojs/web


timkelty:
`test` and `modulesTest` are top-level options, but they should also be able to be set-per rule, if you don't want all loaders for all tests.

Report:
https://spectrum.chat/thread/45508553-d99b-46ff-a2b8-15910326f0ae

timkelty:
I  _think_ we'd just have to add 
Hmm…or should the prescribed way of doing this just be to use style-loader twice?

mrwillis:
I think one application is best. That way it works well with other presets(react, vue, etc) with the style property. Otherwise its kind of awkward and perhaps confusing if they see another style preset applied.

timkelty:
@edmorley This would be nice for v9 if we can sneak it in. I'll see if I can fit some work in.

timkelty:
Came up again here: https://spectrum.chat/thread/5770e7d3-f2b1-4c04-b3c5-f16ccb9e3d82

timkelty:
So I started down this path but realized it doesn't make a ton of sense, because of the structure of `module.rules`. In the end we'd _need_ multiple rules with their own `test` and corresponding loaders, so it would be same end result as using `@neutrinojs/style-loader` a second time (as shown [here](https://spectrum.chat/thread/45508553-d99b-46ff-a2b8-15910326f0ae?m=MTUyOTkyNDU1MTgyMg==).

@mrwillis's [comment](https://github.com/neutrinojs/neutrino/issues/964#issuecomment-401490737) is valid though, as other middleware wouldn't know about the `style-2` rule (or whatever). Not sure Neutrino has a solution for that?

Might be that it isn't that big of a deal…/cc @edmorley @eliperelman

edmorley:
It seems like the use-case of wanting to customise the style rules is pretty common (compared to customisation of other parts of the configuration), and something that is not overly straightforwards at present.

Perhaps we could add some additional options to `@neutrinojs/web` under `style`, which if used, cause it to use `@neutrinojs/style` twice? There's probably more we could do with the docs as well, with additional "how to" sections.

What would help with this is gathering a list of common configurations, since as someone who doesn't use anything but the default style loaders, I'm having trouble visualising the desired end result.

timkelty:
> with additional "how to" sections

Yep. Adding an FAQ/How To would be nice: https://github.com/neutrinojs/neutrino/issues/747

[...]

PR #1162, opened 2018-10-10
Add checks for middleware being used twice unintentionally

edmorley:
To try and prevent the confusion caused when users do things like:
* using `@neutrinojs/react` followed by `@neutrinojs/web`, which clobbers the React parts of the config:
  https://github.com/neutrinojs/neutrino/issues/1129#issuecomment-428169383
* using `@neutrinojs/react` followed by `@neutrinojs/style`, which causes incorrect `extract` and `hot` configuration:
  https://github.com/neutrinojs/neutrino/issues/1129#issuecomment-423947761
  https://github.com/neutrinojs/neutrino/issues/755#issuecomment-374913991
  https://github.com/neutrinojs/neutrino/issues/803#issuecomment-388445048
  https://github.com/neutrinojs/neutrino/issues/869#issue-322397071

### DIFF
--- a/packages/compile-loader/index.js
+++ b/packages/compile-loader/index.js
@@ -1,13 +1,9 @@
 const babelMerge = require('babel-merge');
+const { DuplicateRuleError } = require('neutrino/errors');
 
 module.exports = (neutrino, { ruleId = 'compile', useId = 'babel', ...options } = {}) => {
   if (neutrino.config.module.rules.has(ruleId)) {
-    throw new Error(
-      '@neutrinojs/compile-loader has been used twice with the same `ruleId`.\n' +
-      'If you are including this preset manually to customise the compile rules\n' +
-      "configured by another preset, instead use that preset's own options to do so\n" +
-      '(such as the `babel` option when using the Neutrino web/react/node/... presets).'
-    );
+    throw new DuplicateRuleError('@neutrinojs/compile-loader', ruleId);
   }
 
   neutrino.config.module
--- a/packages/compile-loader/index.js
+++ b/packages/compile-loader/index.js
@@ -1,12 +1,21 @@
 const babelMerge = require('babel-merge');
 
-module.exports = (neutrino, options = {}) => {
+module.exports = (neutrino, { ruleId = 'compile', useId = 'babel', ...options } = {}) => {
+  if (neutrino.config.module.rules.has(ruleId)) {
+    throw new Error(
+      '@neutrinojs/compile-loader has been used twice with the same `ruleId`.\n' +
+      'If you are including this preset manually to customise the compile rules\n' +
+      "configured by another preset, instead use that preset's own options to do so\n" +
+      '(such as the `babel` option when using the Neutrino web/react/node/... presets).'
+    );
+  }
+
   neutrino.config.module
-    .rule(options.ruleId || 'compile')
+    .rule(ruleId)
     .test(options.test || neutrino.regexFromExtensions())
     .when(options.include, rule => rule.include.merge(options.include))
     .when(options.exclude, rule => rule.exclude.merge(options.exclude))
-    .use(options.useId || 'babel')
+    .use(useId)
     .loader(require.resolve('babel-loader'))
     .options({
       cacheDirectory: true,
@@ -17,8 +26,8 @@ module.exports = (neutrino, options = {}) => {
 
   neutrino.register('babel', (neutrino) =>
     neutrino.config.module
-      .rule('compile')
-      .use('babel')
+      .rule(ruleId)
+      .use(useId)
       .get('options')
   );
 };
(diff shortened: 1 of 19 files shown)


timkelty:
> using @neutrinojs/react followed by @neutrinojs/style, which causes incorrect extract and hot configuration:

@edmorley there is some crossover here with #964.

I've had a few people asking how they can configure `@neutrinojs/style ` to have separate _rules_ (e.g. one less and one sass), each with different `test` and loaders.

…and you can't. My only suggestion to them thus far was doing exactly what you're referencing here: use `@neutrinojs/style` _after_ `@neutrinojs/web|react|vue` and configure a second separate rule.

edmorley:
> My only suggestion to them thus far was doing exactly what you're referencing here: use `@neutrinojs/style` _after_ `@neutrinojs/web|react|vue` and configure a second separate rule.

The check in the PR uses `neutrino.config.module.rules.has(options.ruleId)`, so if I've understood the workaround for #964 correctly, it shouldn't be affected, since people are using a different `ruleId` (otherwise the rules would have been overwritten anyway).

However perhaps the wording of the error message show needs to say "... Or if using twice is intentional, remember to set a different `ruleId`" etc? (I just thought it was implied given the first sentence of the message, but perhaps best to be explicit?)

packages/compile-loader/index.js
```diff
@@ -1,12 +1,21 @@
 const babelMerge = require('babel-merge');
 
-module.exports = (neutrino, options = {}) => {
+module.exports = (neutrino, { ruleId = 'compile', useId = 'babel', ...options } = {}) => {
+  if (neutrino.config.module.rules.has(ruleId)) {
+    throw new Error(
+      '@neutrinojs/compile-loader has been used twice with the same `ruleId`.\n' +
+      'If you are including this preset manually to customise the compile rules\n' +
+      "configured by another preset, instead use that preset's own options to do so\n" +
+      '(such as the `babel` option when using the Neutrino web/react/node/... presets).'
```
timkelty: I'm confused by the double quote placement here…

packages/compile-loader/index.js
```diff
@@ -1,12 +1,21 @@
 const babelMerge = require('babel-merge');
 
-module.exports = (neutrino, options = {}) => {
+module.exports = (neutrino, { ruleId = 'compile', useId = 'babel', ...options } = {}) => {
+  if (neutrino.config.module.rules.has(ruleId)) {
+    throw new Error(
+      '@neutrinojs/compile-loader has been used twice with the same `ruleId`.\n' +
```
timkelty: Perhaps the other messages should be updated with this langauge (referencing "with the same `ruleId`")?

packages/compile-loader/index.js
edmorley: It avoids having to escape the single quote in the middle of the string (it's what https://prettier.io/playground/ outputs when using the "single quote" option and this string).

packages/compile-loader/index.js
edmorley: I omitted the "same ruleId" part of the message for middleware that doesn't allow customisation of the ruleId, since it seemed misleading to hint that the middleware could be used with different ruleIds when that's not possible. However there may be a better way to word all of them - open to ideas :-)

timkelty: Oh gotcha – looks good then.

timkelty:
@edmorley My only thought is you could consider simplifying the messages and including a link to a page in the docs detailing more specifics.

edmorley:
Yeah true. I think the reason I didn't initially, is that there isn't a good permalink for many of the settings, since they appear in the middle of "here are all the options" sections. A hardcoded URL would also get out of date once we release new major versions of Neutrino (though we have this problem already in a few places).

packages/compile-loader/index.js
```diff
@@ -17,8 +26,8 @@ module.exports = (neutrino, options = {}) => {
 
   neutrino.register('babel', (neutrino) =>
     neutrino.config.module
-      .rule('compile')
-      .use('babel')
+      .rule(ruleId)
```
eliperelman: Good catch, this was definitely a bug.

eliperelman:
Another way we could have a more generic error message would be to subclass Error:

```js
class DuplicateRuleError extends Error {
  constructor(name, ruleId) {
    super(
      `${name} has been used twice with the same `ruleId` of ${ruleId}.\n` +
      'If you are including this preset manually to customise the rules\n' +
      "configured by another preset, instead use that preset's own options to do so."
    );
  }
}

// ...

if (neutrino.config.module.rules.has(ruleId)) {
  throw new DuplicateRuleError('@neutrinojs/compile-loader', 'compile');
```

edmorley:
Yeah that may be better (I'd decided against it before since each message was subtly different eg the preset option name, but perhaps that's unnecessary). Should `DuplicateRuleError` live in `neutrino`?

eliperelman:
> Should `DuplicateRuleError` live in `neutrino`?

That would be my recommendation since every middleware we have peer-depends on it.

edmorley:
PR updated with a new `ConfigurationError` and `DuplicateRuleError`

PR #1165, opened 2018-10-11
dev-server: Stop setting contentBase

edmorley:
By default `webpack-dev-server` attempts to serve webpack-generated files in response to requests, followed by passing them on to any proxy (if set). However when the `contentBase` setting is used, requests can also fall back to any file in the specified directory, even if it's not part of the webpack build.

This can cause confusion since when performing a production build, `webpack-dev-server` isn't used so no such fallback occurs, meaning a seemingly working dev site then might not work in production.

The `contentBase` option has been set since Neutrino 4:
https://github.com/neutrinojs/neutrino/commit/30e3f5ba2cf7ed99ffc9295dfb3fb53cb3b98c68#diff-92e7b3255c5e8759497a7b3247b2e1e8R161

...and I believe is a leftover from the time when `copy-webpack-plugin` (prior to v2.0.1) didn't support making the copied files available to `webpack-dev-server`.


edmorley:
Thank you for the review(s) :-)

Issue #1171, opened 2018-10-12
Consider setting output.publicPath to '/' by default


edmorley:
The [output.publicPath](https://webpack.js.org/configuration/output/#output-publicpath) option controls the URL prefix used to refer to generated assets. webpack defaults it to `''` (`@neutrinojs/web` currently overrides this to `'./'` - however that's equivalent), which means that the generated HTML refers to assets using relative paths, eg `index.js`.

However if sites are using the HTML 5 `pushState` history API, and have page routes under nested paths (eg `/foo/bar` or even just `/foo/`), then even if they correctly configure their webserver/hosting redirects (ie to serve `index.html` for requests that would otherwise 404), requests to assets will fail.

For example, without `publicPath` manually set in their `.neutrinorc.js`, the page at https://health.graphics/android/ (which is really the contents of `/index.html`) would make a request to the non-existent:
`https://health.graphics/android/runtime.1ca11f96.js`

...rather than the actual URL of:
`https://health.graphics/runtime.1ca11f96.js`

This affects both production and development (the latter possible due to a webpack-dev-server bug).

The downside to setting `publicPath` by default, is that for apps that don't use `pushState` and are not hosted at the site root (eg `foo.com/docs/` or `user.github.io/repo/`), then they'll have to override `publicPath` otherwise their site won't work at all in production. However if these sites were using `pushState`, then they already had to be setting `publicPath` to the correct path (eg `'/docs/'`) otherwise reloads wouldn't have worked anyway. (And at least the new failure mode is more obvious)

So the question is, for `@neutrinojs/web` do we:
1. Do nothing (ie continue to set `historyApiFallback: true` giving the impression `pushState` will work, when it really won't, without customising `publicPath`)
2. Just add some docs about needing to set `publicPath` for `pushState` sites to work with direct links
3. Same as (2) but also stop setting `historyApiFallback` and add it to the list of things we say they have to set in the docs (at least the default config would be less misleading)
4. Give a warning/info message if projects don't explicitly set a `publicPath`. We could also have create-project prompt them for it. Though this would be unnecessary noisy for all projects that are at the site root.
5. Same as (4) but make it an error if not set. Though this is pretty disruptive.
6. Change the default to `'/'` so that `historyApiFallback: true` works, and just add docs for people who are not hosting at site root. We could optionally prompt during create-project and if they add a custom value that gets set, otherwise no addition to `.neutrinorc.js`.
[...]

PR #1175, opened 2018-10-13
Clean up dev-server options

### LINKED ISSUE #1169: Several issues with the default @neutrinojs/dev-server configuration
Looking deeper into the devServer default configuration (and following on from #1165, #1168 and https://github.com/neutrinojs/neutrino/pull/434#issuecomment-429091460), I've found a few issues and also potential for cleanup in `@neutrinojs/dev-server`:
* the [host](https://webpack.js.org/configuration/dev-server/#devserver-host) setting tries to default to only listening on localhost, however there is a bug in `isLocal()` which means it returns `0.0.0.0` for a `host` that has a port number (eg `localhost:5000`). This means that all `yarn start` invocations are actually publicly exposed by default.
* the [headers](https://webpack.js.org/configuration/dev-server/#devserver-headers-) option is set to `host: publicHost` (eg `host: 'localhost:5000'`), but this option is used for *responses* not *requests*, so it's not clear why that's needed? (ie it's not for proxy requests)
* it sets [public](https://webpack.js.org/configuration/dev-server/#devserver-public), which as far as I can see is only necessary in more advanced workflows (eg proxying requests to webpack-dev-server using nginx), which we can probably remove and let the 1% of people who need it pass the option explicitly
* it sets [publicPath](https://webpack.js.org/configuration/dev-server/#devserver-publicpath-) to `'/'` and [https](https://webpack.js.org/configuration/dev-server/#devserver-https) to `false`, which are unnecessary since they are the default
* the docs section that lists the middleware default options is out of date (eg `stats` and others)

To be honest, I think we should remove most of the custom handling and then just have a few examples in the docs (eg for setting `0.0.0.0` if using from within a docker container). Looking at the git history there have been many regressions due to the complex handling, and I believe it may no longer be required, now that we're using the CLI and not having to:
* patch in the client to the entrypoint:
  https://github.com/neutrinojs/neutrino/pull/852/files#diff-f7eeb689dea27b680d5b7fd733b51ec9L234
* generate the URL for the "open" feature:
  https://github.com/neutrinojs/neutrino/pull/852/files#diff-58db6dba04eacc1bb7d5628fbf89929dL71

...which would have made it hard for someone to configure themselves.

In addition for `@neutrinojs/web` we should:
* see if we can use the empty string for [output.publicPath](https://webpack.js.org/configuration/output/#output-publicpath) instead of `'./'`
* add a comment explaining the need for the `url.resolve()` for [devServer.publicPath](https://webpack.js.org/configuration/dev-server/#devserver-publicpath-)

@eliperelman / @timkelty, thoughts?

edmorley:
* (BREAKING) Stops setting `host` / `public`, since they are only needed in a minority of use-cases (such as running inside a Docker container or having a proxy in front of webpack-dev-server), and the current implementation for them in `@neutrinojs/dev-server` is buggy.
* (BREAKING) Stops supporting the `@neutrinojs/web` shorthand of setting `devServer.proxy` to a string, since it could only be used if the options it set were exactly what were required - which was often not the case given that proxy configuration is very project-specific.
* Stops setting a `host` header on *responses* to the client, since it makes no sense (particularly it was set to the public host). It's possible that this was set thinking they would be set for *requests* to a proxy target, but that's not the case (and for that, `changeOrigin: true` would be more appropriate anyway).
* Changes `output.publicPath` from `'./'` to `''` since the latter is equivalent (and is also the webpack default), however it correctly allows webpack-dev-server to fall back to `'/'` when needed:
  https://github.com/webpack/webpack-dev-server/blob/v3.1.9/lib/Server.js#L176
* Stops setting `devServer.publicPath` explicitly since it inherits from `output.publicPath` so should not normally need to be overridden:
  https://github.com/webpack/webpack-dev-server/blob/v3.1.9/bin/webpack-dev-server.js#L142-L152
* Stops setting `https: false`, since that's the default.

Fixes #1169.

### DIFF
--- a/packages/dev-server/README.md
+++ b/packages/dev-server/README.md
@@ -11,6 +11,7 @@
 - Yarn v1.2.1+, or npm v5.4+
 - Neutrino 9
 - webpack 4
+- webpack-dev-server 3
 
 ## Installation
 
@@ -42,27 +43,15 @@ neutrino.use(devServer);
 
 // Usage with custom options (default options are shown)
 neutrino.use(devServer, {
-  https: false,
   port: 5000,
-  host: 'localhost',
-  public: 'localhost:5000',
   hot: true,
+  // Redirect 404s to index.html, so that apps that use the HTML 5 History API work.
   historyApiFallback: true,
-  publicPath: '/',
-  headers: {
-    host: 'localhost:5000'
-  },
+  // Only display compile duration and errors/warnings, to reduce noise when rebuilding.
   stats: {
-    assets: false,
-    children: false,
-    chunks: false,
-    colors: true,
+    all: false,
     errors: true,
-    hash: false,
-    modules: false,
-    publicPath: false,
-    timings: false,
-    version: false,
+    timings: true,
     warnings: true
   }
 });
@@ -80,27 +69,15 @@ module.exports = {
 module.exports = {
   use: [
     ['@neutrinojs/dev-server', {
-      https: false,
       port: 5000,
-      host: 'localhost',
-      public: 'localhost:5000',
       hot: true,
+      // Redirect 404s to index.html, so that apps that use the HTML 5 History API work.
       historyApiFallback: true,
-      publicPath: '/',
-      headers: {
-        host: 'localhost'
-      },
+      // Only display compile duration and errors/warnings, to reduce noise when rebuilding.
       stats: {
-        assets: false,
-        children: false,
-        chunks: false,
-        colors: true,
+        all: false,
         errors: true,
-        hash: false,
-        modules: false,
-        publicPath: false,
-        timings: false,
-        version: false,
+        timings: true,
         warnings: true
       }
     }]
@@ -111,9 +88,6 @@ module.exports = {
 By default this middleware will start a development server with Hot Module Replacement support on
 `http://localhost:5000`. To enable HMR with your application, read the documentation of corresponding Neutrino
 preset or middleware.
-
-It is recommended to call this middleware only in development mode when `process.env.NODE_ENV === 'development'`.
-More information about usage of Neutrino middleware can be found in the [documentation](https://neutrinojs.org/middleware/).
 
 ## Middleware options
 
--- a/packages/web/index.js
+++ b/packages/web/index.js
@@ -7,7 +7,6 @@ const htmlTemplate = require('@neutrinojs/html-template');
 const clean = require('@neutrinojs/clean');
 const loaderMerge = require('@neutrinojs/loader-merge');
 const devServer = require('@neutrinojs/dev-server');
-const { resolve } = require('url');
 const merge = require('deepmerge');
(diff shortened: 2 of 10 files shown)


packages/dev-server/README.md
```diff
@@ -112,9 +89,6 @@ By default this middleware will start a development server with Hot Module Repla
 `http://localhost:5000`. To enable HMR with your application, read the documentation of corresponding Neutrino
 preset or middleware.
 
-It is recommended to call this middleware only in development mode when `process.env.NODE_ENV === 'development'`.
```
edmorley: Now that the `@neutrinojs/dev-server` preset only sets the `devServer` option, there's no real real why it shouldn't always be used.

edmorley:
>     * Changes `output.publicPath` from `'./'` to `''` since the latter is equivalent (and is also the webpack default), however it correctly allows webpack-dev-server to fall back to `'/'` when needed

Note we may be changing this again in #1171, however I wanted to get the ~"no-op" cleanup out of the way first.

packages/web/index.js
```diff
@@ -93,18 +98,6 @@ module.exports = (neutrino, opts = {}) => {
...
-      '**': {
-        target: options.devServer.proxy,
-        changeOrigin: true,
-        headers: {
-          Forwarded: 'by=_webpack-dev-server'
-        }
-      }
-    };
-  }
-
```
timkelty: Should we include the `headers` bit in our example? If not here, perhaps a page in the docs for "Using Neutrino with Existing SSR Architecture" would make sense.

edmorley: I think we could easily write several pages on how to set up SSR. I'm pretty sure html-webpack-plugin doesn't fully work with out of the box and needs some experimental options enabled.

I think this example is perhaps best aimed at the more common case of "I have a standard SPA but it uses a backend and I want to proxy to it in development". (Not that it precludes adding a more advanced SSR guide somewhere else in the future)

edmorley: Also, perhaps an SSR guide is best suited as an extra page in webpack's docs? Ultimately Neutrino is not meant to be a replacement for all of webpack/...'s documentation, but a very thin layer of abstraction that assists with configuration generation, and that makes it easy to mentally map between a webpack guide and how to add that configuration via `.neutrinorc.js` or a custom Neutrino preset.

timkelty: #1177 :)

edmorley: > I think we could easily write several pages on how to set up SSR

I underestimated...!
https://ssr.vuejs.org/

timkelty: Also: it's really 2 different things if you're talking about SSR with React/Vue vs. using webpack/neutrino for "assets", and something else for SSR of html (PHP, Ruby, etc)

edmorley: Yeah true

constgen:
@edmorley I believe the story of these settings is here https://github.com/neutrinojs/neutrino/issues/187

PR #1182, opened 2018-10-19
eslint: Refactor eslint-loader and eslintrc config generation

### LINKED ISSUE #382: Reducing the potential for confusion if configuring ESLint outside of Neutrino

### LINKED ISSUE #1181: Issues with @neutrinojs/eslint
Whilst looking into #382 I've found a few issues with the current `@neutrinojs/eslint` implementation, which I'll fix as part of a refactor. I'm documenting these issues here so as to make the reasoning behind the changes in the upcoming PR clearer.

First some background:
* Our linting approach changed significantly for Neutrino 9 due to #852 - in that for `yarn lint` we no longer use ESLint's API (via [CLIEngine.executeOnFiles()](https://eslint.org/docs/developer-guide/nodejs-api#executeonfiles)), and instead rely on `.eslintrc.js`. This is good news since it means we've gone from 3 ways linting may occur (see breakdown in #382) to 2 ways (`.eslintrc.js` and [eslint-loader](https://github.com/MoOx/eslint-loader)).
* However there is still a difference in configuration for both, since our generated `.eslintrc.js` must adhere to the schema [here](https://github.com/eslint/eslint/blob/v5.7.0/conf/config-schema.js), whereas `eslint-loader` uses [CLIEngine.executeOnText()](https://eslint.org/docs/developer-guide/nodejs-api#executeontext) which has subtly different options.
* This means that `@neutrinojs/eslint` has to convert between the two formats when using the `eslintrc()` output handler
* eslint-loader itself is quite frustrating to configure correctly, since:
  - invalid options are silently ignored, since neither the loader nor CLIEngine performs validation (I've filed webpack-contrib/eslint-loader#252 and eslint/eslint#10272) - making it hard to debug why a config isn't working as expected when the wrong names were used
  - some options have names that are very subtly different between their top-level and `baseConfig` versions (for example `envs` vs `env`)
  - some options have completely different data types when top-level options vs when under `baseConfig` (for example `globals` is an array of strings, whereas `baseConfig.globals` is an object whose keys are the global names, and the value is `true` or `false`)
  - some options are available as both top-level and `baseConfig` variants, whereas for others (such as `settings` and `extends`) the only way to set them is under `baseConfig`
  - the [eslint-loader docs](https://github.com/webpack-contrib/eslint-loader#options) only really discuss in detail the options that are used by eslint-loader itself, rather than give examples/list any of the gotchas above (I'll open a PR at some point)
  - the [ESLint CLIEngine docs](https://eslint.org/docs/developer-guide/nodejs-api#cliengine) don't explain `baseConfig` in detail, meaning a user has to work out on their own that it relates to the schema in `conf/config-schema.js` (I'll open a PR at some point)

On to the issues with `@neutrinojs/eslint`:

1) The [eslintrc mapping process](https://github.com/neutrinojs/neutrino/blob/14b47045c5af9f691476b3a1560ab0c6e3fe41ac/packages/eslint/index.js#L24-L77) ignores all loader options nested under `baseConfig`, apart from `settings` and `extends` (other possible options are: `env`, `globals`, `overrides`, `parser`, `parserOptions`, `plugins`, `root` and `rules`). Notably, `overrides` cannot be set anywhere other than inside `baseConfig` (it's not supported as a top-level option), so this bug prevents people from using `overrides` with eslint-loader.

2) The eslintrc mapping process merges in several loader options that are actually not valid (for example top-level `extends`, `settings`, `overrides`, `root` options), which encourages people to use broken loader configurations that give different linting results depending on whether used via `.eslintrc.js` or `eslint-loader`. This is made worse by (1), since the valid location for many of these options is actually ignored, meaning even if someone knows the correct way to write the configuration, the options will be ignored by the eslintrc mapping.

3) The eslintrc mapping process has an [incomplete list](https://github.com/neutrinojs/neutrino/blob/14b47045c5af9f691476b3a1560ab0c6e3fe41ac/packages/eslint/index.js#L34-L43) of which keys must be filtered out, meaning if options like `ignore` or `reportUnusedDisableDirectives` are used, then the `.eslintrc.js` file will fail to load with a validation error. Any key not specified in [the schema](https://github.com/eslint/eslint/blob/v5.7.0/conf/config-schema.js) must be removed, however there are many [eslint-loader options](https://github.com/webpack-contrib/eslint-loader#options) (and more that are undocumented) as well as [CLIEngine options](https://eslint.org/docs/developer-guide/nodejs-api#cliengine). As such a whitelist rather than blacklist approach would probably be simpler long-term.

4) The custom merge implementation [here](https://github.com/neutrinojs/neutrino/blob/14b47045c5af9f691476b3a1560ab0c6e3fe41ac/packages/eslint/index.js#L6-L23) only applies the special rule-merging logic to the top-level `rules` option, and not also `baseConfig.rules`. It also doesn't merge rules the same way that ESLint does internally for [certain](https://github.com/eslint/eslint/blob/5984820670c3085becc98b75827fc4899001e605/tests/lib/config/config-ops.js#L305-L319) [cases](https://github.com/eslint/eslint/blob/5984820670c3085becc98b75827fc4899001e605/tests/lib/config/config-ops.js#L481-L494). It would be best to use ESLint's own merge function instead to ensure accuracy/consistency and also reduce our maintenance burden.

5) It doesn't validate the options passed to the preset (which would be less of an issue if eslint-loader or CLIEngine did so themselves).

6) It uses the top-level `globals` and `envs` options, which whilst slightly more convenient (since they are arrays rather than the object form under `baseConfig`), requires users to know how they relate to the more commonly seen `.eslintrc.js` form when trying to understand the output of `--inspect`. This has been known to cause confusion when debugging (https://github.com/neutrinojs/neutrino/issues/1166#issuecomment-429067147), makes it harder to just copy and paste the config into a static `.eslintrc.js`, and means one can't simply merge in eg `env: { foo: false }` to disable an option set by an existing preset (instead requiring manual array filtering).

7) There is no easy way to suppress the default ESLint-config-specific options in `@neutrinojs/eslint` when a project wants to instead use their own non-generated `.eslintrc.js` (see #382).

I have a branch locally that fixes these - will open a PR once I've finished updating the docs/tests.

edmorley:
* Adds option validation to `@neutrinojs/eslint`, so that it now throws when passed options that are not supported by eslint-loader.
* The eslintrc conversion now correctly handles all supported options under `baseConfig` rather than silently ignoring them.
* The eslintrc conversion no longer looks for settings in locations that are invalid for eslint-loader (such as top-level `settings`, `extends`, `root` or `overrides` options).
* The eslintrc conversion now no longer errors if passed options that were previously missing from the `omit()` list, such as `ignore` and `reportUnusedDisableDirectives`.
* The custom lint `merge()` function has been replaced with ESLint's own merging function, which correctly handles several cases that were broken before.
* All non-loader related configuration has now been moved under the `baseConfig` key, since the options under it more closely relate to the options that users will have previously seen in `.eslintrc` files, thereby reducing user-confusion and making it possible to copy and paste the contents of the `baseConfig` section to a static `.eslintrc.*`.
* Support has been added for projects that want to use their own non-generated `.eslintrc.*` file. They now only need to set `useEslintrc` to `true`, which will make the linting presets only set the loader-related options, preventing conflicts with the options set in their static `.eslintrc.*` file.

Fixes #1181.
Fixes #382.


packages/airbnb-base/README.md
```diff
@@ -168,64 +168,26 @@ module.exports = {
...
-
-The following is a list of rules and their identifiers which can be overridden:
-
-| Name | Description | NODE_ENV |
-| --- | --- | --- |
-| `lint` | Lints JS and JSX files from the `src` directory using ESLint. Contains a single loader named `eslint`. This is inherited from `@neutrinojs/eslint`. | all |
-
-### Information
-
-If you want your preset or middleware to also extend from another **ESLint configuration or preset** that you have made
```
edmorley: This is now covered by all of the defaults above being listed under `baseConfig` instead

packages/airbnb-base/README.md
```diff
@@ -168,64 +168,26 @@ module.exports = {
...
-a dependency, you must use `baseConfig.extends` rather than just `extends`. This is a limitation of ESLint, not this
-middleware.
-
-### Override configuration
```
edmorley: This section is unnecessary IMO (given most people should not be using this method of customisation), and given the pain points we see around lint configuration, I think we need to take a strict "less is more" approach to what we put in the docs for it.

packages/airbnb-base/README.md
```diff
@@ -168,64 +168,26 @@ module.exports = {
   use: [
     ['@neutrinojs/airbnb-base', {
       eslint: {
-        rules: {
```
edmorley: It's worth noting that this form still works, it's just not what we're using for the defaults / what we're recommending. Any preset that pass in the old form will still correctly override the values under `baseConfig`, since ESLint's CLIEngine (plus our eslintrc conversion function) merges them in such a way that the top-level options take precedence.

packages/airbnb-base/test/airbnb_test.js
```diff
@@ -56,3 +56,169 @@ test('exposes eslintrc method', t => {
...
+              'Immutable.Set',
+              'Immutable.List'
+            ],
+            newIsCap: true,
+            newIsCapExceptions: []
+          }
+        ],
+        'babel/no-invalid-this': 'off',
+        'babel/no-unused-expressions': [
+          'warn',
```
edmorley: This case is an example of one of the tests that would have failed previously - since before the resultant config would have discarded the rule settings, and just had the value of `'warn'`.

packages/eslint/README.md
```diff
@@ -42,68 +42,53 @@ const eslint = require('@neutrinojs/eslint');
 
 // Usage shows default values
 neutrino.use(eslint, {
-  test: neutrino.regexFromExtensions(), // Uses extensions from neutrino.options.extensions
+  // Uses extensions from neutrino.options.extensions
```
edmorley: I've adjusted the wrapping here so the code block in the docs doesn't need to have horizontal scroll bars.

packages/eslint/README.md
```diff
@@ -42,68 +42,53 @@ const eslint = require('@neutrinojs/eslint');
...
-    settings: {},
-    globals: ['process'],
-    rules: {}
+    // The options under `baseConfig` correspond to those
+    // that can be used in an `.eslintrc.*` file.
+    baseConfig: {
+      env: {
+        es6: true
+      },
+      extends: [],
```
edmorley: I've deliberately included empty config options here to give hints (both in the docs and via `--inspect`) as to where the correct place is to pass each of the options.

packages/eslint/index.js
```diff
@@ -1,82 +1,81 @@
-const deepmerge = require('deepmerge');
-const clone = require('lodash.clonedeep');
-const omit = require('lodash.omit');
+const pick = require('lodash.pick');
+const { merge: eslintMerge } = require('eslint/lib/config/config-ops');
 const { ConfigurationError, DuplicateRuleError } = require('neutrino/errors');
 
-const merge = (source, destination) => {
```
edmorley: This old merge function missed several cases:
* it didn't apply the rule-merge to `baseConfig.rules`
* the rule-merge doesn't keep the settings of the old rule, when wanting to just adjust the off/warn/error level (unlike eslint's own merge)
* it doesn't merge `overrides` correctly

See these testcases:
https://github.com/eslint/eslint/blob/5984820670c3085becc98b75827fc4899001e605/tests/lib/config/config-ops.js#L305-L319
https://github.com/eslint/eslint/blob/5984820670c3085becc98b75827fc4899001e605/tests/lib/config/config-ops.js#L481-L494

packages/jest/src/index.js
```diff
@@ -1,23 +1,32 @@
...
+  const lintRule = neutrino.config.module.rules.get('lint');
+  if (lintRule) {
+    lintRule.use('eslint').tap(
+      // Don't adjust the lint configuration for projects using their own .eslintrc.
+      lintOptions => lintOptions.useEslintrc
+        ? lintOptions
+        : merge(lintOptions, {
+            baseConfig: {
+              env: {
+                'jest/globals': true
```
edmorley: I've converted the top-level `envs` array form to the object form under `baseConfig.env`, since:
* users will be more familiar with it (ref https://github.com/neutrinojs/neutrino/issues/1166#issuecomment-429067147)
* it then means it's possible to override an `env` by passing eg `'jest/globals': false` (with arrays manual filtering would have to be performed)
* it means the resultant `--inspect` output is much closer to a form that can be copy-pasted into an `.eslintrc.*` file

PR #1215, opened 2018-11-19
Use webpack's default value for `minimizer`

edmorley:
Since as of webpack 4.26.0 `terser-webpack-plugin` is the new default, so we no longer need to set it manually:
https://github.com/webpack/webpack/releases/tag/v4.26.0

This is a partial revert of #1158.


eliperelman (approved):
:100:

PR #1217, opened 2018-11-19
dev-server: Enable in-page overlay for display of errors

### LINKED ISSUE #1131: Decide whether we should enable devServer.overlay
This seems like it might be a DX win (eg CRA has an overlay, albeit a custom one):
https://webpack.js.org/configuration/dev-server/#devserver-overlay

edmorley:
See:
https://webpack.js.org/configuration/dev-server/#devserver-overlay

Note this change 'relies' on #1216, in that without that change all linter errors would be displayed in the overlay too, which is likely too noisy (given that many rules are for stylistic issues that should not block viewing the dev-server page output).

Fixes #1131.


edmorley:
"breaking" change only in the sense that it causes people's workflow to potentially change slightly.

PR #1219, opened 2018-11-23
mocha: Use correct function for merging lint configuration

edmorley:
The refactor in #1182 inadvertently used babel-merge's `merge()` to merge ESLint configuration in this preset rather than deepmerge, since the existing `merge` import was not the same as in other presets from which the lint rule merging block was copied.

In a later PR I think we should stop exporting a merge function from `@neutrinojs/compile-loader` named `merge`, to avoid the reduce the chance of this kind of naming conflict.

Fixes:
https://github.com/taskcluster/taskcluster-web-server/pull/59#issuecomment-440801799

### DIFF
--- a/packages/mocha/package.json
+++ b/packages/mocha/package.json
@@ -24,6 +24,7 @@
     "@babel/plugin-transform-modules-commonjs": "^7.1.0",
     "@babel/register": "^7.0.0",
     "babel-merge": "^2.0.1",
+    "deepmerge": "^1.5.2",
     "lodash.omit": "^4.5.0"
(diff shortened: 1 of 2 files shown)


helfi92:
@edmorley Any idea when this will be published?

edmorley:
Publishing new releases of the monorepo is pretty broken on Windows at the moment, so I've been leaving it to Eli to publish new versions. Happy to have a new beta published whenever :-)

helfi92:
A new beta sounds good to me :)

Issue #1197, opened 2018-11-02
Consider adding a .gitignore during create-project


edmorley:
This would save users having to manually ignore `node_modules/` and `build/` themselves, and is what CRA does.

Whilst it does mean us making an assumption about their choice of VCS:
1. git is pretty common
2. if they don't use Git, they can at least convert the `.gitignore` contents to the VCS format they use (and it will remind them to do so), or just delete the file

Note: We'll need to be careful if we decide to include the file in the neutrino package, since `npm pack` renames any found `.gitignore` (so we'll either need to ship as a different name inside create-project and rename during creation, or else create the file from scratch during create-project).

rajatsharma:
Can I pick this?

rajatsharma:
Please review my PR #1237

edmorley:
Fixed in #1340.

PR #1237, opened 2018-12-04
Generating gitignore at project creation

rajatsharma:


eliperelman:
I'm skeptical of forcing this to output a gitignore file, especially since there are other version control systems that Neutrino users are using. What if we instead recommended them running a command to output most of the relevant gitignore content, and adding Neutrino specifics like the `build` directory?

```bash
# From https://www.npmjs.com/package/gitignore
npx gitignore node
```

edmorley:
I'm happy with that or wontfixing #1197 too.

eliperelman:
OK, I think we can generate this for them since git is much more prevalent and a conversion to hgignore should be as easy as a file rename. I'll add a relevant review.

eliperelman:
Resolved in #1340.

PR #1246, opened 2018-12-08
Only allow functions as middleware

eliperelman:
Addresses the comment in https://github.com/neutrinojs/neutrino/issues/1129#issuecomment-444504351

cc: @yordis

- Switches API to only support functions as middleware
- No more package loading, packages must be manually required
- Add new migrate tool which codemods a `.neutrinorc.js` array and string formats to the function call format
- Tests updated
- Documentation updated
- Migration guide updated


edmorley:
Thank you for looking into this! Some initial thoughts (I'll have time to review this in more detail in the next day or so):
* having a migrate tool is a great idea, though it's going to need many many test cases to ensure we handle all possible permutations of middleware usage
* even though we have the migrate tool, should we make this non-breaking for v9? ie: allow people to optionally migrate, but still support the old string based middleware (perhaps outputting a deprecation message)? Support for it could then be removed in v10. It's just we have a lot of breaking changes already for v9, so many so that it might put people off upgrading.
* perhaps we could make the migrate tool support handling some of the other v9 breaking changes too? (in a separate PR after this)

yordis:
@edmorley the cost of maintaining the old string is higher than coming up with a tool that would help them to be honest.

Breaking changes are bad when there is not `clear` documentation of what they need to do about it and you leave people alone figuring out so many things that they didn't understand in the first place.

I would prefer to focus on a good migration guideline than keep maintaining old code that does not add any value.

That is what I would do, especially that from end users is just a replacement for a module instead of a string.

And middleware authors is either move to a closure or simply remove the `options` out of the function.

edmorley:
> the cost of maintaining the old string is higher than coming up with a tool that would help them to be honest

I wasn't meaning to not write the migration tool; I was saying we should do both (keep backwards compat for one version and also have a migration tool).



> Breaking changes are bad

Breaking changes are often also classed as "bad" when there is a high percentage of people affected (in this case 100%, which is a lot higher than for many of our other v9 breaking changes), and where there is no deprecation process (if we don't have a 1 version overlap).

I agree support for string middleware shouldn't be kept indefinitely, but could you explain more about the downsides to a 1 version overlap of the two?

timkelty (commented):
This is such a good move!
This will make my job of introducing neutrino to my colleagues much easier.

I skimmed through the PR and it looks good on the surface, but didn't dig  that deep.

edmorley:
> having a migrate tool is a great idea, though it's going to need many many test cases to ensure we handle all possible permutations of middleware usage

For me this is the main reason we shouldn't make this breaking for v9.

We're so so close to releasing v9 (after many many hours on my part trying to get the loose ends/polish of the release resolved), that I'm very keen to ensure we don't need multiple rounds of alphas/betas again before we can release it.

timkelty:
> For me this is the main reason we shouldn't make this breaking for v9.

@edmorley what about leaving the formats in-place but updating the docs and putting in a deprecation notice for v10 using any non-fn middlware?

edmorley:
Yeah that's what I'd be happy with (ie this can land, so long as it's not breaking).

yordis:
@edmorley well, funny enough I already changed my middlewares to be all `fn` middlewares. Since the closure could be done today. I just need to read the `opts` from the closure instead of expecting any second parameter for it.

So even in the current version is working.

Just let's try to move away from non-fn middlewares ❤️

edmorley (changes requested):
Let's make this non-breaking for now

eliperelman:
I was originally on board with leaving this for Neutrino 9, but I think this puts a greater burden on middleware maintainers. Instead of having to release a breaking change only for 9, they will most likely have to release a breaking change for 10 as well.

I'd like to hear a stronger argument on if this tradeoff is worth needing to have these users do breaking changes twice.

edmorley:
I don't quite follow - could you elaborate? (Specifically, why breaking twice for them)

For me the biggest reason to leave this for now is that it's not ready, and it's already 6+ months after we'd said we'd release Neutrino 9, but still haven't 😢

timkelty:
> For me the biggest reason to leave this for now is that it's not ready, and it's already 6+ months after we'd said we'd release Neutrino 9, but still haven't 😢

I think this is the best argument against, but at the same time with v9's delay, I'm feeling a bit like "what harm is a bit more?" 😬

 > I'd like to hear a stronger argument on if this tradeoff is worth needing to have these users do breaking changes twice.

Considering v9 is already a pretty monumental release and everyone seems to think this is good idea in general, my inclination is make this breaking change now, in v9.

My hope is v9's architecture changes and exposure of native tooling will appeal to a wider audience that may have dismissed Neutrino before as an "uncessary abstraction".

If the community does indeed grows with this release, it could be an argument to make this kind of breaking change now.

That all said, I know @edmorley has done a lot of work to get v9 out the door, so I'm really ok with whatever ya'll decide :)

eliperelman:
I agree with both of you. Most of the work is done in this patch, but I also **do not want to do anything that would postpone the release more than a couple days**. I will rebase, and we can discuss if that is feasible, and if not, we can move forward without this. This item is the lowest priority, but I would like to see at least some effort into whether we think we can pull this off for v9.

timkelty:
@eliperelman I'll be around to test & review.

edmorley:
Sounds good to me :-)

eliperelman:
I tried writing shim functionality but ran into issues in being able to support both neutrinorc files and middleware authors working in the 2 different formats. We either need to pull the trigger on this or wait until Neutrino 10.

Something else to consider on whether to make this land in Neutrino 9 is upcoming v10 features. So far nothing is on the radar for breakages, so the release is essentially only for semantic improvement.

So do we release a v10 immediately and cause release churn, or artificially delay releasing this?

timkelty:
My vote is pull the trigger, but I think there are decent arguments for either…

eliperelman:
I can also investigate adding a few more migration automations to the migration tool.

yordis:
I think that the effort to be compatible with this change is not that big, for the most part, instead of using strings you will need to import the package.

With a good migration guide and preparation for incoming issues, we should be fine.

Like I said in the issue, I am more than happy to take the lead on supporting this movement and explaining to people how to fix it, so feel to bug me on helping middleware authors and other people (just tag me on it)

timkelty (approved):
❤️ this. Tried the migration tool on a couple existing `.neutrinorc.js` files and it worked flawlessly.

edmorley:
Thinking about this more, I'm fine with landing as a breaking change in v10.
I won't have a chance to do the review until tomorrow, but will definitely take a look then :-)

eliperelman:
OK cool, let's target v10 then, with the caveat that I want a shorter turnaround from v9 to v10. If this is the only breaking change then I am cool with that, and will make the migration pretty easy from there.

edmorley:
Sorry I meant to say v9, not v10.

edmorley:
(😆 )

eliperelman:
Ah good to know! I'll revise then.

timkelty:
> Sorry I meant to say v9, not v10.

@edmorley I _thought_ you phrased that oddly. 😆 

Looking forward to your review!

packages/mocha/README.md
```diff
@@ -68,10 +68,13 @@ Now update your project's `.neutrinorc.js` to add the `@neutrinojs/mocha` preset
...
+
 module.exports = {
   use: [
-    '@neutrinojs/node',
-    '@neutrinojs/mocha'
+    node(),
+    mocha()
   ]
 };
 ```
```
edmorley: The "Preset options" section just below here needs updating to migrate the `['@neutrinojs/mocha', { bail: true }]` and also the "Use an array pair instead of a string" sentence.

packages/clean/README.md
```diff
@@ -37,36 +37,36 @@ as well as all unused webpack assets after every successful rebuild.
...
-    ['@neutrinojs/clean', {
+    clean({
       verbose: neutrino.options.debug,
       // Override pluginId to add an additional banner plugin instance
       pluginId: 'clean'
-    }]
+    })
   ]
 };
 ```
```
edmorley: The "Clean external directories" section just below here needs updating (presume this raced with the PR that added it).

packages/migrate/bin/migrate.js
```diff
@@ -0,0 +1,47 @@
...
+    description: 'do not write to stdout or stderr',
+    boolean: true
+  })
+  .help()
+  .wrap(null)
+  .argv;
+
+run(
+  resolve(__dirname, '../transforms/middleware.js'),
+  cli.files.map(file => join(process.cwd(), file)),
```
edmorley: If I pass an invalid filename, I get:

```
migrate invalid
Skipping path C:\Users\Ed\src\treeherder\invalid which does not exist.
TypeError: array is not iterable
    at concatAll (C:\Users\Ed\src\neutrino\node_modules\jscodeshift\src\Runner.js:72:27)
    at processTicksAndRejections (internal/process/next_tick.js:81:5)
```

Should we try and avoid the `TypeError` here, or is it not worth it?

docs/migration-guide.md
```diff
@@ -15,6 +15,36 @@ native tools and utilities for which they were originally created. The biggest b
...
 and others' native CLIs will be used in tandem with Neutrino.
 
+Additionally, the Neutrino API no longer accepts any middleware format other than functions.
+The `.neutrinorc.js` file is an exception, which still supports an object-based definition
+for setting Neutrino options. There is now a new package to assist in migration changes to core
+Neutrino middleware which are now middleware factories, since packages can no longer be required
+for you:
+
+```bash
+npx @neutrinojs/migrate rc
```
edmorley: If I use this command as stated I get a "Skipping path C:\Users\Ed\src\treeherder\rc which does not exist.".
I'm presuming this should now be `npx @neutrinojs/migrate` with no additional arguments?

packages/migrate/transforms/middleware.js
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Transform .neutrinorc.js files to remove legacy middleware string-based
+ * formats in module.exports.use and neutrino.use(). Will move string formats
+ * as require()s to top of file.
+ * use: ['@neutrinojs/react'] -> use: [react()]
+ * use: [['@neutrinojs/react', { options }]] -> use: [react({ options })]
+ * neutrino.use('@neutrinojs/react') -> neutrino.use(react())
+ * neutrino.use('@neutrinojs/react', options) -> neutrino.use(react(options))
+ */
+module.exports = ({ source }, { jscodeshift: j }) => {
```
edmorley: When I tried migrate against Treeherder's `.neutrinorc.js`, it left these middleware usages unchanged:
https://github.com/mozilla/treeherder/blob/4c01ee963b1f227101885864f881ccf3918561eb/.neutrinorc.js#L54-L56
https://github.com/mozilla/treeherder/blob/4c01ee963b1f227101885864f881ccf3918561eb/.neutrinorc.js#L130-L132

This style of usage is likely not that common, but thought I'd mention in case it would be worth seeing if it was easy to support without false positives. (They would probably also want to be written using inline `require()` to keep startup time low if the conditional is not true.)

packages/neutrino/Neutrino.js
```diff
@@ -205,39 +138,15 @@ module.exports = class Neutrino {
...
 
-      if (middleware.use) {
-        if (Array.isArray(middleware.use)) {
-          middleware.use.map(usage => this.use(usage));
-        } else {
-          this.use(middleware.use);
-        }
-      }
+    if (typeof middleware !== 'function') {
+      throw new ConfigurationError('Middleware must be a function');
```
edmorley: If I try running `yarn test` without first changing Treeherder's `.neutrinorc.js`, I get:

```
treeherder $ yarn test
yarn run v1.13.0
$ node ./node_modules/jest/bin/jest

An error occurred when loading the Neutrino configuration.

ConfigurationError: Middleware must be a function
    at Neutrino.use (C:\Users\Ed\src\neutrino\packages\neutrino\Neutrino.js:147:13)
    at use.forEach.use (C:\Users\Ed\src\neutrino\packages\neutrino\index.js:42:37)
    at Array.forEach (<anonymous>)
    at module.exports (C:\Users\Ed\src\neutrino\packages\neutrino\index.js:42:13)
    at Object.<anonymous> (C:\Users\Ed\src\treeherder\jest.config.js:5:18)
    at Module._compile (internal/modules/cjs/loader.js:734:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:745:10)
    at Module.load (internal/modules/cjs/loader.js:626:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:566:12)
    at Function.Module._load (internal/modules/cjs/loader.js:558:3)
```

This is better than a more obscure TypeError or similar, however I feel that since this is likely the first error people see upon updating to Neutrino 9 (eg in CI when a bot opens an update PR), it might be worth making the error message slightly longer/clearer (eg "As of Neutrino 9, middleware must now be passed as a function and not an array or string"), and linking to the migration guide/migrate tool? 

Also, would it be possible to have the error mention where the invalid usage occurred? eg to differentiate top-level `.neutrinorc.js` vs one of their third-party presets not having been updated to support Neutrino 9?

Relatedly, if I put just `react` and not `react()` inside the `.neutrinorc.js` `use` array, at the moment I don't get a Neutrino error message and instead an obscure webpack message. I feel that an easy/common mistake for people to make, will be confusing the "middleware that takes options" and "middleware that does not" forms. 

As such, perhaps Neutrino should help catch this case? eg by either (or both of):
* verifying the passed middleware takes one and only one argument (using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length)
* checking the passed middleware doesn't return a function of it's own

packages/migrate/README.md
```diff
@@ -0,0 +1,147 @@
+# Migrate a Neutrino Project
+
+`@neutrinojs/migrate` can assist with some of the changes needed to migrate to
+Neutrino 9.
+
+## Getting Started
+
+Run the following command to start the process. Substitute `<...tasks>` with
```
edmorley: I think `<...tasks>` here doesn't yet exist?

edmorley (changes requested):
This is great! The number of changes required for end users is actually pretty small, the new syntax much cleaner (and incidentally I think more Prettier-friendly than the array syntax), and the Neutrino API less complex.

I spotted a few things, but nothing that should block including this in Neutrino 9 :-)

packages/neutrino/index.js
```diff
@@ -1,14 +1,19 @@
...
 const Neutrino = require('./Neutrino');
 const { webpack, inspect } = require('./handlers');
 
-module.exports = (middleware = {}) => {
-  const use = isPlainObject(middleware) && !middleware.use
-    ? { ...middleware, use: ['.neutrinorc.js'] }
-    : middleware;
+const extractMiddlewareAndOptions = (format) =>
+  typeof format === 'function'
+    ? { ...format, use: format }
```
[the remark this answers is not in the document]
eliperelman: Functions are objects and can have their properties spread. That means you could technically have a `.neutrinorc.js` do this:

```js
module.exports = Object.assign((neutrino) => {

}, {
  options: {
    root: __dirname
  }
});
```

eliperelman:
@edmorley OK, I think I resolved all your concerns to the best of my judgement in the latest commit. Please let me know if you have any more questions or concerns. 😃

packages/neutrino/Neutrino.js
```diff
@@ -52,6 +32,18 @@ module.exports = class Neutrino {
       ...clone(opts)
     };
 
+    if ('node_modules' in options) {
+      throw new ConfigurationError(
+        'options.node_modules has been removed. Use `neutrino.config.resolve.modules` instead.'
+      );
+    }
+
+    if ('env' in options) {
```
edmorley: My bad for using the phrase `options.env` in the review comment previously - `env` was actually a top-level property and not under `options`. As such, this check doesn't catch existing usages any more. 

ie: Trying locally with `env` in `.neutrinorc.js` I don't get the `ConfigurationError`.

packages/neutrino/Neutrino.js
```diff
@@ -205,39 +150,55 @@ module.exports = class Neutrino {
...
+      throw new ConfigurationError(
+        `"${middleware[0]}" is specified as an array, but as of Neutrino 9,\n` +
+          'middleware can only be passed as functions.\n' +
+          'Use the migration tool and see the migration guide for details:\n' +
+          '  https://neutrinojs.org/migrate\n' +
+          '  https://neutrinojs.org/migration-guide'
+      );
+    } else {
+      throw new ConfigurationError(
+        'As of Neutrino 9, middleware can only be passed as functions.\n' +
```
edmorley: I think the only time this case will be hit is when someone passes something completely invalid (eg an integer, an object or `true`). As such, perhaps the error message does not need to mention Neutrino 9 or the migration guide (since this would have been invalid under Neutrino 8 or earlier too) but instead a more generic "invalid middleware passed, must be a function ..." type message?

packages/neutrino/test/api_test.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-vars */
```
edmorley: I think we should disable this rule per-line rather than globally for the file, otherwise genuinely unused variables won't be reported.

edmorley (approved):
This is great! Thank you for those changes - the migrate tool now handles the Treeherder `.neutrinorc.js` well.

(I've left a few more comments, but happy for them to be fixed without needing further review)

packages/neutrino/Neutrino.js
eliperelman: My concern here was the object case, as I did find a couple of those in the tests, and it was technically valid middleware. I will still remove the migration bits.

eliperelman:
Here we go! 🌮

edmorley:
I think this part should be removed, now that the `env` handling was fixed?
https://github.com/neutrinojs/neutrino/commit/faaa112755f91d1012c182dd770bf3997ba0f0f5#diff-f29c4044a287268f0e44b5115647d02eR41

eliperelman:
@edmorley good catch.

eliperelman:
I pushed this change directly to master. 🤷‍♂️

edmorley:
Thank you :-)

PR #1252, opened 2018-12-12
Only default to chunks to 'all' when using generated html.

timkelty:
This seems like a better default, as if you're _not_ using html-webpack-plugin/auto-generated html (`html: false`), you're likely relying on a manifest.json and need predictable chunk names.

And/or to make this easier to configure, we could expose the splitChunks config object as an option for  @neutrinojs/web.

Thoughts?

### DIFF
--- a/packages/web/index.js
+++ b/packages/web/index.js
@@ -180,7 +180,7 @@ module.exports = (neutrino, opts = {}) => {
         // ever-changing file list breaking users who don't auto-generate their HTML):
         // https://webpack.js.org/plugins/split-chunks-plugin/#optimization-splitchunks-chunks-all
         // https://github.com/webpack/webpack/issues/7064
-        chunks: 'all',
+        chunks: options.html ? 'all' : 'async',
         // Override the default limit of 3 chunks per entrypoint in production, since
         // it speeds up builds and the greater number of requests is mitigated by use
         // of long term caching (and a non-issue if using HTTP2).


edmorley:
> you're likely relying on a manifest.json and need predictable chunk names.

The manifest would ideally expose the mapping of entry-points to chunks - however doesn't at the moment, albeit there are some workarounds in:
danethurber/webpack-manifest-plugin#133

Would you mind chasing that issue up?

I'm not immediately against this change (since it doesn't harm the common-case of using html-webpack-plugin), however even with this change users will still need to manually hardcode the runtime filename, which could be avoided if webpack-manifest-plugin was fixed.

timkelty:
@edmorley it appears that https://github.com/webdeveric/webpack-assets-manifest supports entryPoint mapping, though I have yet to test.

So, we could just switch to that plugin, OR if we want to slim things up, we could just drop the auto-inclusion of any manifest plugin and leave it up to the user.

I'm also playing around with leveraging `html-webpack-plugin` [like this](https://github.com/jantimon/html-webpack-plugin/issues/1123) instead of messing with a manifest to get around issues like this.

timkelty:
Hmmm, so I played around with `webpack-assets-manifest` and it seems to do what we're after:

by passing `entrypoints: true`, you'll get a manifest like so:

```
{
  "entrypoints": {
    "index": {
      "js": [
        "assets/runtime.1e93d1b9.js",
        "assets/2.68770d46.js",
        "assets/index.1d658dfa.js"
      ],
      "css": [
        "assets/index.f513dc62.css"
      ]
    }
  },
  "index.css": "assets/index.f513dc62.css",
  "index.js": "assets/index.1d658dfa.js",
  "runtime.js": "assets/runtime.1e93d1b9.js"
}
```

@edmorley Since the upgrade to `webpack-manifest-plugin` is [already breaking](https://github.com/neutrinojs/neutrino/issues/1250) for v9, do you think we should switch over? Too late? Or should we just drop a manifest plugin altogether? The nice part about having it in `@neutrinojs/web` is knows to only do it when `isProduction`, but that's about it.

timkelty:
Yep, makes sense.

Issue #1250, opened 2018-12-11
v8 → v9 paths in manifest include publicPath


timkelty:
I'm thinking this probably wasn't intentional…

in v8, manifest.json paths would look like this:

```
  "index.css": "index.ad53685581180eded46aa8c728c445a2.css",
  "index.js": "index.6cb2d5ac4d3e6547bdef.js",
```

in v9, they seem to include the `publicPath`: 

```
  "index.css": "/build/assets/index.89918c52.css",
  "index.js": "/build/assets/index.38ac4f50.js",
```

with a config like:

```
module.exports = {
  options: {
    root: __dirname,
    output: 'public/build'
  },
  use: [
    neutrino => neutrino.use('@neutrinojs/web', {
      publicPath: `/build/`,
      html: false,
    });
  ]
}
```

timkelty:
Looks like this was a change in manifest v1 to v2: https://github.com/neutrinojs/neutrino/pull/784

timkelty:
Indeed, it now has a default (`output.publicPath`): https://github.com/danethurber/webpack-manifest-plugin#optionspublicpath

edmorley:
So do you think this is working as expected (and so wontfix)?

timkelty:
@edmorley yep I think so, I’ll add a not to the migration guide.

timkelty:
Eh theres already a note about "Various dependencies have been updated which may bring their own breaking changes"…probably good enough.

If anyone finds this thread later and wants to revert to the old behavior, you would do so by passing options like this to `@neutrinojs/web` (or vue/react etc):

```
    neutrino => neutrino.use('@neutrinojs/web', {
      html: false,
      manifest: {
        publicPath: ''
      },
    })
```

PR #1253, opened 2018-12-12
Replace webpack-manifest-plugin with webpack-assets-manifest

timkelty:
Replace webpack-manifest-plugin with webpack-assets-manifest for entrypoint chunk support.

Fixes #1252 and #1250


timkelty:
Ok, lets axe it then!

PR #1254, opened 2018-12-13
Remove manifest plugin

timkelty:
As discussed here: https://github.com/neutrinojs/neutrino/pull/1253#issuecomment-447009554

### DIFF
--- a/packages/react-components/index.js
+++ b/packages/react-components/index.js
@@ -10,7 +10,6 @@ module.exports = (neutrino, opts = {}) => {
     html: process.env.NODE_ENV === 'development' && {
       title: 'React Preview'
     },
-    manifest: process.env.NODE_ENV === 'development',
     externals: opts.externals !== false && {},
     style: {
       extract: {
--- a/packages/web/index.js
+++ b/packages/web/index.js
@@ -38,7 +38,6 @@ module.exports = (neutrino, opts = {}) => {
       hot: opts.hot !== false
     },
     style: {},
-    manifest: opts.html === false ? {} : false,
     clean: opts.clean !== false && {
       paths: [neutrino.options.output]
     },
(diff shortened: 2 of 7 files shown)


packages/react-components/index.js
```diff
@@ -10,7 +10,6 @@ module.exports = (neutrino, opts = {}) => {
     html: process.env.NODE_ENV === 'development' && {
       title: 'React Preview'
     },
-    manifest: process.env.NODE_ENV === 'development',
     externals: opts.externals !== false && {},
```
timkelty: @edmorley any idea what purpose this was serving, if any?

packages/web/index.js
```diff
@@ -38,7 +38,6 @@ module.exports = (neutrino, opts = {}) => {
       hot: opts.hot !== false
     },
     style: {},
-    manifest: opts.html === false ? {} : false,
```
edmorley: Should we now throw if a `manifest` option is passed? (Like occurs for the other removed options in this file)

timkelty: 👍

packages/react-components/index.js
edmorley: I have no idea :-/

@eliperelman / @helfi92 do you?

packages/web/index.js
```diff
@@ -82,6 +81,13 @@ module.exports = (neutrino, opts = {}) => {
     );
   }
 
+  if ('manifest' in options) {
+    throw new ConfigurationError(
```
edmorley: I don't suppose you could add a test for this? Similar to:
https://github.com/neutrinojs/neutrino/blob/21225bfae3d83ca6e687858f246d8a4f883399d9/packages/web/test/web_test.js#L257-L311

timkelty: @edmorley added test!

PR #1256, opened 2018-12-14
Issue 1198 - React create-project example page text is missing a space between words

fredmontet:
Just removed the line break and added a space. This closes the issue 1198.

### DIFF
--- a/packages/create-project/commands/init/templates/react/src/App.jsx
+++ b/packages/create-project/commands/init/templates/react/src/App.jsx
@@ -12,8 +12,7 @@ class App extends Component {
     return (
       <div className="App">
         <h1>
-          Welcome to
-          {name}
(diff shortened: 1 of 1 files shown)


packages/create-project/commands/init/templates/react/src/App.jsx
```diff
@@ -12,8 +12,7 @@ class App extends Component {
     return (
       <div className="App">
         <h1>
-          Welcome to
-          {name}
+          Welcome to {name}
```
eliperelman: Removing the extra line here could make airbnb/eslint complain. Could you instead switch the expression to be:

```jsx
Welcome to
{` ${name}`}
```

eliperelman:
Resolved in #1340.

Issue #1264, opened 2018-12-19
Test create-project using npm too in CI


edmorley:
Currently the create-project tests only create projects using yarn.

We should run at least a subset of the tests using npm too, to help catch things like #1261.

PR #1266, opened 2018-12-19
create-project: Run correct command for initial lint when using npm

### LINKED ISSUE #1261: create-project doesn't pass the `--fix` option correctly to `npm run lint`
When `create-project` runs, if the user chose a lint preset then ESLint is run with the `--fix` option to ensure the template project adheres to the chosen style guide out of the box. (We can't do this in the Neutrino repository since some of the style guides conflict eg regarding trailing commas or semi-colons.)

This currently works fine when using yarn (create-project runs `yarn lint --fix`), however fails with npm since `create-project` runs `npm run lint --fix` which passes the `--fix` to npm instead of ESLint.

The correct command for NPM is instead `npm run lint -- --fix`.

This affects both Neutrino 8 and Neutrino 9.

edmorley:
Previously the `--fix` option was not passed to ESLint when using NPM, since it requires that any options passed be preceded with `--`.

This caused failures during create-project when users picked a lint preset whose style differed from the template project, since the failures would not be fixed, and instead ESLint exit non-zero causing create-project to abort and delete the new project.

Fixes #1261.

(The create-project tests currently only run the preset test matrix using yarn - I've opened #1264 for adding coverage of NPM too.)

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -187,7 +187,7 @@ if desired.
 }
 ```
 
-Note: When using NPM any options must be preceeded with `--`, so the `lint:fix` command
+Note: When using npm any options must be preceeded with `--`, so the `lint:fix` command
 becomes `"npm run lint -- --fix"`.
 
 To tell ESLint to load Neutrino middleware or presets for its configuration, create a
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -187,6 +187,9 @@ if desired.
 }
 ```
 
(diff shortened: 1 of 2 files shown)


docs/usage.md
```diff
@@ -187,6 +187,9 @@ if desired.
 }
 ```
 
+Note: When using NPM any options must be preceeded with `--`, so the `lint:fix` command
```
eliperelman: npm should be lowercase.

PR #1276, opened 2019-01-03
Vue: Allow css modules to co-exist with global styles

### LINKED ISSUE #1243: vue/v9: css modules cannot co-exist with global styles (could in v8)
In v8, it was possible to have vue use css modules, but have co-existing global (non-module) styles.

With the same config in v9, this doesn't work:
```
Module parse failed: Unexpected token (19:0)
```

Working v8 example: https://github.com/timkelty/vue-scss/tree/v8-working
Broken v8 example: https://github.com/timkelty/vue-scss/tree/v9-broken

You can still achieve this in v9, but have to do thing differently:

1. set `modulesTest` equal to `test` (https://github.com/neutrinojs/neutrino/pull/1242 should help with that)
2. use `:global` in your css where you want global (non-module) styles.

-----

If using modules everywhere and requiring `:global` is the only way to do it in v9, that's fine, but it should probably be noted somewhere, as you could do it in v8. A the very least I think we should merge in https://github.com/neutrinojs/neutrino/pull/1242 so you don't have to set both `test` and `modulesTest`

timkelty:
Fixes #1243 

This converts the style rule to a oneOf, to allow for both global and module styes. The `oneOf` approach was cribbed from https://github.com/vuejs/vue-cli/blob/b559005ce64dc801c280e38194f28dd583292095/packages/%40vue/cli-service/lib/config/css.js#L72

> I'm also wondering if restructuring the style and style-modules rule created by style-loader into oneofs might make things easier to deal with? Wouldn't have to worry about multiple applied rules and excluding things, etc.
- https://github.com/neutrinojs/neutrino/issues/1243#issuecomment-450875429:

### DIFF
--- a/docs/migration-guide.md
+++ b/docs/migration-guide.md
@@ -321,6 +321,7 @@ module.exports = {
 ```
 
 - **BREAKING CHANGE** `@neutrinojs/style-loader` now creates a single `style` rule with `oneOfs`, instead of multiple rules for handling css modules ((#1281)[https://github.com/neutrinojs/neutrino/pull/1281]).
+- **BREAKING CHANGE** `@neutrinojs/vue` now adds oneOfs instead of replacing existing style rules ((#1276)[https://github.com/neutrinojs/neutrino/pull/1276]).
 - ESLint caching is now enabled by default for new projects, so it is recommended to specify `.eslintcache` as being
 ignored from your source control commits.
 
--- a/packages/neutrino/package.json
+++ b/packages/neutrino/package.json
@@ -28,7 +28,7 @@
     "deepmerge": "^1.5.2",
     "is-plain-object": "^2.0.4",
     "lodash.clonedeep": "^4.5.0",
-    "webpack-chain": "^5.0.1",
+    "webpack-chain": "^5.1.0",
     "yargs-parser": "^11.1.1"
   }
 }
--- a/packages/vue/README.md
+++ b/packages/vue/README.md
@@ -290,12 +290,14 @@ If the need arises, you can also compile `node_modules` by referring to the rele
 
 ### Rules
 
-The following is a list of additional rules and their identifiers this preset defines, in addition
+The following is a list of additional rules/oneOfs and their identifiers this preset defines, in addition
 to the ones provided by `@neutrinojs/web`, which can be overridden:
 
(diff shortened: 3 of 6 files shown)


packages/style-loader/index.js
```diff
@@ -44,9 +44,9 @@ module.exports = (neutrino, opts = {}) => {
         test: options.modulesTest,
         exclude: options.modulesExclude,
         ruleId: `${options.ruleId}${options.modulesSuffix}`,
-        styleUseId: `${options.styleUseId}${options.modulesSuffix}`,
-        cssUseId: `${options.cssUseId}${options.modulesSuffix}`,
-        extractId: `${options.extractId}${options.modulesSuffix}`,
+        styleUseId: `${options.styleUseId}`,
```
eliperelman: These don't need to be wrapped in template strings I think.

timkelty: In fact, they don't need to be there at all, now that the ids are the same…

timkelty:
Once https://github.com/neutrinojs/neutrino/pull/1281 is merged and https://github.com/neutrinojs/webpack-chain/pull/133 is release, I can rebase this PR and it should be good to go, bump the webpack-chain dep, and should be good to go.

timkelty:
Ok @edmorley, I've rebased this off `master` following the merge of https://github.com/neutrinojs/neutrino/pull/1281. Now all we're waiting on is a published version of webpack-chain with https://github.com/neutrinojs/webpack-chain/pull/133

edmorley:
Thank you for rebasing. The new version of webpack-chain was released yesterday:
https://github.com/neutrinojs/webpack-chain/pull/133#issuecomment-454759300

timkelty:
Doh, missed that! I'll change that and do some final tests now.

edmorley:
Travis is failing in the create-project test matrix. Perhaps it would be useful to have an additional test that verifies the output of `.toConfig()` so we can confirm the nested rules are laid out as expected?

timkelty:
@edmorley investigating. Initially this was the problem: https://github.com/neutrinojs/webpack-chain/issues/140, but I'm now using `.values()`

timkelty:
Ok @edmorley , Travis should be happy now.
I revised the test as well.

PR #1278, opened 2019-01-06
clean: Document removal of external directories

### LINKED ISSUE #362: Setting options.output to relative directory above project directory causes clean middleware to not work

priyankparker:
Fixes #362

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -303,22 +303,6 @@ module.exports = {
 };
 ```
 
-#### Clean external directories
-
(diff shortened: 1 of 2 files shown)


packages/react/README.md
```diff
@@ -303,6 +303,22 @@ module.exports = {
 };
 ```
 
+#### Clean external directories
```
edmorley: I think we should only add a section to the `@neutrinojs/clean` docs rather than the top-level presets too.

priyankparker: Thanks for the input. I have updated it.

PR #1281, opened 2019-01-14
style-loader: use oneof

timkelty:
https://github.com/neutrinojs/neutrino/issues/1243#issuecomment-454037093

Following this merge, we can rebase https://github.com/neutrinojs/neutrino/pull/1276 and merge that in.

### DIFF
--- a/packages/style-loader/index.js
+++ b/packages/style-loader/index.js
@@ -8,6 +8,7 @@ module.exports = (neutrino, opts = {}) => {
   const options = merge({
     test: neutrino.regexFromExtensions(['css']),
     ruleId: 'style',
+    oneOfRuleId: 'style',
     styleUseId: 'style',
     cssUseId: 'css',
     css: {
@@ -17,7 +18,6 @@ module.exports = (neutrino, opts = {}) => {
     modules,
     modulesTest,
     modulesSuffix: '-modules',
-    exclude: modules && modulesTest,
     loaders: [],
     extractId: 'extract',
     extract: {
@@ -39,14 +39,10 @@ module.exports = (neutrino, opts = {}) => {
   const rules = [options];
 
   if (options.modules) {
-    rules.push(
+    rules.unshift(
       merge(options, {
         test: options.modulesTest,
-        exclude: options.modulesExclude,
-        ruleId: `${options.ruleId}${options.modulesSuffix}`,
-        styleUseId: `${options.styleUseId}${options.modulesSuffix}`,
-        cssUseId: `${options.cssUseId}${options.modulesSuffix}`,
-        extractId: `${options.extractId}${options.modulesSuffix}`,
+        oneOfRuleId: `${options.oneOfRuleId}${options.modulesSuffix}`,
         css: {
           modules: options.modules
         }
@@ -76,11 +72,11 @@ module.exports = (neutrino, opts = {}) => {
 
     loaders.forEach(loader => {
       styleRule
-        .test(options.test)
-        .when(options.exclude, rule => rule.exclude.add(options.exclude))
-        .use(loader.useId)
-          .loader(loader.loader)
-          .when(loader.options, use => use.options(loader.options));
+        .oneOf(options.oneOfRuleId)
+          .test(options.test)
+          .use(loader.useId)
+            .loader(loader.loader)
+            .when(loader.options, use => use.options(loader.options));
     });
   });
 
--- a/packages/vue/index.js
+++ b/packages/vue/index.js
@@ -9,10 +9,10 @@ module.exports = (neutrino, opts = {}) => {
   const options = merge({
     style: {
       ruleId: 'style',
-      styleUseId: 'style'
-      // exclude: [],
-      // modulesTest: opts.style && opts.style.test ? opts.style.test : neutrino.regexFromExtensions(['css']),
-      // modulesSuffix: ''
+      styleUseId: 'style',
+      exclude: [],
+      modulesTest: opts.style && opts.style.test ? opts.style.test : neutrino.regexFromExtensions(['css']),
+      modulesSuffix: ''
     }
   }, opts);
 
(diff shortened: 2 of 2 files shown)


timkelty:
Tests are passing now!

timkelty:
@eliperelman if you have a chance for another review, take a look. I made some tweaks.

packages/style-loader/index.js
```diff
@@ -8,6 +8,7 @@ module.exports = (neutrino, opts = {}) => {
   const options = merge({
     test: neutrino.regexFromExtensions(['css']),
     ruleId: 'style',
+    oneOfId: 'normal',
```
edmorley: I'm wondering if we should just remove the configurability of `oneOfId`, `styleUseId`, `cssUseId`, `extractId` and so on, instead hardcoding the ids? (Leaving `ruleId` as the only configurable id)

The main reason to allow renaming is to avoid clashes, however that only really applies at the top level (eg if wanting to invoke `@neutrinojs/style-loader` twice with different settings).

Doing this would make the docs/options less daunting IMO.

packages/style-loader/README.md
```diff
@@ -243,12 +222,13 @@ ready to make changes.
...
+The following is a list of default rules/oneOfs and their identifiers which can be overridden:
 
 | Name | Description | NODE_ENV |
 | --- | --- | --- |
-| `style` | Allows importing CSS stylesheets from modules. By default contains two loaders named `extract` and `css` which use `MiniCssExtractPlugin.loader` and `css-loader`, respectively. If `options.extract` is `false`, then the `extract` loader is replaced by `style`, which uses `style-loader`. | all |
-| `style-modules` | Allows importing CSS Modules styles from modules. By default contains two loaders named `extract` and `css` which use `MiniCssExtractPlugin.loader` and `css-loader`, respectively. If `options.extract` is `false`, then the `extract` loader is replaced by `style`, which uses `style-loader`. | all |
+| `style` | The parent rule containing `oneOf` rules. | all |
+| `style.style` | Allows importing CSS stylesheets from modules. By default contains two loaders named `extract` and `css` which use `MiniCssExtractPlugin.loader` and `css-loader`, respectively. If `options.extract` is `false`, then the `extract` loader is replaced by `style`, which uses `style-loader`. | all |
```
edmorley: s/style.style/style.normal/ and similar for the modules one below

packages/style-loader/index.js
timkelty: I agree. I'll axe em.

timkelty: The other option would be to leave it as is, but just remove them from the docs, making them "unsupported" options.

timkelty: I pushed a version that removes them from the docs but doesn't change the code (you technically _can_ still set those options).

timkelty: @edmorley Here's where I ended up: I removed oneOfId, styleUseId, cssUseId, modulesOneOfId as configurable options (as they're all nested under the style `ruleId` namespace anyhow.

I left `extractId` as we seem to do that with other added plugins, and I could see that being a conflict someone had to override.

packages/style-loader/README.md
```diff
@@ -83,32 +79,11 @@ module.exports = {
...
-      extract: {
-        enabled: process.env.NODE_ENV === 'production',
-        loader: {},
-        plugin: {
-          filename: process.env.NODE_ENV === 'production'
-            ? 'assets/[name].[contenthash:8].css'
-            : 'assets/[name].css'
-        }
-      }
+      test: /\.scss$/,
```
edmorley: What do you think about replacing this example with `modules: false` or `extract: false` etc, since the changing of `test` here might suggest that's all that is required for supporting SCSS? (And it also overlaps with the section below)

packages/style-loader/index.js
```diff
@@ -8,16 +8,12 @@ module.exports = (neutrino, opts = {}) => {
...
     css: {
       importLoaders: opts.loaders ? opts.loaders.length : 0
     },
     style: {},
     modules,
     modulesTest,
-    modulesSuffix: '-modules',
-    exclude: modules && modulesTest,
     loaders: [],
     extractId: 'extract',
```
edmorley: > I left `extractId` as we seem to do that with other added plugins, and I could see that being a conflict someone had to override.

I think the main use-case for overriding `extractId` might be someone wanting to use `@neutrinojs/style-loader` twice and have mini-css-extract-plugin be added to the configuration twice under different Ids. However mini-css-extract-plugin isn't designed to be used more than once, and doesn't provide a way to filter the set of files on which it operates (ie so not possible to use plugin instance for css modules vs normal css).

As such, should we drop it too? We could always re-add if someone requests it -- it just seems like throughout all of the presets we've perhaps pre-emptively added configuration options before knowing of use-cases which actually require them.

packages/style-loader/README.md
```diff
@@ -243,12 +213,13 @@ ready to make changes.
 
 ### Rules
 
-The following is a list of default rules and their identifiers which can be overridden:
+The following is a list of default rules/oneOfs and their identifiers which can be overridden:
```
edmorley: Could you update the rules section for the `@neutrinojs/web` readme with these changes too? :-)

packages/style-loader/index.js
timkelty: > it just seems like throughout all of the presets we've perhaps pre-emptively added configuration options before knowing of use-cases which actually require them

I think that is true. Realistically, if any mw would need to provide this level of customization, it would be a third-party one, not a core one like this.

I'll get rid of it.

Do you think its worth auditing other middlewares for similar options to drop?

edmorley: > Do you think its worth auditing other middlewares for similar options to drop?

Yeah quite possibly (in other PRs) :-)

packages/style-loader/README.md
timkelty: I updated it, but only referenced the top-level `style` rule and referred to `@neutrinojs/style-loader`. If you think I should duplicate all the oneOf details here, I can.

packages/style-loader/index.js
timkelty: All set.

timkelty:
> Could you add the migration guide entry

Forgot! Now added 🎉

PR #1309, opened 2019-02-26
Add eslint-plugin-react-hooks to @neutrinojs/react

helfi92:
Resolves https://github.com/neutrinojs/neutrino/issues/1300.


helfi92:
Hm, I think it should be fine. According to https://reactjs.org/docs/hooks-intro.html#no-breaking-changes, hooks are 100% backwards-compatible.

PR #1317, opened 2019-03-08
Remove package.json config in favor of config files for husky and lint-staged

eliperelman:
Prevents the current warning you see in the console during development when committing.


edmorley (approved):
Thank you! 

(I'm on mobile at the moment, will be back and able to do the more involved reviews after the weekend)

PR #1320, opened 2019-03-13
Upgrade dependencies

eliperelman:


edmorley:
(Since lockfile changes don't affect the published packages, I've tended to use the label `development` instead, and not include in the milestone)

PR #1322, opened 2019-03-13
Remove targets from library preset to force es2015 target

### LINKED ISSUE #1279: @neutrinojs/library compiles down to IE9 but doesn't include polyfills
(Split out of https://github.com/neutrinojs/neutrino/issues/1251#issuecomment-449016651)

As of #994, `@neutrinojs/library` generates output that is compatible with IE9 (the idea being to make sure it's fully ES5 compatible). However since Babel has been told to target IE9, it wants to use regenerator-runtime (to polyfill async functions) which isn't available unless users manually import `@babel/polyfill` in their entrypoints. This results in:
`ReferenceError: regeneratorRuntime is not defined`

...even though for many people `regenerator-runtime` is actually unnecessary given the widespread async support in modern browsers (https://caniuse.com/#feat=async-functions).

The options appear to be:
a) Document the need to add a `@babel/polyfill` import to the entrypoint when using `@neutrinojs/library`
b) Automatically add a `@babel/polyfill` import to entrypoints when using `@neutrinojs/library`
c) Partially revert #994 and default to last 2 browser versions like before, but add an option (or document) how to generate IE9 compatible output for those that need it. 
d) Leave the default at `'ie 9'` (and do either A or B for people who use that default), but also make it easier to override the browsers list when using `@neutrinojs/library`, since it doesn't have a shorthand helper like `@neutrinojs/web` does. This would avoid having to use boilerplate like:

```js
    ['@neutrinojs/library', {
      // ...
      babel: {
        presets: [
          [
            require.resolve('@babel/preset-env'),
            {
              debug: false,
              useBuiltIns: 'entry',
              targets: {
                browsers: [
                  'last 2 Chrome versions',
                  'last 2 Firefox versions',
                  'last 2 Edge versions',
                  'last 2 Opera versions',
                  'last 2 Safari versions',
                  'last 2 iOS versions'
                ],
              }
            }
          ]
        ],
      },
    }],
```

@neutrinojs/core-contributors - thoughts?

My inclination is that `@neutrinojs/library` should not support Internet Explorer by default. ie: We should do option (c) above - or at least raise the target browser version to something inbetween IE9 and what it was before ("last 2 versions").

eliperelman:
Fixes #1279.

Just an initial implementation; happy to change if this seems incorrect.

### DIFF
--- a/packages/library/index.js
+++ b/packages/library/index.js
@@ -36,9 +36,7 @@ module.exports = (opts = {}) => {
           [require.resolve('@babel/preset-env'), {
             debug: neutrino.options.debug,
             useBuiltIns: 'entry',
-            targets: options.target === 'node' ?
-              { node: '8.10' } :
-              { browsers: 'ie 9' }
+            forceAllTransforms: true
           }]
         ]
       }, options.babel)
--- a/packages/library/index.js
+++ b/packages/library/index.js
@@ -35,8 +35,7 @@ module.exports = (opts = {}) => {
         presets: [
           [require.resolve('@babel/preset-env'), {
             debug: neutrino.options.debug,
-            useBuiltIns: 'entry',
-            forceAllTransforms: true
+            useBuiltIns: 'entry'
           }]
         ]
       }, options.babel)
--- a/packages/library/index.js
+++ b/packages/library/index.js
@@ -24,8 +24,13 @@ module.exports = (opts = {}) => {
       target: 'web',
       libraryTarget: 'umd',
       babel: {},
-      externals: {}
+      externals: {},
+      targets: {}
     }, opts);
+
+    if (options.targets === false) {
+      options.targets = {};
+    }
 
     Object.assign(options, {
       babel: babelMerge({
@@ -35,7 +40,8 @@ module.exports = (opts = {}) => {
         presets: [
           [require.resolve('@babel/preset-env'), {
             debug: neutrino.options.debug,
-            useBuiltIns: 'entry'
+            useBuiltIns: 'entry',
+            targets: options.targets
           }]
         ]
       }, options.babel)
--- a/packages/library/README.md
+++ b/packages/library/README.md
@@ -17,6 +17,12 @@
 - Automatically marks dependencies as external
 - Easily extensible to customize your library as needed
 
+**Important! This preset does not include 
+@babel/polyfill for size reasons. If you need
+polyfills in your library code, consider 
+importing @babel/polyfill, core-js, or other 
+alternative.**
+
 ## Requirements
 
 - Node.js ^8.10 or 10+
--- a/packages/library/README.md
+++ b/packages/library/README.md
@@ -249,15 +249,17 @@ module.exports = {
 
       // Disable cleaning the output build directory
       clean: false,
+      
+      // Target specific browsers or Node.js versions with @babel/preset-env
+      targets: {},
 
       // Add additional Babel plugins, presets, or env options
       babel: {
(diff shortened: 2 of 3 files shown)


packages/library/index.js
```diff
@@ -35,9 +35,7 @@ module.exports = (neutrino, opts = {}) => {
         [require.resolve('@babel/preset-env'), {
           debug: neutrino.options.debug,
           useBuiltIns: 'entry',
-          targets: options.target === 'node' ?
-            { node: '8.10' } :
-            { browsers: 'ie 9' }
+          forceAllTransforms: true
```
helfi92: Should this be `false` when running in development?

eliperelman: I don't *think* so, but there may be subtle bugs otherwise such as missing polyfills.

eliperelman:
Re-tagging for review to see if we can get approval or changes.

edmorley:
I'm a little hazy on this area, since it's been a while since I thought about #1279, but in that issue I believe the main problem was that `@neutrinojs/library` could output code that was broken unless people also imported a polyfill, but neither imported it for them, nor made it clear that they may need to so themselves.

The addition of the readme note in this PR address the final point of that, however the change to force all transforms seems to not be required, or at least orthogonal, given that the target of IE9 already activated most transforms, and that's why the pollyfill was required in the first place?

It seems that we need to decide the following:

1. Should the default target for `@neutrinojs/library` be "~all browsers", "modernish browsers", or both (using a combination of `main` and `module` in the resultant `package.json`)?
2. If for (1) we decide "~all browsers" or "both", is it better to implement this using a target of "IE9" or using an empty target and/or `forceAllTransforms: true`? (Is the latter more confusing for people who want to target only "modern browsers" and so are expecting to set a `target`?)
3. If for (1) we decide "~all browsers" or "both", is there a viable way to import the polyfill automatically (perhaps things have changed with Babel 7?) or should we just document? (And even if we just document, perhaps the create-project project should include the polyfill with a suitable comment saying to remove if not needed?)

Re the PR implementation as-is, I think the library readme's "Preset Options" section would also need updating for the new default options.

eliperelman:
Found this in the Babel docs for env:

> By default, it has the same behavior as previous presets to compile ES2015+ to ES5.

By not specifying any targets, it will compile to ES5 by default. I think that means we don't need the force transformation.

We unfortunately can't output 2 builds unless there are 2 compilations, and this is controlled by the user's `webpack.config.js`.

For the polyfill, we can prompt the user for it during create-project installation if you think that is a good path. I am still cautious to go down this route.

edmorley:
> We unfortunately can't output 2 builds unless there are 2 compilations, and this is controlled by the user's `webpack.config.js`.

Perhaps `.webpack()` could return an array? (Either all the time, or just for the library preset)

eliperelman:
It may be possible, but we would need a way to pass the preset twice with a different variation:

```js
// Should return [libraryWithMain, libraryWithModuleAndWithoutClean]
```

eliperelman:
I think we should explore better multiple configuration experience in a feature branch after v9. I may have some ideas, but I don't think it should block this.

edmorley (approved):
Should we also make it easier to override `targets`? (Like is possible with the web preset)
At the moment doing so requires quite a bit of boilerplate (see example in https://github.com/neutrinojs/neutrino/issues/1279#issue-398549571).

eliperelman:
@edmorley I added `targets` as a top-level option for this preset.

edmorley:
Looks good :-)

PR #1325, opened 2019-03-15
Update to Lerna v3 and switch to a 2-stage release process

### LINKED ISSUE #639: Devise and document an official release process
Now that Neutrino is more popular/has more moving parts, I think it's a good time to revisit our release/development process, changing/automating/documenting as appropriate.

We've already moved to the `release/8` branch system (where most development occurs on `master` rather than a `next` type branch), which is a good first step. However I think we should consider also:
* making releases multi-step - where:
  - step 1 is generating the code changes (version bumps/changelogs/...) and opening a PR,
  - step 2 is someone else reviewing the PR,
  - step 3 being merging the PR and pushing the packages to NPM.
  - ...this will allow us to catch any bugs in the automated changelog generation, as well as have a reviewer double-check that a breaking change wasn't accidentally included in a minor version bump etc.
* having a policy of releasing beta versions of new major version releases, before the final version
* having manual testing of a new release, before publishing to NPM (eg using `yarn pack`), until we're confident our end to end tests are fixed and working as expected
* perhaps moving the "publish to NPM" step to CI by making Travis publish when a tag is set
* only releasing at most once a week, unless it's an emergency

...plus anything else people can think of?

edmorley:
Releases will now occur in two phases, with the first being code-only changes that can be reviewed in a PR, followed by the tagging/publishing phase once it's merged.

Fixes #639.
Closes #1311.


edmorley:
These may help when reviewing:
https://github.com/lerna/lerna/tree/master/commands/version#readme
https://github.com/lerna/lerna/tree/master/commands/publish#readme

docs/contributing/development.md
```diff
@@ -159,3 +159,29 @@ the latest changes from the master branch.
 ❯ git remote add upstream https://github.com/neutrinojs/neutrino.git
 ❯ git pull upstream master
 ```
+
+## Releasing a new version
+
+1. Decide whether the new version should be a major/minor/patch/prerelease version bump.
+2. From the root of the Neutrino repository, run: `git checkout -b version-bump master`
```
eliperelman: Can we make this `git checkout -b version-bump upstream/master` to ensure we are using this from upstream or does that not matter? Also should we pair this with:

```sh
git fetch upstream --quiet && git checkout -b version-bump upstream/master
```

to ensure master is up to date?

docs/contributing/development.md
```diff
@@ -159,3 +159,29 @@ the latest changes from the master branch.
...
+3. Then run `yarn release:prepare` and pick the desired new version.
+4. Check the changes in the working directory and adjust if necessary.
+
+    If bumping to a new major version, or incrementing the pre-release version,
+    you will want to manually increase the presets' version of `neutrino` in
+    `peerDependencies`.
```
eliperelman: Maybe we could also add a note here on how to make this easier via `sed`? This is the command I use on macOS:

```sh
sed -i '' -- 's/"neutrino": "^9.0.0-0"/"neutrino": "^9.0.0-rc.0"/g' packages/*/package.json(D.)
```

package.json
```diff
@@ -20,8 +20,10 @@
     "docs:serve": "mkdocs serve",
     "link:all": "lerna exec yarn link",
     "lint": "eslint --no-eslintrc --c ./.eslintrc.js --cache --report-unused-disable-directives --format codeframe --ext js,jsx \".*.js\" packages",
-    "release": "lerna publish",
-    "release:preview": "lerna publish --skip-git --skip-npm",
+    "release:prepare": "lerna version",
+    "release:publish": "lerna publish from-package",
+    "release:publish-next": "yarn release:publish --dist-tag next",
+    "release:ci": "lerna publish prerelease --registry http://localhost:4873 --no-git-reset --yes",
```
eliperelman: I don't see `prerelease` as a valid positional for the `lerna publish` command, am I missing something?

https://github.com/lerna/lerna/tree/master/commands/publish#positionals

eliperelman: It also appears that `SKIP_CHANGELOG` isn't being respected:

https://travis-ci.org/neutrinojs/neutrino/jobs/506989820#L606

eliperelman:
```suggestion
    "release:ci": "SKIP_CHANGELOG=true lerna publish --registry http://localhost:4873 --no-git-reset --yes --canary prerelease",
```

eliperelman: I think updating the `version` script to the following should work:

```sh
[ -z $SKIP_CHANGELOG ] || yarn changelog --package && git add CHANGELOG.md
```

scripts/test-create-project-ci.sh
```diff
@@ -15,11 +15,7 @@ while ! nc -zw 1 localhost 4873; do sleep 1; done
 # The version will be bumped to the next pre-release suffix (`-0`) and the
 # package.json changes left in the working directory so that create-project
 # can read the correct version for installing matching monorepo packages.
-yarn release \
-  --registry http://localhost:4873/ \
-  --yes \
-  --skip-git \
-  --cd-version prerelease
+yarn release:ci
```
eliperelman: I also removed L5 as I moved this to the npm script.

eliperelman:
I tested out this PR in Travis with: https://github.com/neutrinojs/neutrino/pull/1327 and it appears to be working with my suggestions.

docs/contributing/development.md
edmorley: Good idea :-)

docs/contributing/development.md
edmorley: Yeah good idea. The example sed command in your comment didn't work for me, however this did:
`sed -i 's/"neutrino": "^9.0.0-0"/"neutrino": "^9.0.0-rc.0"/g' packages/*/package.json`

It might be worth testing that on your machine to check the OS X `sed` (which iirc is a little quirky) handles that form ok (or whether it needs an additional `-e` before the script command).

package.json
edmorley: > I don't see `prerelease` as a valid positional for the `lerna publish` command, am I missing something?

The `lerna publish` command is a superset of `lerna version`, so supports the options version supports:
https://github.com/lerna/lerna/tree/master/commands/version#semver-bump

I'd originally wanted to have the `yarn release:ci` command follow the actual release process as closely as possible - ie: make it do something like:
`yarn release:prepare prerelease --yes && yarn release:publish --registry http://localhost:4873 --yes`

However it turns out when `lerna publish` is run using either of the `from-git` or `from-package` options, lerna insists that the working directory is clean, which it won't be in CI, since we deliberately miss out the ```Commit the changes using: `git commit -a -m 'vN.N.N'``` step that is performed during a real release.

As such, I avoided this by making `yarn release:ci` perform the publish in one step, by running `lerna publish` in "do all steps" mode (ie: by specifying a version bump, rather than `from-git` or `from-package`).

> It also appears that `SKIP_CHANGELOG` isn't being respected:

Yeah I saw that locally on Windows too. Yarn must not propagate the modified env for sub-commands, strange.

edmorley: Re this suggested change - moving the `SKIP_CHANGELOG` here is a good idea, however the `from-package` option conflicts with the other options passed, so doesn't make sense in this context. See:
https://github.com/lerna/lerna/blob/a7ad9b60d27b390fde21fd2837f2d97320c4603e/commands/publish/index.js#L135-L143

Re `--canary`, we used to use that, however it used to append the Git SHA to the version in a way that created a semver-spec-violating version string if the Git SHA happened to have a leading zero, so stopped doing so. However it looks like that was fixed in lerna v3 (lerna/lerna/issues/1118), so we can switch back to using it :-)

edmorley: > However it looks like that was fixed in lerna v3 ([lerna/lerna/issues/1118](https://github.com/lerna/lerna/issues/1118)), so we can switch back to using it :-)

Ah `--canary` ignores the Neutrino version in `lerna.json` and instead tries to base it on the Git tags, which (a) doesn't work well on Travis (see https://travis-ci.org/neutrinojs/neutrino/jobs/507436675#L564), (b) means it uses a v8.x.x version rather than v9.x.x, since we don't yet have any Neutrino v9 Git tags.

Given we're not actually the packages in canary form, we don't benefit much from `--canary`, so I've just switched this back to how it was before.

PR #1328, opened 2019-03-16
Reconfigure airbnb and prettier for parity and make recommended fixes

eliperelman:
It appears that prettier wasn't actually configured to do anything, so this should now be in effect. I tried to use settings that would bring good parity with airbnb and prettier, so there will be some comma and max-len changes. Line length is now the recommended 80 from prettier.


.neutrinorc.js
```diff
@@ -1,65 +1,81 @@
...
-                'import/no-extraneous-dependencies': 'off'
-              }
+    [
+      './packages/airbnb',
+      {
+        // See the package.json `lint` script for which files are linted.
+        // Excludes are managed via `.eslintignore`.
+        eslint: {
+          baseConfig: {
+            extends: ['plugin:prettier/recommended'],
```
edmorley: Should we also use `'prettier/react'` here?

.neutrinorc.js
```diff
@@ -1,65 +1,81 @@
...
-                }
-              },
-              rules: {
-                // With Preact the use of `class` is recommended over `className`,
-                // so we have to add `class` to the ignore list, to prevent:
-                // `Unknown property 'class' found, use 'className' instead`
-                'react/no-unknown-property': ['error', { ignore: ['class'] }]
-              }
+            rules: {
+              'prettier/prettier': [
```
edmorley: Do most editors support reading the Prettier config out of the `.eslintrc.js` file, or should we move the prettier config to `.prettierrc.js` (which `eslint-plugin-prettier` reads too)? eg:
https://github.com/mozilla/treeherder/blob/master/.prettierrc.js

.eslintignore
```diff
@@ -8,3 +8,14 @@ packages/neutrino/test/fixtures/test-module/errorMiddleware.js
 # RC files. We then negate this ignore so they are still linted, but no longer
 # used as configuration by our development lint command.
 !packages/create-project/bin/commands/init/templates/*/.eslintrc.js
+
+# Ignore directories that may potentially have JS we don't need linted
+coverage
+docs
+build
+
+# Lint dotfiles
```
edmorley: Should we use the wildcard form here, to save us forgetting to add new files? eg:
`!.*.js`

.neutrinorc.js
```diff
@@ -1,65 +1,81 @@
...
-          ]
-        }
-      }
-    }]
-  ]
+            overrides: [
+              {
+                files: [
+                  'packages/create-project/commands/init/templates/**',
+                  'packages/create-project/commands/init/templates/*/.eslintrc.js',
```
edmorley: If `.eslintignore` is changed to use `!.*.js`, then locally I get an `import/no-extraneous-dependencies` error in: 
`packages/create-project/commands/init/templates/mocha/.mocharc.js`. 

Perhaps this entry should also be changed to use a wildcard? eg:
`packages/create-project/commands/init/templates/*/.*.js`

edmorley (approved):
Thank you for doing this!

You may already have planned to do this, but perhaps it's worth landing this PR after #1246 to reduce the amount of rebasing required, given the bulk of this PR can be discarded and re-generated?

edmorley:
Also one thing I found when using Prettier with Treeherder via `eslint-plugin-prettier`, was that if one enables the editor Prettier integration, then this typically uses Prettier directly, which means that when making changes to non-JS file types that Prettier happens to also support (such as `.css`, `.md`, `.yml`, `.html` etc) then large formatting changes end up being made that aren't covered by CI.

In Treeherder this was solved by running Prettier directly in CI too, eg:
https://github.com/mozilla/treeherder/blob/c79fc628232937c9419a1ef7b7926bd9e41b0ea8/package.json#L102

Though another approach would also just be to disable these types entirely for now, using a `.prettierignore` like so:

```
# Ignore Prettier-supported filetypes that we haven't yet converted to its style,
# so that "format on save" doesn't try to format them when making unrelated changes.
# When these are removed prettier will need to be run on Travis alongside ESLint,
# since the latter only supports JS/JSX.
*.css
*.html
*.json
*.md
*.yaml
*.yml
```

.neutrinorc.js
eliperelman: I've used both, and they each seem to work as expected. We can certainly add a command for it.

eliperelman:
I've made the recommended changes, and also reverted all the linting fixes until we are ready to move forward with this. This will fail CI until those fixes are pushed.

edmorley:
The changes look good; want to rebase and run `yarn prettier:fix`? :-)

Issue #837, opened 2018-05-01
Add more test coverage of non-default middleware


edmorley:
Some of the Neutrino middleware isn't used by any of the top level presets, and instead are meant as optional extras. For example ~`@neutrinojs/pwa`~, ~`@neutrinojs/stylelint`~, ~`@neutrinojs/image-minify`~, `@neutrinojs/copy` and ~`@neutrinojs/fork`~.

Whilst some of these have basic tests - they typically only check that the generated `.toConfig()` validates, and not that the build would actually succeed. The latter is covered by the create-project end to end tests for the top level presets, however we'll have to do something else for middleware that create-project doesn't know anything about.

This would help catch issues like:
https://github.com/mozilla-neutrino/neutrino-dev/pull/824#issuecomment-384942483

eliperelman:
As an update here, `@neutrinojs/stylelint`, `@neutrinojs/image-minify`, and `@neutrinojs/fork` have since been removed.

edmorley:
After the recent removals, the only middleware that is not tested via the E2E create-project tests is `@neutrinojs/copy`, which is straight-forwards enough that I think additional testing (beyond its unit test) is unnecessary.

PR #1330, opened 2019-03-18
Remove unmaintained pwa middleware

eliperelman:


edmorley (approved):
Works for me :-)

(This will also mean one less preset to worry about in #837)

eliperelman:
Yay!

PR #1334, opened 2019-03-24
jest: Exclude mjs from moduleFileExtensions

### LINKED ISSUE #1326: SyntaxError when using @neutrinojs/jest with MJS dependencies
In mozilla/treeherder/pull/4625, Travis is failing with:

```
$ node ./node_modules/jest/bin/jest
 FAIL  tests/ui/unit/react/bugfiler_test.jsx
  ● Test suite failed to run
    /home/travis/build/mozilla/treeherder/node_modules/node-fetch/lib/index.mjs:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import Stream from 'stream';
                                                                                                    ^^^^^^
    SyntaxError: Unexpected identifier
      at ScriptTransformer._transformAndBuildScript (node_modules/jest-runtime/build/ScriptTransformer.js:440:17)
      at Object.<anonymous> (node_modules/fetch-mock/src/server.js:1:42)
```

This is because [node-fetch](https://github.com/bitinn/node-fetch) v2 has a `.mjs` entrypoint, referenced from `module` in it's `package.json`, and Jest doesn't support ES6 style imports. 

However it does not occur with Jest defaults, only when using `@neutrinojs/jest`.

This is because:
* Neutrino does not configure dependencies to be transpiled using Babel, only the project's first party sources. (Or more specifically, it leaves Jest `transformIgnorePatterns` at the default of `["/node_modules/"]`).
* `@neutrinojs/jest` overrides Jest's default value for [moduleFileExtensions](https://jestjs.io/docs/en/configuration.html#modulefileextensions-array-string) (which is `["js", "json", "jsx", "ts", "tsx", "node"]`) to instead include additional extensions, such as `mjs`.

Until Jest supports ESM, the configuration needs to be changed so that this doesn't occur. It seems that `moduleFileExtensions` is probably the best place to fix this.

@neutrinojs/core-contributors, thoughts?

edmorley:
Since Jest does not yet support ES6 modules, so will error if imports resolve to `.mjs` files.

Fixes #1326.

### DIFF
--- a/packages/jest/src/index.js
+++ b/packages/jest/src/index.js
@@ -52,6 +52,9 @@ module.exports = (options = {}) => (neutrino) => {
       moduleDirectories: modulesConfig.length ? modulesConfig : ['node_modules'],
       moduleFileExtensions: neutrino.config.resolve.extensions
         .values()
+        // Jest does not yet support ES6 modules, see:
+        // https://github.com/facebook/jest/issues/4842
+        .filter(ext => ext !== '.mjs')
         .map(extension => extension.replace('.', '')),
       moduleNameMapper:
         Object
--- a/packages/jest/test/jest_test.js
+++ b/packages/jest/test/jest_test.js
@@ -1,6 +1,7 @@
 import test from 'ava';
 import airbnbPreset from '../../airbnb';
 import eslintPreset from '../../eslint';
+import reactPreset from '../../react';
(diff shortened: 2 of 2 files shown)


sarah-clements:
When will this be included in the next release?

PR #1355, opened 2019-04-13
Fix generation of changelog during release:prepare

### LINKED ISSUE #762: 'yarn version' fails with 'TypeError: Invalid Version: vundefined'
Running `yarn run version` on master (which is run by `yarn release*`; note is not the same as `yarn version`) fails with:

```
yarn run v1.5.1
$ yarn changelog --package && git add CHANGELOG.md
$ auto-changelog --remote upstream --commit-limit false --package
TypeError: Invalid Version: vundefined
    at new SemVer (C:\Users\Ed\src\neutrino-dev\node_modules\semver\semver.js:305:11)
    at compare (C:\Users\Ed\src\neutrino-dev\node_modules\semver\semver.js:578:39)
    at eq (C:\Users\Ed\src\neutrino-dev\node_modules\semver\semver.js:617:10)
    at Function.diff (C:\Users\Ed\src\neutrino-dev\node_modules\semver\semver.js:512:7)
    at parseReleases (C:\Users\Ed\src\neutrino-dev\node_modules\auto-changelog\lib\releases.js:37:66)
    at _callee$ (C:\Users\Ed\src\neutrino-dev\node_modules\auto-changelog\lib\run.js:110:52)
    at tryCatch (C:\Users\Ed\src\neutrino-dev\node_modules\regenerator-runtime\runtime.js:65:40)
    at Generator.invoke [as _invoke] (C:\Users\Ed\src\neutrino-dev\node_modules\regenerator-runtime\runtime.js:303:22)
    at Generator.prototype.(anonymous function) [as next] (C:\Users\Ed\src\neutrino-dev\node_modules\regenerator-runtime\runtime.js:117:21)
    at step (C:\Users\Ed\src\neutrino-dev\node_modules\auto-changelog\lib\run.js:31:191)
```

I believe this is because auto-changelog's `--package` option only looks in the monorepo root's `package.json`, and not the individual packages or `lerna.json`.

The fix should be as simple as re-adding the version to the root `package.json` (was previously removed before we used auto-changelog).

edmorley:
Previously we were using auto-changelog's `--package` option, which does not work with Lerna monorepos, since there is no version in the repo root `package.json`. Now the version is instead read from `lerna.json`.

In addition, `SKIP_CHANGELOG` has been removed, since:
(a) using environment variables in package.json scripts makes them incompatible with Windows
(b) we now use the `--no-git-reset` option with `release:ci`, so the working directory will be dirty regardless, so it's fine to modify the changelog too (and it means changelog generation has CI coverage)

Finally, some of the other release process steps have been converted to package.json scripts to simplify the steps.

Fixes #762.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,16 +15,20 @@
     "packages/*"
   ],
   "scripts": {
+    "changelog": "auto-changelog --latest-version $(yarn -s print-version)",
     "docs:bootstrap": "pip install -r docs/requirements.txt",
     "docs:serve": "mkdocs serve",
     "link:all": "lerna exec yarn link",
     "lint": "eslint --no-eslintrc --c ./.eslintrc.js --cache --report-unused-disable-directives --format codeframe --ext js,jsx .",
     "prettier:check": "prettier --ignore-path .eslintignore --check \"**/*.{css,html,js,jsx,json,md,yaml,yml}\"",
     "prettier:fix": "prettier --ignore-path .eslintignore --write \"**/*.{css,html,js,jsx,json,md,yaml,yml}\"",
-    "release:prepare": "lerna version",
+    "print-version": "jq -r .version lerna.json",
+    "release:prepare": "lerna version && prettier --write lerna.json",
+    "release:commit": "git commit --all --message v$(yarn -s print-version) --edit",
+    "release:tag": "git tag v$(yarn -s print-version) && git push upstream v$(yarn -s print-version)",
     "release:publish": "lerna publish from-package",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -15,16 +15,20 @@
     "packages/*"
   ],
   "scripts": {
+    "changelog": "auto-changelog --latest-version $(yarn -s print-version)",
```
edmorley: The use of bash subcommands (`$(...)`) makes this command not compatible with Windows (same as before this change), however this can be resolved by fixing CookPete/auto-changelog#106.

PR #1356, opened 2019-04-14
Refresh lockfile & bump in-range dependency versions

edmorley:
Bumping the versions in theory shouldn't be necessary, however package managers tend to only update sub-dependencies lazily, so this ensures users are on the latest versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,23 +40,23 @@
     "version": "yarn changelog"
   },
   "devDependencies": {
-    "auto-changelog": "^1.11.0",
-    "ava": "^1.3.1",
-    "eslint": "^5.15.1",
+    "auto-changelog": "^1.12.1",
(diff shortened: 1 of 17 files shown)


edmorley:
These changes were generated using [npm-check-updates](https://www.npmjs.com/package/npm-check-updates) v3, by:
1. `yarn global add npm-check-updates`
2. Check to see if any packages need excluding (eg deepmerge), using `yarn && yarn outdated`
3. From the root of the repo, run eg (remove the `-u` to dry run): 
  * `ncu -x deepmerge,yargs -u` 
  * `for d in packages/*; do ncu --dep prod,dev -x deepmerge,yargs --packageFile "${d}/package.json" -u;  echo; done`
4. Regenerate the lockfile using `rm -rf yarn.lock && yarn`

PR #1358, opened 2019-04-14
v9.0.0-rc.1

edmorley:
This was created by following the steps at:
https://master.neutrinojs.org/contributing/development/#releasing-a-new-version

Of note I've switched the Neutrino peer dependency to no longer use the caret range for now, since with pre-releases every release can be breaking. When we release v9.0.0 final, we should switch back to a caret version.


edmorley:
This ended up on my fork's `master` branch due to the way the documented process runs the Git checkout command. I've opened #1359 which among other things add `--no-track` to the command to prevent this in the future.

timkelty:
@edmorley @eliperelman Any chance of an `rc.1` release this week?

edmorley:
Continuing to follow [the release process](https://master.neutrinojs.org/contributing/development/#releasing-a-new-version):
* Git tagged and pushed using `yarn release:tag`
* Checked the tag/changelog looks ok: https://github.com/neutrinojs/neutrino/tree/v9.0.0-rc.1
* (Not in the steps, but since first time) I tested publishing to verdaccio and then pointing create-project at it -> worked fine
* Published to the `next` tag using `yarn release:publish-next`

At that last step, publishing failed halfway through, due to the `neutrino` package having the "require 2FA" box checked, and Lerna not supporting prompting for 2FA. (I do have 2FA enabled on my NPM account, however just for the admin currently and not for publishing).

This left the publish in a half-finished state, where about a third of the packages were published, which was not ideal.

To prevent this kind of thing (at least until Lerna fixes lerna/lerna/issues/1091), we should really have that "require 2FA" box ticked for all Neutrino packages or none of them, rather than just some - as otherwise it can cause the publishing to break halfway through.

This situation was made worse by `lerna publish from-package` not supporting resuming as it's supposed to. ie: It's designed to detect what needs publishing vs what's already published, but that didn't work for me, and instead it bailed after the first package complaining it was already published.

I next tried to fix up manually using `lerna exec 'npm publish ...'` however that gave other strange errors, so I had to resort to:
`for p in packages/*; do echo $p && npm publish ./${p} --tag next && echo; done` 

Anyway all looks good now (verified via `cd packages; for p in *; do echo $p && yarn info -s @neutrinojs/${p} dist-tags.next && echo; done`).

edmorley:
> at least until Lerna fixes [lerna/lerna/issues/1091](https://github.com/lerna/lerna/issues/1091)

Lerna now supports `--otp`:
https://github.com/lerna/lerna/releases/tag/v3.14.0

PR #1359, opened 2019-04-14
Further tweaks to the release process

edmorley:
* Call `yarn changelog` from `release:prepare` rather than during the `version` lifecycle, since `lerna.json` isn't always written by that point.
* Add missing "v" prefix to the changelog version number, and switch to using auto-changelog 1.13.0's new `--package <file>` support, so the changelog command doesn't need to use bash command substitution and so is compatible with Windows.
* Add `--no-track` to the Git checkout command, so that the new branch doesn't automatically track master (which means pushes end up on the wrong branch).
* Mention how to adjust the sed command for it to work on OS X.

### DIFF
--- a/docs/contributing/development.md
+++ b/docs/contributing/development.md
@@ -165,7 +165,7 @@ the latest changes from the master branch.
 1. Decide whether the new version should be a major/minor/patch/prerelease version bump.
 2. From the root of the Neutrino repository, run:
 
-    `git fetch upstream --quiet && git checkout -b version-bump upstream/master`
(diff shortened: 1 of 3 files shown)


edmorley:
I've updated the PR to take advantage of the new `--package <file>` support adding in auto-changelog v1.13.0 (https://github.com/CookPete/auto-changelog/commit/aafce3e62179d7505204db5fbfd98071c0ee8e48), which now means the `yarn changelog` command is compatible with Windows.

Issue #1360, opened 2019-04-14
Karma JS integration broken when "source-map-support" is listed in devDependencies. 


colmbrady:
### Bug or issue?

BUG

- What version of Neutrino are you using?

9.0.0-0

- Are you trying to use any presets? If so, which ones, and what versions?

"neutrino": "^9.0.0-0",
 "@neutrinojs/karma": "^9.0.0-0",
  "@neutrinojs/library": "^9.0.0-0",

- Are you using the Yarn client or the npm client? What version?

YARN

- What version of Node.js are you using?

v11.2.0

- What operating system are you using?

Mac

- What did you do?

See repo to reproduce https://github.com/colmbrady/neutrinojs-issue. **README contains a root cause description.**

- What did you expect to happen?

Karma tests to run with no error.

- What actually happened, contrary to your expectations?

Karma tests fail with:

```json
{
    "message": "Uncaught ReferenceError: require is not defined\nat test/index_test.js:1:1\n\nReferenceError: require is not defined\n    at test/index_test.js:1:1",
    "str": "Uncaught ReferenceError: require is not defined\nat test/index_test.js:1:1\n\nReferenceError: require is not defined\n    at test/index_test.js:1:1"
  }
```

colmbrady:
https://github.com/colmbrady/neutrino/blob/v8.3.0/packages/karma/index.js#L61

Seems like the karma integration used to omit plugins previously.

I wonder is the fix for this is to ensure that the plugins are omitted again at this line:

https://github.com/neutrinojs/neutrino/blob/master/packages/karma/index.js#L80

edmorley:
Hi! The testcase repo is missing the `.neutrinorc.js` - could you add it? :-)

colmbrady:
https://github.com/colmbrady/neutrinojs-issue/blob/master/.neutrinorc.js

Woops. Sorry, not sure how that did not get comitted first time.

I think I have noticed that my issues has actually been addressed by this commit.

https://github.com/neutrinojs/neutrino/blob/985e17992ea40618089940ccaba832cce347ec1c/packages/karma/index.js#L59

So perhaps we can just close this?

edmorley:
Ah yes this is a duplicate of #1251, which is fixed by #1319 and due to be released shortly in 9.0.0-rc.1 (#1358).

PR #1361, opened 2019-04-14
Reinstate 8.x behaviour where Karma config preset would omit the Webpa…

colmbrady:
…ck plugins before starting Karma server.

Resolves https://github.com/neutrinojs/neutrino/issues/1360

### DIFF
--- a/packages/karma/index.js
+++ b/packages/karma/index.js
@@ -77,7 +77,7 @@ module.exports = (options = {}) => neutrino => {
               warnings: true,
             },
           },
-          webpack: omit(neutrino.config.toConfig(), ['entry']),
+          webpack: omit(neutrino.config.toConfig(), ['entry', 'plugins']),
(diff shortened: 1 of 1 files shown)


colmbrady:
I had a quick look at how I could write a test, but there was no existing coverage on that part of the code. Happy to add a test with some guidance.

edmorley:
Hi! Thank you for the PR. 

The reason this change was made in Neutrino 9, is so that project that rely upon plugins such as ProvidePlugin or DefinePlugin aren't broken when running tests. There might be another fix we can do to both preserve that behaviour and address #1360 - however to figure that out I'll need to be able to reproduce #1360 first (https://github.com/neutrinojs/neutrino/issues/1360#issuecomment-483144981).

edmorley:
Closing since this issue was fixed on `master` by #1358 and will be released in #1358 soon.
The PR is much appreciated however :-)

PR #1365, opened 2019-04-16
Fix using create-project with a single project selection

eliperelman:
Using create-project with only a single project selected, e.g., no linting or testing preset, results in:

Screen Shot 2019-04-16 at 2 17 50 PM


helfi92:
Image is not here xD

Screen Shot 2019-04-16 at 3 23 45 PM

eliperelman:
@helfi92 weird lol, how about now?

helfi92 (approved):
Lovely!

PR #1372, opened 2019-04-25
Use native Node.js for pulling version

eliperelman:


package.json
```diff
@@ -22,7 +22,7 @@
     "lint": "eslint --no-eslintrc --c ./.eslintrc.js --cache --report-unused-disable-directives --format codeframe --ext js,jsx .",
     "prettier:check": "prettier --ignore-path .eslintignore --check \"**/*.{css,html,js,jsx,json,md,yaml,yml}\"",
     "prettier:fix": "prettier --ignore-path .eslintignore --write \"**/*.{css,html,js,jsx,json,md,yaml,yml}\"",
-    "print-version": "jq -r '\"v\" + .version' lerna.json",
+    "print-version": "node -e \"process.stdout.write(require('./lerna.json').version)\"",
```
edmorley: Iirc `node -p` prints the statement (on mobile or I'd check)

eliperelman: It does, but it inserts a blank line (like `console.log` does), so this ensures no newlines via `process.stdout.write`.

edmorley: Bash strips that off. Using `"print-version": "node -p \"require('./lerna.json').version\"",` I get:

```
$ echo "z$(yarn -s print-version)z"
z9.0.0-rc.1z
```

eliperelman: Yay!

edmorley (approved):

PR #1373, opened 2019-04-25
Release process revisions

eliperelman:


eliperelman:
@edmorley I wanted to also let you know that I went through all the npm packages and set them all to require 2fa.

edmorley (approved):
Looks great! And thank you for setting the 2FA bit on the packages :-)

PR #1396, opened 2019-05-21
Detect yarn in create-project via environment

### LINKED ISSUE #1260: create-project defaults to yarn if present even when run via npx
To try and reproduce #1249 locally I was hoping to run create-project using npm rather than yarn.

However running `npx @neutrinojs/create-project somename` always results in yarn being used (if it's installed) rather than npm, for the various commands that are run during project creation (such as package installation and initial lint fix).

We should make create-project use the same package manager as it was run with.

eliperelman:
Fixes #1260.


helfi92 (approved):
👌

PR #1447, opened 2019-08-17
Update eslint-config-airbnb and eslint-config-airbnb-base

edmorley:
* Combines #1443 and #1444 (since they are dependant on each other)
* Adds the new AirBnB hooks preset to `@neutrinojs/airbnb`
* Removes the React `version` option from `@neutrinojs/airbnb` since it's now configured within `eslint-config-airbnb`
* Switches other React `version` usages to the new `'detect'` mode
* Overrides AirBnB's configuration for `react/state-in-constructor` since `@neutrinojs/react` supports class properties, so using the `always` mode is unnecessary.

Closes #1443 and closes #1444.

### DIFF
--- a/packages/airbnb-base/package.json
+++ b/packages/airbnb-base/package.json
@@ -24,7 +24,7 @@
   },
   "dependencies": {
     "@neutrinojs/eslint": "9.0.0-rc.3",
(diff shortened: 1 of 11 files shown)


edmorley:
@neutrinojs/core-contributors any comments before I merge? :-)

Issue #1423, opened 2019-06-24
ESLint 6.0.0 compatibility


edclement:
### Bug or issue?

Please try to answer the following questions:

- What version of Neutrino are you using? **9.0.0-rc3**
- Are you trying to use any presets? If so, which ones, and what versions?
  - **"@neutrinojs/copy": "9.0.0-rc.3"**
  - **"@neutrinojs/jest": "9.0.0-rc.3"**
  - **"@neutrinojs/react": "9.0.0-rc.3"**
  - **"@neutrinojs/standardjs": "9.0.0-rc.3"**
- Are you using the ~~Yarn~~ client or the **npm** client? What version?
- What version of Node.js are you using? **10.16.0**
- What operating system are you using? **Ubuntu 18.04.2 LTS**
- What did you do? **Update ESLint to 6.0.0**
- What did you expect to happen? **Linting to work**
- What actually happened, contrary to your expectations?
```
# npm run lint

> lint /home/<REDACTED>/Git/<REDACTED>/<REDACTED>
> eslint --format codeframe --ext js ./src ./e2e ./*.js .neutrinorc.js

Error: Cannot read config file: /home/<REDACTED>/Git/<REDACTED>/<REDACTED>/.eslintrc.js
Error: Cannot find module 'eslint/lib/config/config-ops'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:690:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/home/<REDACTED>/Git/<REDACTED>/<REDACTED>/node_modules/@neutrinojs/eslint/index.js:2:32)
    at Module._compile (internal/modules/cjs/loader.js:776:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:787:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! <REDACTED>@0.1.0 lint: `eslint --format codeframe --ext js ./src ./e2e ./*.js .neutrinorc.js`
npm ERR! Exit status 2
npm ERR! 
[...]

PR #1464, opened 2019-09-09
Stop using private ESLint APIs for ESLint config merging

edmorley:
In #1182, the previous broken ESLint config merging functionality was replaced by using an internal ESLint `merge` function from `config-ops`.

Unfortunately in ESLint 6 that function no longer exists in a form that is easy for us to use, so in order to support ESLint 6, I have replaced it with simpler manual merges. These are virtually identical, with one difference - `rules` are now merged by a later rule entry completely replacing the first, rather than updating it.

For example, merging these:
`'foo': ['error', {'someSetting': false}]`
`'foo': 'warn'`

...used to give:
`'foo': ['warn', {'someSetting': false}]`

...but now results in:
`'foo': 'warn'`

I think this difference is a reasonable compromise for not having to rely on ESLint internals and/or implement our own more complex merging. I also believe hitting this will be rare, since:
* it doesn't affect merging with rules defined inside an `extends` external file
* it won't make a difference for rules that were already at default settings (or that were overridden with custom settings in the later rule definition being merged in)

This more simplistic merging behaviour is also what's used by Neutrino 8 (#1182 landed for Neutrino 9 only) - and we still have the other bug fixes included in #1182 that were the primary motivation for refactoring in the first place.

Refs #1423.

### DIFF
--- a/packages/airbnb-base/index.js
+++ b/packages/airbnb-base/index.js
@@ -1,5 +1,4 @@
 const lint = require('@neutrinojs/eslint');
-const { merge: eslintMerge } = require('eslint/lib/config/config-ops');
 const {
   rules: airbnbBaseStyle,
 } = require('eslint-config-airbnb-base/rules/style');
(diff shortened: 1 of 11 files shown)


edclement:
Took a quick peak. Looks good to me...

timkelty (approved):

edmorley:
Thank you for the reviews :-)

PR #1465, opened 2019-09-09
Update eslint-config-standard and switch from -jsx to -react

edmorley:
Updates the ESLint standardjs presets to latest, and switches from `eslint-config-standard-jsx` to `eslint-config-standard-react`, since the latter is a superset of the former and required for full React support.

The new configs set the React version and plugins for us, so we no longer need to set them ourselves.

Fixes #1449 and fixes #1450.

### DIFF
--- a/packages/standardjs/index.js
+++ b/packages/standardjs/index.js
@@ -12,13 +12,9 @@ module.exports = ({ eslint = {}, ...opts } = {}) => neutrino => {
           ...baseConfig,
           extends: [
             require.resolve('eslint-config-standard'),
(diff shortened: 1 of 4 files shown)


edmorley:
@neutrinojs/core-contributors? :-)

timkelty:
@edmorley I'll take a look this morning!

edmorley:
@timkelty? :-)

timkelty:
Whoops, thanks for the nudge. Looking now!

timkelty (approved):
👏 LGTM!

edmorley:
Thank you :-)

PR #1468, opened 2019-09-12
feat: add typings for neutrino config and api

eddyw:
Related to: https://github.com/neutrinojs/neutrino/issues/1129#issuecomment-526181211

**Motivation**: It's hard to work with Neutrino if there is no code editor intellisense or/and auto-completion. Even though, `.neutrinorc.js` is a JS file, we could use JSDocs to specify the type of the configuration object and get intellisense:
```javascript
/**
 * @type {import('neutrino').Configuration}
 */
const config = { ... /* << ---- code editor suggestions work */ }
config.    << ---- code editor suggestions work here too

module.exports = config
```

**PR:**
– It adds typings for Neutrino api (class) and configuration
– It adds typings tests for TS
– It adds typings tests for JS (when using JSDocs)

**Notes**: `tsconfig.json` file has `checkJs` and `allowJs` options set to `true`, so it works checking the types of JSDocs. Tests for JS typings are not extensive, just as good as they can be.

This PR doesn't introduce any breaking changes.

### DIFF
--- a/packages/neutrino/package.json
+++ b/packages/neutrino/package.json
@@ -7,10 +7,6 @@
   "bin": {
     "neutrino": "bin/neutrino.js"
   },
-  "files": [
-    "index.js",
-    "types/*.d.ts"
-  ],
   "keywords": [
     "neutrino",
     "webpack",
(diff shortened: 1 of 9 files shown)


edmorley:
Closing for inactivity. Happy to reopen if the PR is updated :-)

PR #1477, opened 2019-10-10
Allow passed extensions to override defaults

### LINKED ISSUE #1404: [v9] options.extensions not used
### Bug

- What version of Neutrino are you using?
9.0.0-rc.3
- Are you trying to use any presets? If so, which ones, and what versions?
@neutrinojs/react@9.0.0-rc.3 (should be visible with compile-loader only)
- Are you using the Yarn client or the npm client? What version?
npm@6.7.0
- What version of Node.js are you using?
node@10.8.0
- What operating system are you using?
macOS
- What did you do?
1. Run `npx @neutrinojs/create-project@9.0.0-rc.3` with only `React` on
2. Edited `.neutrinorc.js` to look like this:
```
const react = require('@neutrinojs/react')

module.exports = {
  options: {
    root: __dirname,
    extensions: ['mjs', 'jsx', 'js', 'ts'], // <- added 'ts' extension
  },
  use: [
    react(),
  ],
}
```
3. Created empty `src/index.ts` file and removed `src/index.js`/
4. Run `webpack --mode production`
- What did you expect to happen?
Build finishes successfully because `src/index.ts` is resolved by webpack and processed by babel
- What actually happened, contrary to your expectations?
Build fails with error:
```
ERROR in multi ./src/index
Module not found: Error: Can't resolve '/Users/rkostrzewski/GitHub/coinsy/web/src/index' in '/Users/rkostrzewski/GitHub/coinsy/web'
 @ multi ./src/index index[0]
```

Based on quick investigation this might occur because `options.extensions` is not used in `Neutrino.getOptions` instead `moduleExtensions` variable is used that is initialised only with `source` from `neutrino/extensions` module (changing that declaration to include `options.extensions` yields a successful build however on v8 the code looks the same yet extension `'foo'` works in v8)

timkelty:
Fixes #1404

### DIFF
--- a/packages/neutrino/Neutrino.js
+++ b/packages/neutrino/Neutrino.js
@@ -27,11 +27,12 @@ module.exports = class Neutrino {
   }
 
   getOptions(opts = {}) {
-    let moduleExtensions = new Set(source);
     const options = {
       debug: false,
+      extensions: new Set(source),
       ...clone(opts),
     };
+    let moduleExtensions = options.extensions;
 
     if ('node_modules' in options) {
       throw new ConfigurationError(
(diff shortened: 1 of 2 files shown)


timkelty:
@neutrinojs/core-contributors, @edmorley?

PR #1481, opened 2019-11-03
Drop support for Node.js 8

edmorley:
Since it's EOL as of 2019-12-31:
https://nodejs.org/en/about/releases/

...and if we can't support it for the full Neutrino 9 lifecycle, it's best we don't support it from the start at all.

The Travis matrix has also been updated to:
* test against Node.js 13
* only test against Node.js 12 on `master` to reduce the time taken for Travis to complete on PRs/other branches (we only get 5 concurrent builds, and previously there were 6 jobs in the matrix)


timkelty (approved):
👍💯

PR #1484, opened 2019-11-04
Change useBuiltIns to usage

timkelty:
Resolves https://github.com/neutrinojs/neutrino/issues/1470

### DIFF
--- a/docs/migration-guide.md
+++ b/docs/migration-guide.md
@@ -97,6 +97,11 @@ upgrade that may not be outlined here.
 
 A list of changes is detailed below for migrating:
 
+- **BREAKING CHANGE** The Web, React, Vue, Node.js, and Library presets now use `useBuiltIns: 'usage'` as the
+default when configuring `@babel/preset-env` for polyfills. As a result you no longer need to manually import
+`core-js`, `@babel/polyfill`, or `regenerator-runtime/runtime` in your entry points, as was neceesary when
+`useBuiltIns: 'entry'` was the default setting. **Note:** `core-js` still needs to be present in your `package.json`
+for this default to take effect. [#1470](https://github.com/neutrinojs/neutrino/issues/1470)
 - **BREAKING CHANGE** The minimum supported Node.js version is 10. Node.js 6-9 are no longer supported
 [#792](https://github.com/neutrinojs/neutrino/pull/792) [#1481](https://github.com/neutrinojs/neutrino/pull/1481).
 - **BREAKING CHANGE** Consumption of middleware and presets are now dependent on the external CLIs and
--- a/docs/migration-guide.md
+++ b/docs/migration-guide.md
@@ -97,11 +97,6 @@ upgrade that may not be outlined here.
 
 A list of changes is detailed below for migrating:
 
-- **BREAKING CHANGE** The Web, React, Vue, Node.js, and Library presets now use `useBuiltIns: 'usage'` as the
-default when configuring `@babel/preset-env` for polyfills. As a result you no longer need to manually import
-`core-js`, `@babel/polyfill`, or `regenerator-runtime/runtime` in your entry points, as was neceesary when
-`useBuiltIns: 'entry'` was the default setting. **Note:** `core-js` still needs to be present in your `package.json`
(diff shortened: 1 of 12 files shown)


packages/library/README.md
```diff
@@ -19,9 +19,10 @@ for Web or Node.js targets.
 - Automatically marks dependencies as external
 - Easily extensible to customize your library as needed
 
-**Important! This preset does not include @babel/polyfill for size reasons. If
-you need polyfills in your library code, consider importing @babel/polyfill,
-core-js, or other alternative.**
+\*\*Important! If you need polyfills in your library code, consider including
```
edmorley: I couldn't tell if the `\*\*` here and in the other files changed was a shortform syntax that I wasn't familiar with (that the GitHub diff viewer didn't support, but the docs generator does), or if the intention was to display asterisks without bolding? From looking at the preview it seems that the text isn't bold in the docs:
https://deploy-preview-1484--neutrinojs.netlify.com/packages/library/

I think I slightly prefer the previous styling - though the bold is perhaps slightly overwhelming, so if there was something halfway between that might work? eg Making just the "Important!" bold? Or else using one of the banner styles here:
https://squidfunk.github.io/mkdocs-material/extensions/admonition/#types

docs/migration-guide.md
```diff
@@ -97,6 +97,11 @@ upgrade that may not be outlined here.
 
 A list of changes is detailed below for migrating:
 
+- **BREAKING CHANGE** The Web, React, Vue, Node.js, and Library presets now use `useBuiltIns: 'usage'` as the
```
edmorley: The breaking changes here are listed are listed in roughly order of potential impact/number of people affected. What do you think about moving the new entry lower down, perhaps below the `With the necessity of external CLIs means the removal of most of the Neutrino CLI's commands` line?

packages/library/README.md
timkelty: Yeah not intentional…my editor auto-prettier'd on save and did this. I'll fix.

docs/migration-guide.md
timkelty: ah yep

packages/library/README.md
edmorley: Ah that makes sense! I was thinking it must have been intentional given three files formatted the same hehe

timkelty:
@edmorley do you think each preset should have a "If you need polyfills in your code…" callout?

Not sure why currently only Library and React have them currently…

edmorley:
Yeah think adding to all of them makes sense (or at least all of the web related ones; less sure about Node)

timkelty:
> Yeah think adding to all of them makes sense (or at least all of the web related ones; less sure about Node)

I added the note to everything that used `useBuiltIns: usage`, which included Node.

edmorley:
> Remove note from react

I think we should still include the note for the React preset. I think maybe you saw the `useBuiltIns: true` mention in React and so why the docs entry was removed for that? If so, the `useBuiltIns: true` is actually a different setting with the same name - it's passed to the React babel preset ([see docs here](https://babeljs.io/docs/en/babel-preset-react#usebuiltins)) rather than preset-env. 

Perhaps a comment next to it emphasising it's different would be useful, since it confused me yesterday too when reviewing the PR?

timkelty:
@edmorley with the eagle eye! Good catch. 🦅 👁

edmorley:
> Perhaps a comment next to it emphasising it's different would be useful, since it confused me yesterday too when reviewing the PR?

I'm really sorry I meant as a code comment in `packages/react/index.js` rather than in the README. Though happy to go with whatever you think is best :-)

timkelty:
@edmorley yeah I thought about that after I pushed…one sec. Think I should keep both?

edmorley:
@timkelty Babel 7.7.0 (released in last day or so) added a new `useSpread` option to the React preset (only) that supersedes the `useBuiltins` React option, so I'll likely be replacing the option once I've read up more. 

Also even if we don't switch to `useSpread`, users will rarely need to change the React preset option, since whatever the React plugins output gets transpiled after - ie: using built-ins for the React output is desirable either way, so something we perhaps don't need to worry users about.

PR #1488, opened 2019-11-10
react: Use babel-preset-react's new useSpread option

edmorley:
A new `useSpread` option was added in Babel 7.7.0 (babel/babel#10572), which supersedes the babel-preset-react's existing `useBuiltIns` option:
https://babeljs.io/docs/en/babel-preset-react#usespread

Enabling this means that the JSX-> JS conversion will use an inline object with spread elements (for code that spreads props), rather than Babel's extend helper or `Object.assign`.

`@babel/env` will still transpile these down if required for the target browsers (the only reason `useSpread` is not already enabled by default is that for the upstream project this is a breaking change, since they cannot assume that users of `babel-preset-react` are always using it alongside `@babel/env`, even though most likely are).

### DIFF
--- a/packages/react/index.js
+++ b/packages/react/index.js
@@ -43,9 +43,10 @@ module.exports = (opts = {}) => neutrino => {
             {
               // Enable development helpers both in development and testing.
               development: process.env.NODE_ENV !== 'production',
-              // Use the native built-in instead of polyfilling.
-              // Don't confuse this with `@babel/preset-env`'s `useBuiltIns` option, they are different.
-              useBuiltIns: true,
+              // When spreading props, use inline object with spread elements directly instead of
+              // Babel's extend helper or Object.assign. @babel/env will still transpile these down
+              // if required for the target browsers.
+              useSpread: true,
             },
           ],
         ],


packages/react/index.js
```diff
@@ -43,9 +43,10 @@ module.exports = (opts = {}) => neutrino => {
             {
               // Enable development helpers both in development and testing.
               development: process.env.NODE_ENV !== 'production',
-              // Use the native built-in instead of polyfilling.
-              // Don't confuse this with `@babel/preset-env`'s `useBuiltIns` option, they are different.
-              useBuiltIns: true,
```
helfi92: We need to remove `useBuiltIns` to use the useSpread option?

helfi92: Nevermind, just saw that it only accepts only accepts useBuiltIns or useSpread in  https://github.com/babel/babel/pull/10572/files#diff-150912f9d993ec47518d06f0b1f877bcR182.

helfi92 (approved):
LGTM

packages/react/index.js
edmorley: Yeah sorry meant to mention that in the PR description.

edmorley:
Thank you for the review :-)

chiccogirl (commented):
Hhhh

PR #1554, opened 2020-04-11
Exclude all extra options of `babel-loader`

constgen:
We should extract only Babel options to Jest but not `babel-loader` options as it has some extra options that can break Babel configuration validation. For example like this

```
Unknown option: .cacheCompression. Check out https://babeljs.io/docs/en/babel-core/#options for more information about options.
```
jest

Neutrino core presets don't use such options but other authors may use them and `@neutrinojs/jest` will fail

### DIFF
--- a/packages/jest/src/index.js
+++ b/packages/jest/src/index.js
@@ -24,7 +24,7 @@ module.exports = (options = {}) => (neutrino) => {
       'cacheDirectory',
       'cacheCompression',
       'cacheIdentifier',
-      'customize'
+      'customize',
     ];
     const compileRule = neutrino.config.module.rules.get('compile');
     const babelOptions = compileRule
--- a/packages/jest/src/index.js
+++ b/packages/jest/src/index.js
@@ -20,9 +20,10 @@ module.exports = (options = {}) => (neutrino) => {
   }
 
   neutrino.register('jest', (neutrino) => {
+    const babelLoaderOptionsNames = ['cacheDirectory', 'cacheCompression', 'cacheIdentifier', 'customize'];
     const compileRule = neutrino.config.module.rules.get('compile');
     const babelOptions = compileRule
-      ? omit(compileRule.use('babel').get('options'), ['cacheDirectory'])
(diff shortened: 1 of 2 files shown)


edmorley:
Could you also add a test to:
https://github.com/neutrinojs/neutrino/blob/master/packages/jest/test/jest_test.js

It needed call the Jest API; just checking that a couple of those options aren't passed through to the generated config should be enough.

constgen:
This happens when you edit with Web browser.

These options were never exposed to Jest config. They are used in `babel-jest`. This preset passes babel config to `babel-jest` using environment variable https://github.com/constgen/neutrino-dev/blob/7d15b3a6fd82a564b905de249a57e060de8e1dd5/packages/jest/src/transformer.js#L3. I am not sure this is a good idea to test this dirty hack even if it is technically possible. What do you think?

edmorley:
Sorry in my comment above, there was a typo: s/needed/needn't/

ie: The test should not call Jest's API, but just test the generated config returned by `.jest()`, similar to:
https://github.com/neutrinojs/neutrino/blob/aee187e20dadcda4912afa39288d478f27881fcf/packages/jest/test/jest_test.js#L100-L113

constgen:
I understand. But you see, the changes doesn't affect the Jest config returned by `.jest()`

edmorley:
Oh sorry I see I forgot the setting of `process.env.JEST_BABEL_OPTIONS` happened here and not later. I'd still like a test for this - I think it's fine for the test to call `.jest()` and validate that `process.env.JEST_BABEL_OPTIONS` is set to the expected value (if anything we should have had a test for that already, but don't). I agree the env var approach is suboptimal, but it's all that the Jest API gave us - so all the more reason it needs a test :-)

constgen:
Tests added.

There is a chance the same issue may appear here https://github.com/constgen/neutrino-dev/blob/7d15b3a6fd82a564b905de249a57e060de8e1dd5/packages/compile-loader/index.js#L25 . But this is another story.

constgen:
@edmorley I think it is ready to be merged

PR #1577, opened 2020-05-21
jest: Fix configuration of Webpack aliases

### LINKED ISSUE #1552: Preact + Jest: "Could not locate module react mapped as ... node_modules/preact/compat$1"

- What version of Neutrino are you using? 9.1.0
- Are you trying to use any presets? If so, which ones, and what versions? Preact, Jest
- Are you using the Yarn client or the npm client? What version? npm 6.14.4
- What version of Node.js are you using? 13.12.0
- What operating system are you using? macOS 10.14.6

---

I am experimenting with converting a codebase from React to Preact. When running the Jest tests, I see the following error:

```
    Configuration error:
    
    Could not locate module react mapped as:
    /<project-dir>/node_modules/preact/compat$1.
    
    Please check your configuration for these entries:
    {
      "moduleNameMapper": {
        "/^react$/": "/<project-dir>/node_modules/preact/compat$1"
      },
      "resolver": undefined
    }

      at createNoMappedModuleFoundError (node_modules/jest-resolve/build/index.js:540:17)
      at Object.<anonymous> (node_modules/react-router/cjs/react-router.js:5:29)
```

The dependency `node_modules/preact/compat` is installed

My suspicion (unconfirmed) is that this is happening due to the trailing `$1` in the rewritten module, which is being set by this line: https://github.com/neutrinojs/neutrino/blob/master/packages/jest/src/index.js#L60

Judging by eye, it looks like the rule is expecting a capturing group but the regex doesn't contain one (the actual mapping is coming from https://github.com/neutrinojs/neutrino/blob/master/packages/preact/index.js#L31)

This could also be connected to [preact/compat now living in core since preact X](https://preactjs.com/guide/v10/whats-new/#compat-lives-in-core)

---

This is easy to replicate in an empty project:

### package.json

```json
{
  "scripts": { "test": "jest" },
  "dependencies": { "preact": "^10.4.0" },
  "devDependencies": {
    "@neutrinojs/jest": "^9.1.0",
    "@neutrinojs/preact": "^9.1.0",
    "jest": "^25.3.0",
    "neutrino": "^9.1.0",
    "webpack": "^4.42.1",
    "webpack-cli": "^3.3.11",
    "webpack-dev-server": "^3.10.3"
  }
}
```

### .neutrinorc.js

```js
const jest = require('@neutrinojs/jest');
const preact = require('@neutrinojs/preact');
module.exports = { use: [jest(), preact()] };
```

### jest.config.js

```js
module.exports = neutrino().jest();
```

### webpack.config.js

```js
```

### src/index.js

```js
import React from 'react'; // should be auto-mapped to preact/compat by config
export default React;
```

### tests/index.spec.js

```js
import x from '../src/index';
it('does something', () => {
  expect(x).not.toBeNull();
});
```

constgen:
If we configure alias to React in Webpack which is required by `react-hot-loader` the Jest tests fail with an error 
```bash
Configuration error:

    Could not locate module react mapped as:
    C:\Projects\project\node_modules\react$1.

    Please check your configuration for these entries:
    {
      "moduleNameMapper": {
        "/^react$/": "C:\Projects\project\node_modules\react$1"
      },
      "resolver": undefined
    }

    > 1 | import React from 'react';
        | ^
      2 | import PropTypes from 'prop-types';
```

**To reproduce this** we need to add something like that to the Webpack aliases configuration

```js
neutrino.config.resolve.alias.set('react', path.resolve(path.join(projectNodeModulesPath, 'react')))
```

And import any JSX component in tests.

I am curios why we need this `$1` at the end. It only makes sense if the key contains angle brackets `()` like:

*Example from Jest docs* 
```json
"module_name_(.*)": "<rootDir>/substituted_module_$1.js",
"assets/(.*)": [
      "<rootDir>/images/$1",
      "<rootDir>/photos/$1",
      "<rootDir>/recipes/$1"
]
```

But we know that we can't ever have angle brackets in the alias key. 

The issue can be fixed by removing `$1` placeholder

Fixes #1552.

### DIFF
--- a/packages/jest/src/index.js
+++ b/packages/jest/src/index.js
@@ -63,7 +63,7 @@ module.exports = (options = {}) => (neutrino) => {
         moduleNameMapper: Object.keys(aliases).reduce(
           (mapper, key) => ({
             ...mapper,
-            [`^${key}$`]: `${getFinalPath(aliases[key])}$1`,
+            [`^${key}$`]: `${getFinalPath(aliases[key])}`,
           }),
           {
             [extensionsToNames(media)]: require.resolve('./file-mock'),
--- a/packages/jest/test/jest_test.js
+++ b/packages/jest/test/jest_test.js
@@ -161,7 +161,7 @@ test('configures webpack aliases in moduleNameMapper correctly', (t) => {
   const config = api.outputHandlers.get('jest')(api);
 
   t.true(
-    Object.entries(config.moduleNameMapper).some(function ([key, alias]) {
+    Object.entries(config.moduleNameMapper).some(([key, alias]) => {
       return key === '^react$' && alias === reactPath;
     }),
   );
--- a/packages/jest/test/jest_test.js
+++ b/packages/jest/test/jest_test.js
@@ -1,3 +1,4 @@
+import path from 'path';
 import test from 'ava';
 import airbnbPreset from '../../airbnb';
 import eslintPreset from '../../eslint';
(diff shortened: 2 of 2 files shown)


constgen:
Please, review. New tests passed. But unfortunately the build job failed any way. This happens because of incompatibility with the latest ESlint v7 as I understand looking to logs
изображение

edmorley:
> We really should not accept ESlint 7 in peer dependencies

ESLint 7 works fine with Neutrino and the tests passed when #1572 was merged. The issue was that ESLint 7.3.0 introduced a breaking change/regression (eslint/eslint/issues/13427), which was now been fixed in 7.3.1. I've retriggered the tests and they now pass.

PR #1584, opened 2020-06-10
Initial work to support Webpack 5

jacob-ebey:
All tests are passing locally but some touch-up work needs to be done.

Docs need to be updated and migration strategy / tests need to also be updated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "prettier": "^2.0.5",
     "verdaccio": "^4.6.2",
     "verdaccio-memory": "^9.5.0",
-    "webpack": "^4.43.0",
+    "webpack": "^5.0.0-beta.17",
     "webpack-cli": "^3.3.11",
     "webpack-dev-server": "^3.11.0"
   }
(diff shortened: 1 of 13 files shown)


ScriptedAlchemy (approved):

ScriptedAlchemy:
Some options here. At Next.js i introduced version detection and if you installed WP5, the conditionals would switch between the not config formats.

Is there a set list of plugins that are used? Those can be checked and in my experience its about 20 min per plugin to provide wp4 - Wp5 compatibility

edmorley:
> and if you installed WP5, the conditionals would switch between the not config formats.

Yeah that's the approach I'd like us to take if it's not too much hassle. 

> Is there a set list of plugins that are used?

All dependencies in this monorepo unfortunately - so that both CI will pass and end-users don't get surprises when using any of the official Neutrino presets.

edmorley:
Closing since unfortunately this project is no longer maintained. See:
https://github.com/neutrinojs/neutrino/issues/1707
https://github.com/neutrinojs/webpack-chain/issues/358

PR #1588, opened 2020-06-22
jest: Support using the preset standalone

constgen:
Fixes https://github.com/neutrinojs/neutrino/issues/813

I chose a way to extend Webpack config with `js` extension to guarantee the support of this extension. I think it should not break anything. Correct me if I am wrong


edmorley:
Also I wonder in the fallback case whether we should either:
1. Set just `js` as the only extension
2. Not set `moduleFileExtensions` at all, and so let it fall back to the Jest default https://jestjs.io/docs/en/configuration#modulefileextensions-arraystring
3. Use the default Neutrino `neutrino.options.extensions` from https://github.com/neutrinojs/neutrino/blob/230e81524693e3417d8e6f6be39d02e8d726ba88/packages/neutrino/extensions.js#L7

constgen:
As the case of a standalone Jest preset is a case in the project where no other Neutrino presets we may guess that the transpilation is already setup using alternative way. So we may expect JSX and TypeScript working somehow. And they may be imported in tests. That's why the **Option 2** is the most reliable. It considers this.

constgen:
After some investigation got some results

### If thre is no "js" extension in the array
Have just tested with `['jsx']` extension only. May be something changed in recent version in Jest but now it **fails with an error**
```
● Validation Error:

  moduleFileExtensions must include 'js':
  but instead received:
    ["jsx"]
  Please change your configuration to include 'js'.

  Configuration Documentation:
  https://jestjs.io/docs/configuration.html
```

The similar error is for empty array of extensions. So seems like Jest requires `"js"` to be present in the array.

### If extensions are undefined

Tried to define `moduleFileExtensions: undefined,` with a hope that this will trigger default Jest values. But Jest always processes it as an array and **throws an exception**

```
TypeError: Cannot read property 'map' of undefined
```

### If option is remove completely

Tried to remove `moduleFileExtensions` from the config at all. Now Jest started to use its default extensions. But **the parsing as a module is broken with an error**

```
 FAIL  test/utils/merge-samples.test.js
  ● Test suite failed to run

    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:

    Details:

    C:\Projects\project\test\utils\merge-samples.test.js:2
    import mergeSamples from '../../src/utils/merge-samples';
    ^^^^^^

    SyntaxError: Cannot use import statement outside a module

      at Runtime._execModule (node_modules/.pnpm/registry.npmjs.org/jest-runtime/26.0.1/node_modules/jest-runtime/build/index.js:1166:56)
```

But this is expected as Jest is not "teached" to transpile modules by Babel. This settings are just not provided by any preset. At least it may work with CommonJS modules

## Conclusion

Decided to use the Jest internal fallback options. Please review

packages/jest/src/index.js
```diff
@@ -47,19 +47,23 @@ module.exports = (options = {}) => (neutrino) => {
...
+      ? { moduleFileExtensions }
+      : undefined;
 
     return merge(
       {
         rootDir: root,
+        ...extensionsOptions,
         moduleDirectories: modulesConfig.length
           ? modulesConfig
           : ['node_modules'],
```
edmorley, line 66: To simplify this slightly, we can use the spread operator with a conditional statement, like so:

```suggestion

    return merge(
      {
        rootDir: root,
        moduleDirectories: modulesConfig.length
          ? modulesConfig
          : ['node_modules'],
        ...(moduleFileExtensions.length && { moduleFileExtensions }),
```

This is similar to what we do in a few other places, eg:
https://github.com/neutrinojs/neutrino/blob/b740f67a546a8efbb0e59a890ab4a1ab4b3ae9a2/packages/web/index.js#L144

Issue #1596, opened 2020-07-13
Multiple React Components Folder


hjwalt:
### Enhancement

Currently `neutrino.options.components` or `components` in `react-components` preset only allow a string to be supplied and published as component library. I think allowing an array of string would be useful for that configuration.

My use case is that I need to publish from three different reusable folders with three logical groupings, which are:

1. components
2. containers
3. pages

- What is the goal of the change?
To allow setting multiple component source folder

- What are the pros and cons of the change?
pros: multiple source folders allowed with simple configuration

cons: at the moment all the components would still be published into the same folder, which could be confusing unless the component configuration takes a folder map. i.e. :
```
{
    components: '', // Mapping to build/, there could be other ways to do this better
    containers: 'containers', // Mapping to build/containers
    pages: 'pages' // Mapping to build/pages
}
```

- Could this dramatically improve the experience of our users?
I don't know about other use cases but I think the capability of publishing multiple logical grouping of react components could be useful for other uses as well.

p.s. I am a developer too so if neutrino is open, I could develop an initial change implementation to react-components and create a pull.

edmorley:
Hi! Thank you for filing this issue. This does sound like it would be useful. I don't have much spare time for Neutrino development these days, but if you wanted to open a PR I'd be happy to take a look? :-)

hjwalt:
@edmorley thank you for your response, I have created the pull request for this change.

PR #1598, opened 2020-07-19
Multiple React Components Folder

hjwalt:
Submitting pull request per feature #1596 

This is my first time sending pull request to Neutrino, please let me know if there are things that I should be doing or should take note of.

Also, I am unsure of the testing methodology of neutrino even after looking at `component_test.js` in `react-components` package. I will need some guidance on this part.

### DIFF
--- a/packages/react-components/index.js
+++ b/packages/react-components/index.js
@@ -35,20 +35,37 @@ module.exports = (opts = {}) => (neutrino) => {
       neutrino.use(react(options));
     },
     () => {
-      const components = join(
-        neutrino.options.source,
-        options.components || 'components',
-      );
-
       Object.keys(neutrino.options.mains).forEach((key) => {
         delete neutrino.options.mains[key]; // eslint-disable-line no-param-reassign
       });
 
-      readdirSync(components).forEach((component) => {
-        // eslint-disable-next-line no-param-reassign
-        neutrino.options.mains[basename(component, extname(component))] = {
-          entry: join(components, component),
-        };
+      const componentPaths = [];
+      if (Array.isArray(options.components)) {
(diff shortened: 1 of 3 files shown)


packages/react-components/index.js
```diff
@@ -35,20 +35,37 @@ module.exports = (opts = {}) => (neutrino) => {
...
         delete neutrino.options.mains[key]; // eslint-disable-line no-param-reassign
       });
 
-      readdirSync(components).forEach((component) => {
-        // eslint-disable-next-line no-param-reassign
-        neutrino.options.mains[basename(component, extname(component))] = {
-          entry: join(components, component),
-        };
+      const componentPaths = [];
+      if (Array.isArray(options.components)) {
```
edmorley, line 43: I think we can simplify this slightly, by doing something like this psuedo-code:

```
if <options.components is null>
  set default value of `['components']

if <options.components is not an array>
  wrap current options.components value in an array

for each element in options.components
  // do stuff
```

I think we can ignore the empty array case, since it's a bit of an edge case, and it's not the default value, so if the user has passed an empty array, then it's an explicit choice to not set any components (although a nonsensical one).

edmorley (changes requested):
Hi! Thank you for the PR - sorry for the delayed review.

Could you add some tests for this to:
https://github.com/neutrinojs/neutrino/blob/master/packages/react-components/test/components_test.js

Thanks :-)

packages/react-components/index.js
hjwalt: Hi, thank you for reviewing, I have done the changes and tests as requested. Let me know if there is anything I should change.

PR #1630, opened 2020-10-03
Implement Yarn 2 support

### LINKED ISSUE #1277: Add support for Yarn's Plug n Play mode
Yarn has a new experimental and opt-in mode called "Plug n Play" that means packages are no longer installed into `node_modules` and are instead mapped to their cache location at runtime via a custom resolver. See:
https://github.com/yarnpkg/rfcs/pull/101
https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-plug-an-play.md

Whilst this new mode mostly "just works", it requires custom configuration when using certain packages such as webpack or jest. Therefore we should either add native support to Neutrino for these changes (like CRA has) or else document them.

If we were to add native support to Neutrino, we would need to:
* add/configure the various resolver plugins:
  - https://github.com/arcanis/pnp-webpack-plugin
  - https://github.com/arcanis/jest-pnp-resolver
* ensure that all of our dependencies/peer dependencies are correctly defined (Yarn will provide error messages about these, so it's just a case of trying and seeing if it works)
* ensure that we're using `require.resolve()` in all the places we should (which I believe we are, though again yarn will helpfully error if not)
* (optionally) add support for a `--pnp` option to create-project so projects can be created using PnP out of the box (rather than them needing to be converted over after by running `yarn --pnp` by hand)
* document the experimental support for PnP, including any limitations (eg the ESLint issue and whether any of the other test runners/... don't work)
* ask that Yarn hardcode Neutrino's ESLint related package names to work around https://github.com/eslint/eslint/issues/10125#issuecomment-425078944, similar to what they do for CRA (yarnpkg/yarn#6449)

See also:
https://github.com/yarnpkg/pnp-sample-app#plugins
https://github.com/facebook/create-react-app/pull/5136

constgen:
Resolves https://github.com/neutrinojs/neutrino/issues/1277

- [x] Bump minimal Jest version to support Plug and Play
- [x] Add PnP middleware
- [x] Make sure all resolves are PnP compatible
- [x] Use PnP middleware in all presets
- [x] Adopt all ESLint presets as there are known problems with ESLint sharable configs and PnP


constgen:
I can see in the build logs that the job published packages to NPM. But why? It even didn't pass the code formatting check. And this is only Pull Request.

packages/jest/package.json
```diff
@@ -34,7 +34,7 @@
     "lodash.omit": "^4.5.0"
   },
   "peerDependencies": {
-    "jest": "^24.0.0 || ^25.0.0 || ^26.0.0",
+    "jest": "^24.4.0 || ^25.0.0 || ^26.0.0",
```
edmorley, line 37: Bumping the value for a peer dep is a breaking change, so can't be done in Neutrino 9

edmorley (changes requested):
Hi! Thank you for the PR.

I much prefer Yarn over NPM and have wanted PnP support for a while.

That said:
* My understanding is that webpack 5 supports Yarn PnP out of the box, so much of this may be redundant in Neutrino 10+
* Jest needs updating to a minor newer version (but bumping the peer dep is a breaking change), so we'd have to have a docs element to this anyway
* This implementation adds the `pnp-webpack-plugin` dependency and config complexity to NPM users not just yarn

As such, I'm slightly reluctant to add a new middleware for Neutrino 9 only, which would be removed in Neutrino 10 - and think it might make sense for this to be a new section in the docs for Neutrino 9, that contains a snippet for people to add to their `.neutrino.js` along with some suggestions as to Jest upgrades etc.

Then in Neutrino 10 with webpack 5, this could just be supported natively with little implementation needed on Neutrino's side.

Thoughts?

edmorley:
> I can see in the build logs that the job published packages to NPM. But why?

It's not a real NPM server - see:
https://github.com/neutrinojs/neutrino/blob/master/scripts/test-create-project-ci.sh

packages/jest/package.json
constgen: So if we live it as is there is high chance that Jest will work as expected. But it is 99% only. 1% for those users that are on Jest <=24.4.0. May be we can live with this

constgen:
Here are my thoughts.
I have my personal set of Neutrino middlewares and presets for React, ESLint and other. I reuse couple of Neutrino core middlewares, but most are custom. I faced really a lot of problems with Yarn 2. Especially with ESlint. I resolved them. Some solutions were really hacky. So I decided to integrate my experience to the Neutrino core.

**Lets discuss your points**

- PnP will be supported in Webpack 5. This is true. Until this moment configuration need to use PnPPlugin. Some projects need this already today. Internally the plugin checks the PnP environment (Yarn documented how to do this) and enables necessary functionality on demand. Speaking simply it has no effect on NPM and PNPM projects. I also thought about conditionally enabling it but realized that it is easier to inspect Webpack Config when it is always included. This is how CRA and Storybook do - they always include PnPPlugin to configuration.
- I will document the integration of Jest and PnP. But please recommend me where to write "PnP" section 
- I thought that Neutrino goal is to hide the config complexity from the consumer. That's why I decided that `react` and `vue` presets should support this out of the box without thinking how to enable this. And according to the **first** point it should not affect the build process on NPM. Again, CRA does this.

packages/eslint/alias-plugins.js
```diff
@@ -0,0 +1,41 @@
...
+  return function aliasPlugin(pluginName) {
+    let resolvedPluginPath;
+
+    if (pnpApi) {
+      resolvedPluginPath = pnpApi.resolveRequest(pluginName, relativeFilename);
+    } else {
+      resolvedPluginPath = require.resolve(pluginName, {
+        paths: [relativeFilename],
+      });
+    }
```
merceyz, line 31: You should just use `require.resolve` you almost never need to use the pnpapi directly, this is one of those cases

constgen: I know. But unfortunately `__filename` gives a virtual File System path in **Yarn 2** which can't work correctly in `paths` options of `require.resolve`. Module will be not found with an exception.

merceyz: Do you have a reproduction for that? We have tests making sure it works. I could see that happening if you have `enableGlobalCache: true` but that will be fixed in https://github.com/yarnpkg/berry/pull/2068

constgen:
Please, look at the new section in the documentation https://github.com/constgen/neutrino-dev/tree/pnp/packages/eslint#utility-functions

packages/eslint/alias-plugins.js
constgen: In any place of your code in Yarn 2 environment try this `console.log(require.resolve('any-of-your-dependency', { paths: [__filename] }))`. It will not find your module even if it is relative from the current file. The same code works without any problems in regular NPM environment

packages/jest/package.json
constgen: the change is reverted

packages/eslint/alias-plugins.js
merceyz: It will, and it does. As I said we have tests that cover this, but just for fun I did try exactly what you said and it works, if you have a reproduction I'd love to take a look at it


Some tests that check this:
https://github.com/yarnpkg/berry/blob/f8344493b5cf3948e71211697b50a268b756d8fc/packages/acceptance-tests/pkg-tests-specs/sources/pnp.test.js#L1714
https://github.com/yarnpkg/berry/blob/f8344493b5cf3948e71211697b50a268b756d8fc/packages/acceptance-tests/pkg-tests-specs/sources/pnp.test.js#L609-L612
https://github.com/yarnpkg/berry/blob/f8344493b5cf3948e71211697b50a268b756d8fc/packages/acceptance-tests/pkg-tests-specs/sources/pnp.test.js#L647-L651

constgen: I realized that mentioned test case works in the Yarn 2 project itself. But in the **linked** module it fails with not found dependency. As you understand I use linked packages for local environments. Such modules are defined as
```json
{
 "devDependencies": {
   "@neutrinojs/eslint": "portal:C:/Projects/Projects_GitHub/neutrino-dev-fork/packages/eslint",
  }
}
```
`yarn install` is called to setup all links in `.yarn/cache` and `.pnp.js`
And after that I can run local script `"lint": "eslint ./src --ext .js,.jsx",` to run the code and test the case

merceyz: Yeah, that makes more sense. I'm fairly certain the issue you're running into was fixed in https://github.com/yarnpkg/berry/pull/2068, could you try with `yarn set version from sources` or by downloading the `bundle` here https://github.com/yarnpkg/berry/actions/runs/346381133?

constgen: Before: Yarn version `v2.2.2`
After: Yarn version `2.3.3-git.20201112.f5ca1c57`
Result is the same in the linked package: works when `require.resolve('babel-eslint')` and fails  when `require.resolve('babel-eslint', { paths: [__filename] })` 
```
Oops! Something went wrong! :(

ESLint: 7.9.0

Error: Cannot read config file: C:\Users\const\Desktop\yarn-test\.eslintrc.js
Error: Cannot find module 'babel-eslint'
```

constgen:
I think everything related to Yarn 2 is resolved. 
The only left are tests. And I think there is a separate problem. I feel like tests can affect each other and they are not isolated. Seems like they may have a shared neutrino config and all Webpack plugins are applied across test. Is it possible? Because currently just commenting one test can make other tests red or green.

constgen:
@edmorley in Webpack 5 only `pnp-webpack-plugin` will be not necessary. All other fixes to ESLint, Jest and etc. will be highly required. Their issues still remain. So with `pnp-webpack-plugin` or without it we need this PR any way. But without `pnp-webpack-plugin` it will be hard to automate testing in Yarn 2. In my opinion it affects only several presets and it will be really easy to remove it in Neutrino 10 without affecting anything.

packages/eslint/alias-plugins.js
merceyz: Could you try using `__dirname` instead?
Again, if you could provide a reproduction I'll be able to look into it

constgen: **Update:**
After removing `.yarn/cache` and reinstalling dependencies `require.resolve('babel-eslint', { paths: [__filename] })` started to work. It was the version `2.3.3-git.20201112.f5ca1c57`.
Than I upgraded Yarn to the latest `2.4.0` and reinstalled dependencies again. And it still works. So I removed usage of PnP API as you asked

constgen:
@edmorley @eliperelman can you look at final version of this PR?

edmorley:
@constgen Thank you for working on this :-) Before it can be merged (and so before spending time on review), we need to have CI passing, which is blocked on Travis CI credits running out, due to Travis CI's new policy (that is, there are no more free credits). The best way to resolve this would be to migrate to another provider -- I would recommend Circle CI. I don't have time to work on such a migration, but would be happy to review a PR that did so.

constgen:
Is GitHub Actions is an option? I tried it recently for one of my projects as a replacement for Travis. Have to say that it also has free plan limits but it worked for me well

edmorley:
Yeah that works too :-)

edmorley:
Closing since unfortunately this project is no longer maintained. See:
https://github.com/neutrinojs/neutrino/issues/1707
https://github.com/neutrinojs/webpack-chain/issues/358

PR #1660, opened 2021-03-08
Really make tests not fail on Windows due to EOL 

constgen:
Fixes #1628
Continues the work of another PR https://github.com/neutrinojs/neutrino/pull/1629 that not really fixed the issue

PR #1663, opened 2021-03-13
GitHub actions

constgen:
Enable built-in GitHub CI/CD


constgen:
@edmorley 
Can you look at build logs? 
1. Windows tests fail. This is an expected known issue. The PR is open (again) to fix this https://github.com/neutrinojs/neutrino/pull/1660
2. There is a publish job on Git tag. But it is optimized for a regular NPM package, not a monorepo. We should decide are we going to use it at all and how to redesign it for a monorepo
3. **The most important.** Tests fail on Ubuntu with exceptions during running `scripts/test-create-project-ci.sh`. I couldn't get the idea of this script and why it fails in this environment

edmorley:
Thank you for working on this :-)

> Windows tests fail. This is an expected known issue. The PR is open (again) to fix this #1660

The existing Travis run has no Windows testing. Whilst adding this would be great longer term, I would remove it from this PR for now, and aim for parity with the Travis config before adding more. Priority number one is getting working CI back for this repo :-) 

> There is a publish job on Git tag. But it is optimized for a regular NPM package, not a monorepo. We should decide are we going to use it at all and how to redesign it for a monorepo

The current Travis config doesn't support publishing, so let's remove for now. Publishing also currently requires 2FA, so would require some thought (or changes to the 2FA requirement) to automate.

> Tests fail on Ubuntu with exceptions during running scripts/test-create-project-ci.sh. I couldn't get the idea of this script and why it fails in this environment

The `test-create-project-ci.sh` runs a background package server using Verdaccio on localhost, so that it can publish the Neutrino monnorepo packages to that package server, and then run end-to-end tests against them using create-project. There have been strange timeout issues in the past, perhaps something needs tweaking with the Veraccio .yml config. I would recommend also trying with Yarn instead of NPM (the existing Travis config tested both) to see if the timeouts are NPM-specific or not.

constgen:
1. Windows removed from the matrix
2. Publishing job is removed. Actually now days NPM can generate a special 2FA token with which you can publish packages. You can check it in its token generation UI
3. Switched Verdaccio steps from NPM to yarn and it worked.

Looks like all checks pass now https://github.com/constgen/neutrino-dev/actions/runs/652165277

edmorley (commented):
Thank you for working on this!

Some thoughts:
1. The new CI config doesn't set `CREATE_PROJECT_YARN` -- this means it's installing yarn but actually testing NPM during the run, not yarn
2. The new CI config only tests one package manager, when before it tested both yarn and npm for create project. Should we check both?

